Issue #5, opened 2017-02-01
gitignore


CarloVanDriestenBMW:
Adapt the gitignore file - maybe everything except .proto licence and readme.md should be ignored? Other suggestions?

CarloVanDriestenBMW:
cmake ignores are ok - the review should solve the issue

PR #10, opened 2017-02-03
Daimler suggestions

SoykaDaimler:
Hi all,
here are my suggestions. They also include some questions all marked as TODO(FSoyka). So this pull request is not ready to be merged immediately but needs some discussion. I guess we do that on the phone this afternoon. Then we can also discuss the general work flow. For now, I just went ahead with the forking work flow. 
Best,
Florian

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ Makefile
 cmake_install.cmake
 install_manifest.txt
 CTestTestfile.cmake
+.project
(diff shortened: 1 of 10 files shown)


.gitignore
```diff
@@ -5,3 +5,4 @@ Makefile
 cmake_install.cmake
 install_manifest.txt
 CTestTestfile.cmake
+.project
```
CarloVanDriestenBMW: That's a good idea!

CarloVanDriestenBMW: #5

PR #21, opened 2017-02-06
Simplify filenames

mbinna:
I simplified the filenames:

* Remove rendundant prefix "osi_"
* Separate different words with underscores for better readability

Please have a look and let me know what you think about this change.


CarloVanDriestenBMW:
I think it is better like that - I requested @TimoHankeBMW and @nHirsenkorn as reviewer for their approval.

TimoHankeBMW:
Not sure about this.
I liked the prefix in the files as it clarifies imports on the calling code side (e.g. "include <osi_object.pb.h>" instead of "include <object.pb.h>") and avoids possible naming collisions as CMake puts all generated pb.cc and pb.h files in CMAKE_CURRENT_BINARY_DIR, which isn't great for scoping.
Both aren't quite show stoppers, so I'd like to hear other opinions on that.

If we go for the renaming, CMakeLists.txt has to be modified as well.

nHirsenkorn:
I don't care about the naming.

pmai:
I concur with @TimoHankeBMW, removing the osi_ prefix will very likely lead to common conflicts in filenames, especially header files which tend to have to live in global namespaces. Minor regularization of the names might be OK, but I'd definitely keep the prefix (we are just asking for trouble later on, and changing back will be trouble as well).

CarloVanDriestenBMW:
I would keep the current naming. The prefix ssems necessary for me according to the arguments of @TimoHankeBMW and @pmai .
Therefore the "osi_suffix" is also ok in a single word for me - it visualizes the seperation of prefix=osi and the suffix describing the content.

RudolfVocilka:
I would be in favor of simple filenames. It would be nice to use it like:
`#include <osi/common.pb.h>`

It is actually the same for python. You need a package (name osi, version 2.0.0 etc.)  that can contain multiple modules (osi_common_pb2). So the osi should be encoded in the package not in the module names (as prefix).

`from osi import common_pb2`
instead of
`from osi import osi_common_pb2`

mbinna:
@RudolfVocilka: I like your idea. Could you make a suggestion how to achieve the grouping of the C++ headers in the folder `osi` and the Python package `osi`? Do we just have to place the headers in a folder named `osi`?

RudolfVocilka:
Yes that would be enough. In the top answer in the second paragraph in this [thread](http://stackoverflow.com/questions/13521618/c-project-organisation-with-gtest-cmake-and-doxygen) it is explained well.

TimoHankeBMW:
Does this work with the automatically generated header and source files by CMake without manually moving files?

pmai:
I think in order for this to work reliably for outside users the proto files should be moved to a sub-directory named `osi` and the generated files as well, referenced from the CMakeLists.txt in the root directory. The target_include_directories PUBLIC declaration would remain the same, but you might need to include the osi directory in a non-PUBLIC declaration for the internal includes to work (haven't tried this out).

Otherwise we would expose the non-prefixed names to outside projects (via the target_include_directories), and I guarantee that there will be conflicts with object.h, etc.

mbinna:
@pmai Like this?

TimoHankeBMW:
@mbinna Did you test where the compiled proto headers wind up? As I said, the GENERATE_PROTO_CPP function places those in CMAKE_CURRENT_BINARY_DIR and I am not sure, if this changes due to the proto files being placed in a sub-directory.

mbinna:
@TimoHankeBMW: I added another commit. The generated C++ files are placed in `${CMAKE_CURRENT_BINARY_DIR}/osi`. Since `${CMAKE_CURRENT_BINARY_DIR}` is added to the target include directories, a user should have to `#include <osi/object.pb.h>`. I assume this is what we want to achieve. Let me know if I got something wrong.

pmai:
Hmm, I think this still will not work as intended for out-of-directory builds, since cmake will just put the resulting files into a flat build directory, as far as I can tell?

CarloVanDriestenBMW:
@mbinna Do you have to make do a rebase here? I can't merge like that
> This branch has conflicts that must be resolved

RudolfVocilka:
I updated the pull request for Version 2.0.1. Simple names work now with python and cpp. Example is shown in readme.md

CarloVanDriestenBMW:
Great work @RudolfVocilka !!!

I would be very happy if @pmai would verify this and maybe update to v2.1.0 ?
I am not sure if a new release name should be chosen? This would break other builds?

RudolfVocilka:
@CarloVanDriestenBMW This would break every other implementation  that is using the includes osi_common.pb.h instead of osi/common.pb.h introduced in this pull request

pmai:
Not to rain on anyones parade, but I've now found the time to look deeper into integrating this changed build setup into dependent projects, and I'm not seeing great ways of integrating this, mostly due to CMake limitations and brokenness surrounding both FindPackage and ExternalProject_add, and related issues (see e.g. http://stackoverflow.com/questions/12021448/how-can-cmake-arguments-be-forwarded-to-externalproject etc.).

As it stands this setup will force users to first install open-simulation-interface prior to referencing it in their project builds, which is not conducive to modular builds. The work-around of installing into a build subdirectory tends to either break one-pass building of dependent projects, and/or requires lots of manual variable passing to work around cmake brain-damage, which is brittle to say the least (cf. superbuilder pattern and ExternalProject_add passing of cmake cached variables).

I'll continue to look into this (and anyone please feel free to look at e.g. the osi-sensor-model-packaging project as a simple example test case), but I'm starting to feel that working around this is increasingly not worth the effort for the small gain in naming niceness.

Just to reiterate, what I would expect from a usable build setup is that it enables the  following:

- Allow referencing of exact OSI commit via git submodules
- Allow one-stage build of whole project, including embedded OSI
- Pass all cached variables of top-level cmake invocation (e.g. for passing of PROTOBUF directories to use for FindPackage, as well as build type, etc.) to any needed sub-level cmake invocations
- Do not pollute anything outside the top-level build directory
- Do not require maintenance of large lists of variables to be passed along

IMHO this should not be too much to ask, but it increasingly seems to me that this is not that easy to achieve with cmake if one of the external dependencies requires an install step to occur prior to the package being detectable/usable. But please, any CMake guru feel free to correct my impression.

CarloVanDriestenBMW:
I assigned my only hope for this @mbinna # CMakeGuru

PhilippRoethigBMW:
@pmai is the problem exclusively related to the build and install process introduced in this branch, or to the use of the simplified file names?

CarloVanDriestenBMW:
[Let's look at this!!! 💃 ](https://github.com/google/protobuf/tree/master/cmake)

pmai:
The problem mostly stems from the restrictions of the protoc rules, which lead to the generated header files ending up directly in the binary build directory, not in a subdirectory named osi, so that dependent targets cannot reference those header files in the intended manner (i.e. as osi/....h). The install step (and bundling as a package) was hoped to resolve this issue, since the files now end up in a proper location, however the support by CMake for building&installing packages prior to referencing them all leave something to be desired, to say the least, so this work-around doesn't seem to help much at all (it rather makes things more complicated for users, it seems to me). I'd hoped for a better way, and will look further in the next couple of days...

For simple projects like the OSMP stuff, I could hack it so that it works with the unadorned include names, but for larger, proper, projects that is not likely to work due to conflicts.

PR #24, opened 2017-02-08
Simplified CMakeLists.txt

cmuck:
* Reusing the project name which avoids redeclaration and typos
* Using include directories instead of target include directories
  for each target
* Removed FILE GLOB because it is evil and not recommended by Kitware
* Reformatted file

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,13 +1,37 @@
 cmake_minimum_required(VERSION 3.7)
-project(open-simulation-interface)
+
+project(open_simulation_interface)
+
 find_package(Protobuf 2.6.1 REQUIRED)
-file(GLOB OSI_PROTO_FILES osi_*.proto)
-PROTOBUF_GENERATE_CPP(PROTO_SRC PROTO_HEADER ${OSI_PROTO_FILES})
-add_library(open_simulation_interface STATIC ${PROTO_SRC} ${PROTO_HEADER})
-target_include_directories(open_simulation_interface PUBLIC ${PROTOBUF_INCLUDE_DIR})
-target_link_libraries(open_simulation_interface PUBLIC ${PROTOBUF_LIBRARY})
-add_library(open_simulation_interface_pic STATIC ${PROTO_SRC} ${PROTO_HEADER})
-set_property(TARGET open_simulation_interface_pic PROPERTY POSITION_INDEPENDENT_CODE ON)
-target_include_directories(open_simulation_interface_pic PUBLIC ${PROTOBUF_INCLUDE_DIR})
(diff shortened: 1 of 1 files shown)


pmai:
Note that this breaks the build for osi-sensor-model-packaging (and all other users of OSI) for a couple of reasons:

 * `include_directories` does not add the include directories as `PUBLIC`, hence targets in other files (at higher levels of the directory hierarchy) that add the library via `target_link_libraries` will not automatically add the include directories.
 * Removing the cached variable for the generated include files breaks the build, again since `include_directories` are not propagated.

Both problems can be remedied by switching the `include_directories` back to `target_include_directories` with `PUBLIC`, which would also remove the need for the ugly cached variable.

CarloVanDriestenBMW:
@pmai I am out of this "Titan-Battle" :dagger: 

I guess you are  adjust the suggestion of @cmuck to fit all use cases.

CMakeLists.txt
```diff
@@ -1,13 +1,37 @@
...
+    osi_modelinternal.proto
+    osi_object.proto
+    osi_occupant.proto
+    osi_sensordata.proto
+    osi_sensorspecific.proto
+)
+
+protobuf_generate_cpp(PROTO_SRCS PROTO_HEADERS ${OSI_PROTO_FILES})
+
+include_directories(${PROTOBUF_INCLUDE_DIRS})
```
pmai: i.e. if you switch these two `include_directories` to

```
target_include_directories(${PROJECT_NAME} PUBLIC ${PROTOBUF_INCLUDE_DIR})
target_include_directories(${PROJECT_NAME} PUBLIC ${CMAKE_CURRENT_BINARY_DIR})
target_include_directories(${PROJECT_NAME}_pic PUBLIC ${PROTOBUF_INCLUDE_DIR})
target_include_directories(${PROJECT_NAME}_pic PUBLIC ${CMAKE_CURRENT_BINARY_DIR})
```

or a loop to this effect I think the change could be merged.

cmuck:
I don't know which build should be broken because I've tested it with `mkdir build && cd build && cmake .. && make` and everything was fine.

For re-usage and portability reasons outside of this OSS project (here we don't have a higher hierarchy or other targets - only one single CMakeLists.txt), I've changed it to `target_include_directories`. Would be great if you could explain how you use this project.

Can't follow why you want to build something with a cleared CMake cache :-/

pmai:
Sorry for the confusion, the build I was refering to was https://github.com/OpenSimulationInterface/osi-sensor-model-packaging and other users that include OSI as a submodule, which was the reason for the original CMakeLists.txt structuring.

pmai:
D'Oh, I missed the now missing library dependencies during review, will fix in new branch.

PR #25, opened 2017-02-08
Applied PEP8 for Python files

cmuck:


mbinna:
I'd vote for an all-or-nothing approach to PEP8. And since I like PEP8, let's go for the 79 line width.

CarloVanDriestenBMW:
I agree. Therefore @cmuck should alter the commit to adress the line length problem within the setup.py ?

PR #36, opened 2017-02-13
Remove need for sudo

mbinna:
Not using sudo is a good thing: it reduces boot time from 20-52s to 1-6s.

See: https://docs.travis-ci.com/user/ci-environment/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ before_script:
 script:
     - mkdir -p build
     - cd build
-    - cmake -D CMAKE_PREFIX_PATH:PATH=${DEPS_DIR}/protobuf/install ..
+    - cmake -D CMAKE_PREFIX_PATH=${DEPS_DIR}/protobuf/install:${PATH} ..
     - cmake --build .
     - cd ..
     - python setup.py build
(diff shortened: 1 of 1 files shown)


pmai:
In order for this to work you probably need to supply the alternative protobuf install paths to the cmake call via -D, otherwise the protobuf will not be found. Probably also need to adjust the cache check, otherwise this will not build correctly on old caches. I'll try a commit that does this.

.travis.yml
```diff
@@ -41,7 +40,7 @@ before_script:
 script:
     - mkdir -p build
     - cd build
-    - cmake ..
+    - cmake -D CMAKE_PREFIX_PATH:PATH=${DEPS_DIR}/protobuf/install ..
```
mbinna: Why didn't you write it like `CMAKE_PREFIX_PATH=${DEPS_DIR}/protobuf/install:${PATH}`? Does the current version do something different?

pmai: The -D command-line flag vs. environment variable is mostly just our default practice to avoid reliance on environment variables which are passed through to other processes. The reason we don't append the PATH is that CMAKE_PREFIX_PATH is just a set of additional paths to look for packages, which defaults to empty. cmake will continue to use the PATH environment variable normally in any case.

pmai:
Not sure why the build fails, probably the issue of path separators? Anyway including PATH on the CMAKE_PREFIX_PATH is not what we want, since CMAKE_PREFIX_PATH serves a different purpose, and PATH is used by cmake normally anyways, depending on system settings.

Issue #15, opened 2017-02-03
Add polygon contour to Dimension3d


SoykaDaimler:
Can we have an additional polygon base in Dimension3d? 
I'd like a sensor to be able to return an occluded shape different from a rectangle. 

We could implement it as a list of Vector2d or as Vector3d with z=0.

See reference below.

nHirsenkorn:
If required we can add this, but not inside the Dimension3d. Dimension3d is intended for bounding boxes only. Instead you can open a new field.

Currently I don't known of a standard sensor describing a shapea other than a rectangle. Consider adding this polygon field to a certain sensor in  osi_sensorspecific.proto.

If the field was inteded for enabling some freespace description we should consider opening a new message type on the level as GroundTruth and SensorData. 
Sensor-models could then work like this:
GroundTruth -> Freespace
or like this:
GroundTruth -> SensorData -> Freespace

TimoHankeBMW:
Is the use case of a polygon base to allow for "more sophisticated" occlusion calculation or should that type of bounding box be detected, i.e. output, by a sensor?

SoykaDaimler:
Thanks for your comments!
The use case is occluded shapes as sensor outputs. Simple 2D sensor inputs should be rectangular bounding boxes. But the outputs can be any kind of 2D polygon. 
@TimoHankeBMW Your comment about the interface not being designed for that is a bit unclear too me. I thought our purpose within PEGASUS is to extend the interface in such a way that it meets our use cases. Therefore I thought we try to implement all the necessary data structures. 

@nHirsenkorn I have not thought about Freespace here, but within PEGASUS it is a use case. Therefore we do not freespace as an output, so we do need 2D polygons anyway. I think the interface has to be checked against all the definitions and use cases we came up with in PEGASUS anyway. This should be discussed in on of our next meetings -> @raueATconti

TimoHankeBMW:
@SoykaDaimler Yes, the interface should definitely be extended to meet the required use cases. My point above was just meant as a caveat about geometric calculations. Detailed geometric calculations are the playing field for a proper scene graph or similiar specialized data structures.
If there is a specific use case, its fine to include additional geometric shapes for objects. As stated above I'd just advise caution that no one starts having ideas about implementing a depth map or even ray tracer against the OSI data structures.

SoykaDaimler:
Agreed. Thanks for the clarification!

KathaHo:
Hi there!
[...]

Issue #39, opened 2017-02-14
"osi_lane.proto" Boundary Point


WeiskopfDaimler:
Which Reference Frame do I use for Boundary Points? Do I define them related to the lane reference line or in world coordinates?

nHirsenkorn:
Short answer: world coordinates

Extended answer:
At the moment only two reference frames within OSI exist. Objective is to keep it as simple as possible:
1. world reference frame (all quantities which are part of groundtruth)
2. sensor reference frame (all quantities which are part of sensordata)

Therefore the answer depends:
If you are adding a lane to groundtruth use the word reference frame.
If you are adding a lane to detectedlane use the sensor reference frame.

SoykaDaimler:
Here's another example for possible confusions regarding reference frame. As mentioned in #14 I think it would be good to give all coordinates an identifier describing their reference frame. 
A similar point came up in #15 
As @nHirsenkorn described above: the same quantities are used to describe data in the world or the sensor frame, depending on if they are part of groundtruth or sensordata. In my opinion this is very error prone.

Issue #14, opened 2017-02-03
Reference frames specified directly in coordinates


SoykaDaimler:
I saw that reference frames are defined in many places in the interface. From experience I would say that   
it is very valuable to know explicitly for each coordinate in which frame it is defined.  Therefore, I proposed some code in the branch DaimlerSuggestions (see reference below). 
All frames should be documented in the Wiki as well.

TimoHankeBMW:
Defining the reference frame as an enum that goes along with the coordinates allows the possibility of changing the reference frame for a specific message. Do we want this? You would have to check the reference frame and do possibly do a transformation at any time that you access coordinate values, which increases code complexity quite a bit.
We could specify that the reference frame must have a specific value for each message, but then I don't see the added benefit of including the enum.

TimoHankeBMW:
Accidentaly hit close there, please ignore :-)

nHirsenkorn:
I am of the same opinion as Timo Hanke: 
Effectively this would make me call a converter each time a protobuffer with unknown reference frame is received. If this call is forgotten, the Code might silently fail. With the current definition this conversion process is only required once, when filling the protobuffer, not each time when using it.

SoykaDaimler:
I agree that it increases complexity. On the other hand it also increases quality assurance and safety. By explicitly defining the reference frame as part of the message people are forced to think about it. If the reference frame is 'only' defined in the comments it can be overlooked. And in the case of using a message for a new use case there's a decreased likelihood of using it in the wrong coordinate system. 

And 'bugs' that are due to a wrong reference frame are very hard to track down. Some people even go as far as defining units for a message. In that case the Vector3d message for position, velocity and acceleration would differ. 

I agree that it is more code, but it think in the long run the benefits are worth it. 

One more point in favor of including it: adding the enum would not mean that you necessarily have to change already existing code. You can still just assume that the data is in the correct reference system and treat it as before. So you can think of it as an additional optional check.

What do other people think about this issue?

nHirsenkorn:
If decide to add coordinate system support, we could add ROS-style transfer functions:
For example:
A transfer function for sensor positions are defined relative to the bounding box middle of the vehicles
A transfer function for the environment model origin is also defined relative to the bounding box middle.

This would enable a standardized conversion from any cosy to any other. Even for multiple vehicles e.g. for car2car communication/fusion.

SoykaDaimler:
[...]

PR #40, opened 2017-02-14
Contour2d

SoykaDaimler:
Hi all,
based on our discussion in #15 I created a new branch which just includes the suggestions for a 2D contour. And two small additions to .gitignore.
@AndreasVIRES Does this definition comply with the OpenDrive specification?
@nHirsenkorn I added the contour to BaseStationary and BaseMoving, since you did not want it to be mixed up with the Dimension3D bounding box. I hope this fits better?
Best,
Florian

### DIFF
--- a/osi_common.proto
+++ b/osi_common.proto
@@ -12,8 +12,8 @@ message InterfaceVersion
     /// The field containing the version number. Should be left on default, not to be modified by sender. Increments will
     /// happen as part of changes to the whole interface.
     optional uint32 major = 1 [default = 2];
-    optional uint32 minor = 2 [default = 0];
-    optional uint32 patch = 3 [default = 1];
+    optional uint32 minor = 2 [default = 1];
+    optional uint32 patch = 3 [default = 0];
 }
 
 ///
--- a/osi_common.proto
+++ b/osi_common.proto
@@ -33,6 +33,19 @@ message Vector3d
 }
 
 ///
+/// A cartesian 2D vector for positions, velocities or accelerations.
+/// Units are [m] for positions, [m/s] for velocities and [m/s^2] for accelerations.
+///
+message Vector2d
+{
+    /// The x coordinate
+    optional double x = 1;
+
+    /// The y coordinate
+    optional double y = 2;
+}
+
+///
 /// A timestamp.
 /// Names and types of fields chosen in accordance with google/protobuf/timestamp.proto to allow a possible switch in the
 /// future. Definition of zero point in time differs and does not use Unix epoch.
--- a/osi_common.proto
+++ b/osi_common.proto
@@ -162,7 +162,8 @@ message BaseStationary
     /// Usage as sensor data:
     /// The polygon describes the visible part of the object's contour.
     /// General definitions:
-    /// The polygon is defined in the local object frame: x pointing forward and y to the left.      
+    /// The polygon is defined in the local object frame: x pointing forward and y to the left.
+    /// The origin is the center of the bounding box.
     /// As ground truth, the polygon is closed by connecting the last with the first point. Therefore  
     /// these two points are different and there are at least three points.
     /// As sensor data, the polygon is open.
@@ -204,7 +205,8 @@ message BaseMoving
     /// Usage as sensor data:
     /// The polygon describes the visible part of the object's contour.
     /// General definitions:
-    /// The polygon is defined in the local object frame: x pointing forward and y to the left.      
+    /// The polygon is defined in the local object frame: x pointing forward and y to the left.
+    /// The origin is the center of the bounding box.
     /// As ground truth, the polygon is closed by connecting the last with the first point. Therefore  
     /// these two points are different and there are at least three points.
     /// As sensor data, the polygon is open.
--- a/osi_common.proto
+++ b/osi_common.proto
@@ -155,7 +155,12 @@ message BaseStationary
     /// The relative orientation of the landmark w.r.t its parent frame.
     optional Orientation3d orientation = 3;
     
-    /// The two dimensional (flat) contour of the object. 
+    /// The two dimensional (flat) contour of the object. E.g. the projection of the contour onto the street.
+    /// The polygon is closed by connecting the last with the first point. Therefore these two points are 
+    /// different and there are at least three points.
+    /// The polygon is defined counter-clockwise.
+    /// For ground truth data the polygon is defined in two dimensional world coordiantes.
+    /// For sensor data the polygon is defined in two dimensional sensor coordinates.
     repeated Vector2d base_polygon = 4;
 }
 
@@ -186,6 +191,11 @@ message BaseMoving
     /// Obviously, the orientation rate becomes global/absolute if the parent frame is not rotating.
     optional Orientation3d orientation_rate = 6;
     
-    /// The two dimensional (flat) contour of the object. 
+    /// The two dimensional (flat) contour of the object. E.g. the projection of the contour onto the street.
+    /// The polygon is closed by connecting the last with the first point. Therefore these two points are 
+    /// different and there are at least three points.
+    /// The polygon is defined counter-clockwise.
+    /// For ground truth data the polygon is defined in two dimensional world coordiantes.
+    /// For sensor data the polygon is defined in two dimensional sensor coordinates.
     repeated Vector2d base_polygon = 7;    
 }
--- a/osi_common.proto
+++ b/osi_common.proto
@@ -154,6 +154,9 @@ message BaseStationary
 
     /// The relative orientation of the landmark w.r.t its parent frame.
     optional Orientation3d orientation = 3;
+    
+    /// The two dimensional (flat) contour of the object. 
+    repeated Vector2d base_polygon = 4;
 }
 
 ///
@@ -182,4 +185,7 @@ message BaseMoving
     /// The relative orientation rate of the object w.r.t. its parent frame and parent orientation rate.
     /// Obviously, the orientation rate becomes global/absolute if the parent frame is not rotating.
     optional Orientation3d orientation_rate = 6;
+    
+    /// The two dimensional (flat) contour of the object. 
+    repeated Vector2d base_polygon = 4;    
 }
--- a/osi_common.proto
+++ b/osi_common.proto
@@ -155,12 +155,18 @@ message BaseStationary
     /// The relative orientation of the landmark w.r.t its parent frame.
     optional Orientation3d orientation = 3;
     
-    /// The two dimensional (flat) contour of the object. E.g. the projection of the contour onto the street.
-    /// The polygon is closed by connecting the last with the first point. Therefore these two points are 
-    /// different and there are at least three points.
-    /// The polygon is defined counter-clockwise.
-    /// For ground truth data the polygon is defined in two dimensional world coordiantes.
-    /// For sensor data the polygon is defined in two dimensional sensor coordinates.
+    /// Usage as ground truth:
+    /// The two dimensional (flat) contour of the object. This is an extension of the concept of a 
+    /// bounding box as defined in Dimension3d. The contour is the projection of the object outline 
+    /// on the z-plane in the object frame (independent of its current position and orientation). 
+    /// Usage as sensor data:
+    /// The polygon describes the visible part of the object's contour.
+    /// General definitions:
+    /// The polygon is defined in the local object frame: x pointing forward and y to the left.      
+    /// As ground truth, the polygon is closed by connecting the last with the first point. Therefore  
+    /// these two points are different and there are at least three points.
+    /// As sensor data, the polygon is open.
+    /// The polygon is defined counter-clockwise.    
     repeated Vector2d base_polygon = 4;
 }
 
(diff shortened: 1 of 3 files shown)


osi_common.proto
```diff
@@ -154,6 +154,9 @@ message BaseStationary
 
     /// The relative orientation of the landmark w.r.t its parent frame.
     optional Orientation3d orientation = 3;
+    
+    /// The two dimensional (flat) contour of the object.
```
nHirsenkorn: please further extend the comment: is the last point equal to the first point?

SoykaDaimler: Good point! I will further clarify. 
I would suggest that the first and last are not equal to avoid redundancy.
Furthermore I would suggest counterclockwise definition of the points. 
Then there is the question about the reference frame. We usually had the middle of the rear axle. But as far as I understood it, in OSI the standard frame is the middle of the bounding box.
@nHirsenkorn Is this correct?

AndreasVIRES: > Furthermore I would suggest counterclockwise definition of the points.

In OpenDRIVE there is no order foreseen, i.e. the tool would need to be able to calculate and swap the sequence. Is this needed by the sensor modellers?

> Then there is the question about the reference frame.

The contour is given in _relative_ coordinates to that reference point, isn't it?

nHirsenkorn: @SoykaDaimler I agree to your further definition. Consider adding counterclockwise when in bird-view. The question about reference frames will be clarified in #39

SoykaDaimler: Our sensor models require counter clockwise ordering, that's why I suggested it. 
I guess in the end the sensor model has to check the order anyway and potentially correct it if the wrong order was provided. So I would keep the requirement for CCW ordering in the interface if no one objects, but keep it up to the model to really check if this requirement was fulfilled.
I agree with the two reference frames mentioned in #39, ground truth should be in world coordinates and the sensed contour in sensor coordinates. We do need access to the specifications where the sensor is mounted to do further transformations ourselves. I guess this is given in relation to the middle of the bounding box? Then we also need the relations between the middle of the bounding box and the middle of the rear/front axis since we have other coordinate systems which are in relation to these points. 
@nHirsenkorn These specs are not yet implemented in OSI, are they?

AndreasVIRES: > but keep it up to the model to really check if this requirement was fulfilled.

Either it has to be fulfilled by the tool or not. Shouldn't be too complicated for us, I just wanted to clarify.

**Reference frame:**
At the moment I cannot really follow the discussion. Usually you have a position (world, relative, road, whatever), an orientation and a contour (or bounding box) _relative_ to the given position and orientation.

E.g.: 
Object position (reference point): x=10, y=20, z=0
Object orientation: hdg=30°, pit=0°,rol=0°
Object contour: p1[2;0], p2[-2;-2];p3[-2;2]

This results in a rectangle at x=10, y=20, rotated by 30°:
example

nHirsenkorn: Nope. This is already implemented:
determining the absolute position of a sensor requires two transformations. 
See SensorData::mounting_position and Vehicle::relative_ego_origin

SoykaDaimler: @AndreasVIRES Thanks for the nice visualization! In our group, we define the contour in world and sensor coordinates, and not relative to the position/orientation.
@nHirsenkorn Ok, thanks for clarifying. Then I will also add the offset vectors for the rear / front axle in the vehicle description.

AndreasVIRES: @SoykaDaimler This information would also be available with a "position+orientation+contour" definition, you just need to calculate it. Sure, the tool could provide the oultine of an object in world coordinates or sensor coordinates but that is a redundant information because you still need "(reference point)position+orientation".

I also would suggest to discuss the reference point of vehicles again. In my opinion it makes no sense to have different reference point definitions, depending on whether a vehicle is a "Ego" or not. OpenSCENARIO until now does not provide any means to distinguish between vehicles that are an "Ego" and other vehicles. Reference point of a vehicle is always center of rear axis on ground in neutral load conditions.
@raueATconti the whole coordinate topic might be worth discussing it on the workshop.

nHirsenkorn: @AndreasVIRES All objects have the same reference point. Independent of ego or not. What is causing confusion here is:
A position vector consists of two points:
start point: This is defined by the reference frame. (in OSI: sensor-frame or world frame)
end point: often referred to as reference point. It is **always** the middle of the bounding box. This is not depending on being an ego vehicle or not, and not depending on the coordinate system (reference frame).

Moreover a vector always is dependent of the direction in which it is regarded. This is also captured by the reference frame.

In other words: inconsistencies between ego and other vehicles do not exist.

I will capture this in a wiki article and refer to it here.

nHirsenkorn: [Coordinate System Wiki page](https://github.com/OpenSimulationInterface/open-simulation-interface/wiki/Coordinate-systems-and-reference-points)

SoykaDaimler: I think the sensed contour should be given directly in sensor coordinates. This is more natural for a sensor, because a sensor does not sense a position in the middle of a bounding box or where ever, but it senses the contour of an object in sensor coordinates. 
I agree with @AndreasVIRES that we should discuss the coordinate topic in general all together during the workshop.

AndreasVIRES: @nHirsenkorn I still see an inconsistency because (as the name of the vector indicates) only Ego vehicles provide a vector **relative_ego_origin**, i.e. for non Ego vehicles there is less information on the interface.

In my opinion we need to agree on a vehicle reference point for all vehicles in the simulation, which not necessarily needs to be the rear axis but also not necessarily needs to be the center of the bounding box.

By the way, the real bounding box changes, even during runtime (opening doors but also vehicle movement, spring compression, etc.).

nHirsenkorn: @AndreasVIRES I see your point: you would like all vehicles to have the same informations. However for non-ego vehicles this information is never required. We should rather consider dropping relative_ego_origin generally. It seems to cause irritations and can in principal be handled by the environment model. I will open an issue.

nHirsenkorn: @AndreasVIRES the reference point is already identical for all vehicles in the simulation. It never is rear axis, but always the middle of the bounding box.

You are right about the changing bounding box. however it practically has no impact on the sensors. Therefore treating it in a static manner is sufficient :)

CarloVanDriestenBMW: You guys just tell me when to hit merge alright?  ^^

nHirsenkorn: nope. We could hit it ourselves xD

AndreasVIRES: @nHirsenkorn The vehicle reference point is never rear axis?

This is a definition found together with OEMs over years and also the result of many PEGASUS workshops:

> Ego-Fahrzeug:
> - Position: x[m], y[m], z[m] (Weltkoordinaten; **Bezugspunkt Fahrzeug: Mitte Hinterachse**));

This is a blocking point for VIRES for supporting OSI or not and I'll end discussion here, because nobody of the other tool suppliers, modelers and not all OEMs are participating the discussion.

@raueATconti We need to clarify this before we'll start implementation on our side.

nHirsenkorn: @AndreasVIRES End-Points of Positions always refer to the middle of the bounding box. Finding out the position is solely possible if relative_ego_origin is set to the rear axis. However relative_ego_origin is solely required for transforming from world to sensor-coordinates or for transforming from sensor-coordinates to the environment model. It does not play any role when staying in the same coordinate system (world or sensor).

We explicitly did not choose the rear axis as reference point, as it plays no role for sensors and just increases the complexity of finding out other points on the vehicle.

I get your point, we should end the discussion here. Further commitment would be helpful.
I am also not willing to explain every reasoning multiple times just due to the lack of participation of others.

SoykaDaimler: I agree and I think that it is very important that the reasoning you (OSI developers) had in mind when creating OSI, is written down somewhere and explained to all the partners who should be using OSI. We need all the PEGASUS partners on board in order to get through this important initial phase of establishing a common interface.

RudolfVocilka: @SoykaDaimler
> we define the contour in world and sensor coordinates, and not relative to the position/orientation.
If the contour is defined in world coordinates, why is there no Z-component in the contour? On what plane is the z-component projected? Or am I missing something. If z is simply set to 0 there may be problems with lanes with elevation. Or can I imagine it as a list of points on a plane, that is then translated and rotated.

If it is defined in sensor coordinates it should be relative to position / orientation right? 

Regarding the reference frame discussions I would leave it as is for this pull request.

SoykaDaimler: @RudolfVocilka 
Hi, I just discussed this issue with @CarloVanDriestenBMW on the phone. We use the contour in a purely two dimensional context. That's why I also defined a vector2d in order to avoid confusion about the z coordinate. In a 2D world there is no elevation.

I agree with solving the reference frame discussion elsewhere.

SoykaDaimler:
I updated the description of the contour fields based on our discussion. 
I added offset vectors from the bounding box center point to the middle of the front/rear axle to allow for coordinate transformations. 
Please review these changes.
I think the coordinate system discussion in general has to be solved elsewhere including everybody. 
The changes I suggest can later be adapted to the coordinate system of our choosing.

AndreasVIRES:
1. Does "projection to the street" mean that street parameter like superelevation, etc. are considered? If not maybe we should simply call it "projection to the z-plane".
2. "middle of axis" is not specific enough. What does the z-coordinate of that vector mean?

SoykaDaimler:
Thanks @AndreasVIRES for your questions. 
I clarified the "middle of the axis" description: As I understood it, the center of the reference frame is located in the middle of the axis with respect to all three coordinates x,y and z. So that should define the z component as well.
About the "projection to the street": I rewrote the whole description and definition. I understand the contour as an extension to the bounding box. In ground truth, the contour should be defined once in the beginning in the local object frame and should not change anymore (same as the bounding box). During runtime, one can compute the current state based on the position and orientation data. 
As sensor data, the polygon should describe the visible part of the contour.
So now everything is described in the local object frame and I clarified the difference between ground truth and sensor data. I hope this makes it more clear.

AndreasVIRES:
**Middle of axis** 
Sure, middle in x, y and z but under what circumstances? Usually the definition is "neutral load conditions" or similar. The axles are moving during simulation relatively to the rest of the vehicle, shall your vector reflect this behavior?

SoykaDaimler:
Ah yes, thanks! "Neutral load conditions" is what I meant. I'll update the file...

osi_object.proto
```diff
@@ -54,6 +54,15 @@ message Vehicle
...
     optional Identifier driver = 10;
+    
+    /// The vector pointing from the bounding box center point (base.position) to the middle (in x, y and z) 
+    /// of the rear axle under neutral load conditions.
+    optional Vector3d bbcenter_to_rear = 11;
+
+	/// The vector pointing from the bounding box center point (base.position) to the middle (in x, y and z) 
+	/// of the front axle under neutral load conditions.    		
+	optional Vector3d bbcenter_to_front = 12;
+
```
CarloVanDriestenBMW: > optional Vector3d relative_ego_origin = 6;

@SoykaDaimler and @AndreasVIRES 
The description above states, that this is the vector to the rear axle, if the car is a simulated car. This is now duplicated? We could consider removing it and replacing it by the two new definitions?

SoykaDaimler: I agree. In general we need these offsets, no matter if it is the ego car or not. So I would suggest removing the relative_ego_origin vector.

SoykaDaimler: What do other people think? Would replacing relative_ego_origin disrupt some OSI logic?
@nHirsenkorn @TimoHankeBMW Your input would be appreciated.
If you agree I could make the necessary changes.

nHirsenkorn: No, it would not disrupt some logic. If you need it we must add it. However it will delay the initial rollout, as more complexity to the simulator and to the sensor model framework is added. That is the reason why the initial framework has such a minimal number of coordinate systems and reference points.

SoykaDaimler: OK, so we replace relative_ego_origin vector then and use the two other offset vectors. I will make the adjustments in this PR

SoykaDaimler: One more question: in which coordinate system was relative_ego_origin defined? In world coordinates? Then it has to be updated each frame. Or in object coordinates (then it would be constant)? That would be consistent with our previous discussion, about the necessity to rotate relative_ego_origin from object to world frame.
In case we have an object frame, we should also define it in the wiki.

SoykaDaimler: I assumed object coordinates for now.

nHirsenkorn: object coordinates is correct

osi_object.proto
```diff
@@ -55,13 +51,9 @@ message Vehicle
...
     
-    /// The vector pointing from the bounding box center point (base.position) to the middle (in x, y and z) 
-    /// of the rear axle under neutral load conditions.
-    optional Vector3d bbcenter_to_rear = 11;
-
 	/// The vector pointing from the bounding box center point (base.position) to the middle (in x, y and z) 
-	/// of the front axle under neutral load conditions.    		
-	optional Vector3d bbcenter_to_front = 12;
+	/// of the front axle under neutral load conditions. In object coordinates.    		
+	optional Vector3d bbcenter_to_front = 11;
```
nHirsenkorn: please correct the identation

SoykaDaimler: Sorry, there seems to be some problem with tabs and spaces. I corrected it now

nHirsenkorn:
we can merge after a second review.

osi_object.proto
TimoHankeBMW: Replacing relative_ego_origin is fine, especially since it is only a rename and the addition of the front axis coordinate system now.
The (only) purpose of relative_ego_origin is/was to allow the transformation to a frame relative to the rear_axis_coordinate system of a vehicle (specifically ego-vehicles, that's why it wasn't required for all vehicles). That is still possible with the new definitions, so everything is perfectly fine there.

SoykaDaimler:
@nHirsenkorn I thought the versioning is done by the people that merge. Therefore I did not change anything there. So should I do this, or who will change it? Would we then be at 3.0.0?

TimoHankeBMW:
I think the comment for the base_polygon still needs some refinement.
Currently it states: "General definitions: The polygon is defined in the local object frame: x pointing forward and y to the left."
This defines the (x/y-) axes, but not the position of the origin for the purpose of the Vector2d-coordinates. Is "local object frame" relative to the center of the bounding box (i.e. position-coordinates of the vehicle, that would make most sense to me and this meaning is implied as far as I can see) or relative to the front or rear axis?
To avoid confusion, we should specifically include this in the comment (as we now have 3 possible alternatives for the origin of the vehicle frame).

SoykaDaimler:
@TimoHankeBMW 
Thanks for your comments! Yes, I would suggest to use the middle of the bounding box as origin for the object frame. I will add it to the description. 
In general, I think that all coordinate frames should be defined once in a central location instead of defining them over and over again for each field. See my issue #14

AndreasVIRES:
@SoykaDaimler I'm confused... I thought you need world coordinates for the polygon?
```
/// **The polygon is defined in the local object frame**: x pointing forward and y to the left.
/// The origin is the center of the bounding box.
```

SoykaDaimler:
@AndreasVIRES Initially that was my idea, but after our discussion that changed. Therefore I changed the contour to object coordinates. From there I'll calculate the world coordinates myself ;-)

nHirsenkorn:
@SoykaDaimler sounds reasonable to let the merger change the version; moreover it should be regarded as minor change. I added a Wiki article regarding the versioning and added it to the workflow.

Issue #54, opened 2017-03-03
Add "fog" to environment description


hschoen:
In osi-environment.proto precipitation is defined. However, fog cannot be described with mm/h. I suggest to add fog like this:

    /// Fog
    optional Fog fog = 5;

    /// Definition of fog states. Values need to be revised! See https://de.wikipedia.org/wiki/Nebel#Tr.C3.B6pfchengr.C3.B6.C3.9Fe_und_Nebeldichte Use visibility instead? Values for radar attenuation required!
    {
        /// Light intensity fog, when density is < 0.02 g/m³
        PRECIPITATION_FOG_LIGHT = 0;
  
        /// Medium intensity fog, when density is < 0.05 g/m³
        PRECIPITATION_FOG_MEDIUM = 1;
  
        /// High intensity fog, when density is < 0.1 g/m³
        PRECIPITATION_FOG_HIGH = 2;
  
        /// Violent intensity fog, when density is > 0.2 g/m³
        PRECIPITATION_FOG_VIOLENT = 3;  
    }

CarloVanDriestenBMW:
Yes, that would be OK for me, we should talk about that at the workshop and find solid references for both :)

CarloVanDriestenBMW:
The enum here is not defined correctly:

```
    /// Definition of precipitation states. See https://en.wikipedia.org/wiki/Rain#Measurement
    enum Precipitation
    {
        /// Light intensity rain, when the precipitation rate is < 2.5 mm per hour
        PRECIPITATION_RAIN_LIGHT = 0;
  
        /// Medium intensity rain, when the precipitation rate is between 2.5 mm and 10 mm per hour
        PRECIPITATION_RAIN_MEDIUM = 1;
  
        /// High intensity rain, when the precipitation rate is between 10 mm and 50 mm per hour
        PRECIPITATION_RAIN_HIGH = 2;
  
        /// Violent intensity rain, when the precipitation rate is > 50 mm per hour
        PRECIPITATION_RAIN_VIOLENT = 3;  
} 
```

@AhmedMzid and @PhilippRoethigBMW please find better definitions for fog and rain according to DIN standards e.g.

PhilippRoethigBMW:
@CarloVanDriestenBMW what about a rain and fog state similar to this one:

```cmd
    /// No rain, when the precipitation rate is < 0.1 mm per hour (values are not final)
    PRECIPITATION_RAIN_NONE = 2;
```

[...]

PR #59, opened 2017-03-30
Resolve issue #54 (fog definition)

PhilippRoethigBMW:
Changes made in osi_environment.proto:
- Created a new enum defining fog
- Redefined precipitation 

For additional information see issue #54


hschoen:
I have problems understanding the comment "must not be used in ground truth".
= not given in ground truth?
= not used in model?

CarloVanDriestenBMW:
I would like to add a Definition of "Visibility" and correct the intervals (equal is missing). I think Shepard or Collins depends on their definition of "visibility". Let's see :)

hschoen:
In the Radar Handbook I quoted visibility is defined as follows: "Visibility is defined as the greatest distance in a given direction at which it is just possible to see and identify with the unaided eye in the daytime a prominent dark object against the sky at the horizon and at night a known, preferably unfocused, moderately intense light source." 
Where can I buy a standard unaided eye an a  standard moderately intense light source? :)

PhilippRoethigBMW:
When it comes to visibility both Shepard and Perry are referring to the definition from the Meteorological Office. And they define visibility (similar to what you quoted from the Radar Handbook) as "The greatest distance at which a black object of suitable dimensions may be seen and recognized against the horizon, sky, or in the case of night observation, could be seen and recognized if the general illumination were raised to the normal daylight level."
However this definition is not only from 1969 but also depends on the estimation made by an human judging the appearance of distant objects.
In today’s standards "Visibility sensors measure the meteorological optical range which is defined as the length of atmosphere over which a beam of light travels before its luminous flux is reduced to 5% of its original value." (see the [homepage of the Met Office)](http://www.metoffice.gov.uk/guide/weather/observations-guide/how-we-measure-visibility)
Both definition might lead to the same results but I think we have to decide which definition is most suitable for us.

Collins' values might seem more significant to you because there are converted from yards into meters (e.g. 200 yards equals 180 meters). Also I am not sure which definition of visibility Collins is using.

hschoen:
@PhilippRoethigBMW This is valuable information - thanks!
I clearly prefer today's definition of visibility. However, when estimating visibility  from recorded camera data we are probably forced to use the older definition (How many guide posts can we see?).
I assume that environment simulations use today's definition of visibility in order to create fog.  I will ask @Andreas Biehn of Vires and @MHerrmannIPG about their approach.

AndreasVIRES:
For us, maximum visibility means that no object color is visible anymore, for objects at that distance and further. In other words, the object color has completely passed into the fog color no later than this distance.

MHerrmannIPG:
In CarMaker one option of the fog definition is similar to what @AndreasVIRES said. You can specify the distance where the object colour has completely blended into the fog colour. Fog starts at the location of the virtual camera position and the density/blending increases linearly until the distance specified.

CarloVanDriestenBMW:
Therefore it is "just" a visual definition of fog and applicable for cameras only? Additionally it is global? So 2 cameras would always have the same maximum view distance?

AndreasVIRES:
I don't exactly know for which kind of sensor models fog is applicable, but fog is currently defined with "visual range". 
A specific sensor can interpret this value as necessary, e.g. if there is no influence for a simple radar model, it might ignore the value, but for the raw image for a camera sensor model it's relatively clear, what the value means.
Since the fog is part of the environment and two or more cameras "live" in the same environment, both cameras will receive the same "visual range" value, yes.

CarloVanDriestenBMW:
Concluding the current implementations in VTD and Carmaker I think we could choose the newer definition:
> In today’s standards "Visibility sensors measure the meteorological optical range which is defined as the length of atmosphere over which a beam of light travels before its luminous flux is reduced to 5% of its original value." [Link](http://www.metoffice.gov.uk/guide/weather/observations-guide/how-we-measure-visibility)

I am sure that we can find a forumula calculating the "maximum visible distance" from the luminous flux of 5% for the human eye, which is the output of the simulation visualization?
Also adding the comment here from Met Office:
> In most instances this is approximately equivalent to, but not the same as, visibility measured by the contrast of a distant object against its background.

This way the visibility can be applied globally, because cameras of different qualities may have different maximum distances in which they can see a "black object's shape". Defining it the new way, the modelling of the maximum distance can be done in the sensor model. Much better in my opinion!

agree? @MHerrmannIPG @AndreasVIRES @hschoen

AndreasVIRES:
From my point of view it's ok to use this definition because it is only an environment definition, not the parametrization of the image generator(s). A sensor model could theoretically also request the image generator to produce an image with unlimited visibility and create the fog by itself.
For me the more important question would be if effectively 6 enum values for fog with up to 30km bandwidth in the "visibility" are enough.

AndreasVIRES:
The StVO value of 50m is also missing: "thick" is 40...200m, rear fog light on or off?

CarloVanDriestenBMW:
@AndreasVIRES we have adjusted the value for "thick" fog according to the STVO

Issue #69, opened 2017-04-24
Rename Polar3d to Spherical3d


PhilippRoethigBMW:
@StefanBernsteiner from Magna Steyr Engineering proposed a renaming of the message Polar3d to Spherical3d. Because the way this message is defined in osi_common.proto references to spherical rather than to polar coordinates. Since Polar3d is defined with two angels and a radial distance.
With this renaming we could prevent some future misunderstandings.

PR #78, opened 2017-07-13
change the order of width and length (resolves #77)

### LINKED ISSUE #77: Order of width and length in Dimension3d
Each object (vehicle, MovingObject or StationaryObject) defines its own right-handed coordinate system which conventionally use the order x,y,z.
The message Orientation3d, for instance, is defined according to this convention: roll first (rotation around x), pitch second (rotation around y), and yaw third (rotation around z).
However, the same rule does not apply for Dimension3d: According to our definition of the coordinate system (see [this example of a vehicle coordinate system](https://en.wikipedia.org/wiki/Axes_conventions#/media/File:RPY_angles_of_cars.png)), width would  natively be measured along the y-axis and length along the x-axis. The current definition of the message Dimension3d would hence correspond to a different order y,x,z.
That's why I would suggest to change the order of the attributes for the next release:
```
message Dimension3d

{

optional double length = 1;

optional double width = 2;

optional double height = 3;

}
```

AhmedMzid:
For further details see issue #77

### DIFF
--- a/osi_common.proto
+++ b/osi_common.proto
@@ -65,13 +65,13 @@ message Timestamp
 //
 message Dimension3d
 {
-    // The width of the bounding box.
-    //
-    optional double width = 1;
-    
     // The length of the bounding box.
     //
-    optional double length = 2;
+    optional double length = 1;
+    
+    // The width of the bounding box.
+    //
+    optional double width = 2;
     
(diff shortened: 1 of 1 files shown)


CarloVanDriestenBMW:
Could you also resolve #69 please?

Issue #84, opened 2017-09-08
Lane driving direction


solosito:
- Is the supposed driving direction indicated somewhere? I haven't seen it.
- In case it doesn't, why not?
- In case it wasn't considered, would it makes sense to add it?

It could be added to the OSI::Lane, something like this:

```
message Lane
{
    // The rest of the Lane message
    //
    {...}

    // A lane direction describes the supposed driving direction of the lane.
    //
    optional LaneDrivingDirection lane_driving_direction = 10;
    
    // Definition of available lane driving direction types.
    //
    enum LaneDrivingDirection
    {
        // Lane of unknown driving direction (must not be used in ground truth).
        //
        LANE_DRIVING_DIRECTION_UNKNOWN = 0;
        
        // Other (unspecified but known) lane driving direction.
        //
        LANE_DRIVING_DIRECTION_OTHER = 1;
        
        // The lane driving direction is right.
        //
        LANE_DRIVING_DIRECTION_RIGHT_WAY = 2;
        
        // The lane driving direction is wrong.
        //
        LANE_DRIVING_DIRECTION_WRONG_WAY = 3;
    }
}
```

CarloVanDriestenBMW:
Thank you for your suggestion. We will investigate it!

JonathanSchmalhoferBMW:
In [Issue 80](https://github.com/OpenSimulationInterface/open-simulation-interface/issues/80) I assumed that the driving direction is implicitly defined by the direction of the ascending order of LaneBoundary-Points and CenterLine-Points in their respective array.

This is of course not mandatory and makes conversion from OSI-Messages for right-hand-traffic to left-hand-traffic complicated. A flag as proposed by @solosito would therefore be much easier to handle.

solosito:
I have been thinking about this matter and how to define the lane driving direction. The problem is that the lane driving direction it is relative data and not absolute. It depends on the point of view of the observer. A first approximation led me to think about defining it regarding the sensor mounting data, but this assumption would be wrong because it could not be used in the ground truth due to its relativity.

From my point of view, the only plausible solution would be defining it respecting the direction of the vector defining the lane's center line. This would make much easier to handle right-hand and left-hand   traffic systems as point by @JonathanSchmalhoferBMW.

CarloVanDriestenBMW:
The driving direction must be the "intended driving direction" in the GT. I do not think it should be relative to the ego vehicle.

[...]

PR #85, opened 2017-09-09
Add lane driving direction to the message lane (#84)

solosito:
This is the first approximation to the lane driving direction. It is not ready to be merged. First has to be analysed and discussed as described in the issue #84

### DIFF
--- a/osi_lane.proto
+++ b/osi_lane.proto
@@ -83,6 +83,11 @@ message Lane
     //
     optional LaneFraming lane_framing = 9;
 
+    // A lane driving direction describes the driving direction of the lane.
+    // The lane driving direction is defined regarding the direction of the vector defining the center line of the lane.
+    //
+    optional LaneDrivingDirection lane_driving_direction = 10;
+
     // Definition of available lane types.
     //
     enum Type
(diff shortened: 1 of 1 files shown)


CarloVanDriestenBMW:
Thank you for your contribution. I would like to take some time for a review and discuss it with the tool developer.

CarloVanDriestenBMW:
> This is the line, that a typical vehicle is supposed to follow. Thus, if the road narrows (e.g. construction site) this line is supposed to describe the trajectory that the vehicle should follow. **The intended direction of travel on the lane is given by the direction defined by the sequence of points forming the center line.**

This comment in osi:lane center_lane has to be updated then?

solosito:
> This comment in osi:lane center_lane has to be updated then?

Actually yes. It should specify that the intended driving direction it is not described by the center_line but by the lane_driving_direction.

CarloVanDriestenBMW:
Hi @solosito ,
we considered your feedback and will introduce a lane driving direction in the OSI 2.3.0 release on Jan. 26th. We would consider using a vector3d in order to describe the direction. Would this be suitable for you?
Best regards
Carlo

solosito:
Hallo @CarloVanDriestenBMW ,
I think it is a good idea using a vector3d rather than using a flag. The driving direction could be defined more precisely.
Best regards,
Alfonso.

AndreHildebrandt:
Hello @CarloVanDriestenBMW,

I think, that a Vector3D alone would not be sufficient. One would need either a second vector3D or an orientation3D IMHO.
Example: A hairpin curve. The IntendedDrivingDirection is then depending on the position of the vector3D. One needs a point of reference and a driving direction.

Best wishes,
Andre Hildebrandt

solosito:
Hi again,

then the initial suggestion of having a flag is a very good option. There is no need to have a tangential vector of the curve to know which is the intended driving direction as the curve itself contains that information. It is only necessary to know whether the intended driving direction follows the points' storing order or its reverse. 

Thus, in case of conversions from left- to right-hand traffic and vice-versa, it would be as simple as changing this flag to the proper value.
___
@AndreHildebrandt 
> Apart from that, there has been the suggestion to replace the successor and predecessor lane ids by LaneConnectionPair ids. Do they implicitly define the intended driving direction?

 No. What would be the intended driving direction in case of having just a single lane with a permanent id?

carsten-kuebler:
At the OSI workshop in Ulm (2017) there was a discussion to add a driving direction. But that does't suit as @AndreHildebrandt mentioned. Therefore you have to specify start point and driving direction or a flag as proposed from @solosito. 
Bidirectional lanes or small lanes require 2 center lines and there is no left and right. So I proposed #110.

CarloVanDriestenBMW:
@solosito is the following solution ok for you? You can find it in #110

`optional bool center_line_is_driving_direction = 4;`

CarloVanDriestenBMW:
@pmai there are some new comments in #110 - I will close the issue and you can merge the comments.

Issue #75, opened 2017-06-23
Unit of measurement for Ambient Illumination


ndunningbmw:
We should try to assign a scientific unit to the Ambient Illumination (lux for example), rather than the loosely defined ambient illumination index.

CarloVanDriestenBMW:
Can you make a proposal @ndunningbmw ?

[Maybe like this?](https://www.noao.edu/education/QLTkit/ACTIVITY_Documents/Safety/LightLevels_outdoor+indoor.pdf)

ndunningbmw:
Yes the provided suggestion would be good, as it also provides a standardized value calculation.

Is the value for Ambient Light calculated at the current position where the Ground Truth is filled? I think this is logical but it does not state anywhere in the osi_environment file that this is the case.

Sheema60:
I have added a short report of some research on the definition, units and classification of ambient lighting. For the classification I decided to stick to measurements based on illuminance as that appears to be more generally used. I also made an initial classification and added it to the osi_environment.proto file as a first step. Feedback and comments would be appreciated. 


[Ambient Lighting_part1.pdf](https://github.com/OpenSimulationInterface/open-simulation-interface/files/1306489/Ambient.Lighting_part1.pdf)

ndunningbmw:
Thanks for the update; as far as my use case is concerned, the proposed update to the osi environment would be a good and acceptable solution.

CarloVanDriestenBMW:
@AhmedMzid could you please check the proposal, so that we can discuss it tomorrow?

Sheema60:
I would like to remove the Media College reference as their copyright policies are a little ambiguous and in any case the wording for the definition of ambient light is our own and does not rely solely on this website. I'll make another push. I will leave the report as is, because it is just a source of information in case anyone needs it.

Sheema60:
I added one level and pushed the change. Here is the
[Ambient Lighting_part1_edited.pdf](https://github.com/OpenSimulationInterface/open-simulation-interface/files/1318493/Ambient.Lighting_part1_edited.pdf)
 edited report.

PR #87, opened 2017-09-14
AFTMMODELS-143 - define ambient illumination

Sheema60:
Definition of ambient illumination from various sources plus the SI units.
Categorization of various ambient illumination levels.

### DIFF
--- a/osi_environment.proto
+++ b/osi_environment.proto
@@ -111,18 +111,16 @@ message EnvironmentalConditions
 
     // Definition of ambient illumination states.
     // Ambient light is any light in the vehicle's environment not emitted by the vehicle 
-    // itself [5][6]. It can include sun/moon light, light from other cars, street lights, etc.
+    // itself [5]. It can include sun/moon light, light from other cars, street lights, etc.
     // 
     // Lux is the SI unit of luminance or illumination of a one metre square area, which is 
-    // equal to one lumen per square meter[7][8]. 1lx = 1lm/m²
+    // equal to one lumen per square meter[6]. 1lx = 1lm/m²
     // The lumen (symbol: lm) is the SI derived unit of luminous flux, a measure of the total
     // quantity of visible light emitted by a source. The lumen is defined in relation to the candela as 1lm=1cd.sr
     //
     // \par References:
-    // [5] http://www.mediacollege.com/lighting/ambient/
-    // [6] https://en.wikipedia.org/wiki/Available_light 
-    // [7] https://physics.nist.gov/cuu/Units/units.html 
-    // [8] https://en.wikipedia.org/wiki/Lux
+    // [5] http://www.mediacollege.com/lighting/ambient/ 
+    // [6] https://physics.nist.gov/cuu/Units/units.html 
     enum AmbientIllumination
     {
         // Ambient illumination is unknown (must not be used in ground truth).
--- a/osi_environment.proto
+++ b/osi_environment.proto
@@ -134,7 +134,7 @@ message EnvironmentalConditions
         // Other (unspecified but known) ambient illumination.
         //
         AMBIENT_ILLUMINATION_OTHER = 1;
-        // Illumination between [0.001, 0.01[ Lux is Level 1.
+        // Illumination between ]0.001, 0.01[ Lux is Level 1.
         // E.g. Night with no artificial light.
         AMBIENT_ILLUMINATION_LEVEL1 = 2;
         // Illumination between [0.01, 1[ Lux is Level 2.
--- a/osi_environment.proto
+++ b/osi_environment.proto
@@ -110,18 +110,19 @@ message EnvironmentalConditions
     }
 
     // Definition of ambient illumination states.
-    //Ambient light is any light in the vehicle's environment not emitted by the vehicle 
-    //itself. It can include sun/moon light, light from other cars, street lights, etc.
-    //http://www.mediacollege.com/lighting/ambient/
-    //https://en.wikipedia.org/wiki/Available_light
+    // Ambient light is any light in the vehicle's environment not emitted by the vehicle 
+    // itself [5][6]. It can include sun/moon light, light from other cars, street lights, etc.
+    // 
+    // Lux is the SI unit of luminance or illumination of a one metre square area, which is 
+    // equal to one lumen per square meter[7][8]. 1lx = 1lm/m²
+    // The lumen (symbol: lm) is the SI derived unit of luminous flux, a measure of the total
+    // quantity of visible light emitted by a source. The lumen is defined in relation to the candela as 1lm=1cd.sr
     //
-    //Lux is the SI unit of luminance or illumination of a one metre square area, which is 
-    //equal to one lumen per square meter. 1lx = 1lm/m²
-    //The lumen (symbol: lm) is the SI derived unit of luminous flux, a measure of the total
-    //quantity of visible light emitted by a source. 
-    //The lumen is defined in relation to the candela as 1lm=1cd.sr
-    //https://physics.nist.gov/cuu/Units/units.html
-    //https://en.wikipedia.org/wiki/Lux
+    // \par References:
+    // [5] http://www.mediacollege.com/lighting/ambient/
+    // [6] https://en.wikipedia.org/wiki/Available_light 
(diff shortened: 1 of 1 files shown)


Sheema60:
Related issue ist #75

Sheema60:
I have added a short report of some research on the definition, units and classification of ambient lighting. For the classification I decided to stick to measurements based on illuminance as that appears to be more generally used. I also made an initial classification and added it to the osi_environment.proto file as a first step. Feedback and comments would be appreciated.

[Ambient Lighting_part1.pdf](https://github.com/OpenSimulationInterface/open-simulation-interface/files/1306543/Ambient.Lighting_part1.pdf)

CarloVanDriestenBMW:
Thx! @Sheema60 have you checked, if it is correctly displayed in the doxygen document? The sectioning of the comment differs in style from the "fog" comment for example?

CarloVanDriestenBMW:
I will check the content in detail, but it is very important not to use Wikipedia as a reference. Please always check original sources.

Sheema60:
@CarloVanDriestenBMW I added the wikipedia references just for the convenience of the user as it is a source for light and quick reading. Will remove it now.

osi_environment.proto
```diff
@@ -119,18 +128,31 @@ message EnvironmentalConditions
...
+        // Illumination between [3, 10[ Lux is Level 4.
+        // E.g. Average to full twilight / Minimum lighting on collector roads /
+        // Minimum lighting on major and expressway roads with low to average pedestrian conflict.
+        AMBIENT_ILLUMINATION_LEVEL4 = 5;
+        // Illumination between [10, 20[ Lux is Level 5.
+        // E.g. Minimum lighting on major and expressway roads with high pedestrian conflict.
+        AMBIENT_ILLUMINATION_LEVEL5 = 6;
+        // Illumination between [20, 1000[ Lux is Level 6.
+        // E.g. Roads with more lighting / Dark or overcast day.
+        AMBIENT_ILLUMINATION_LEVEL6 = 7;
```
CarloVanDriestenBMW: We would suggest to include another level here. The 400 lux limit would correspond to a sunrise on a clear day which might be relevant for a camera model.

osi_environment.proto
```diff
@@ -119,18 +128,31 @@ message EnvironmentalConditions
...
+        // Illumination between [1000, 10000[ Lux is Level 7.
+        // E.g. Average to full daylight.
+        AMBIENT_ILLUMINATION_LEVEL7 = 8;
+        // Illumination between [10000, 120000[ Lux is Level 8.
+        // E.g. Full daylight to intense sunlight.
+        AMBIENT_ILLUMINATION_LEVEL8 = 9;
```
CarloVanDriestenBMW: The defined levels seem to fit quite well for the purpose!

osi_environment.proto
```diff
@@ -119,18 +128,31 @@ message EnvironmentalConditions
...
-        AMBIENT_ILLUMINATION_MEDIUM = 3;
-        // Ambient illumination index is "low".
-        // E.g. during twilight (dawn/dusk) or in a tunnel.
-        AMBIENT_ILLUMINATION_LOW = 4;
-        // Ambient illumination index is "dark".
-        // E.g. night time, no sunlight.
-        AMBIENT_ILLUMINATION_DARK = 5;
+        // Illumination between [0.001, 0.01[ Lux is Level 1.
+        // E.g. Night with no artificial light.
+        AMBIENT_ILLUMINATION_LEVEL1 = 2;
```
CarloVanDriestenBMW: I am not sure about the naming here:
The meaning of LevelXY is somehow not easy to grasp... but I understand that it is hard to find a more suitable naming convention...

osi_environment.proto
```diff
@@ -110,7 +110,16 @@ message EnvironmentalConditions
...
+    // Ambient light is any light in the vehicle's environment not emitted by the vehicle 
+    // itself. It can include sun/moon light, light from other cars, street lights, etc.
+    // 
+    // Lux is the SI unit of luminance or illumination of a one metre square area, which is 
+    // equal to one lumen per square meter[5]. 1lx = 1lm/m²
+    // The lumen (symbol: lm) is the SI derived unit of luminous flux, a measure of the total
+    // quantity of visible light emitted by a source. The lumen is defined in relation to the candela as 1lm=1cd.sr
     //
+    // \par Reference:
+    // [5] [The NIST Reference on Constants, Units, and Uncertainty](https://physics.nist.gov/cuu/Units/units.html)
```
CarloVanDriestenBMW: The references for the "how/why and from where did I pick the limits for the different levels are missing. The PDF in the issue is readable in Github, but can not be accessed by a user of OSI.

Sheema60: This is just an initial classification. If it is acceptable I can add the references. These would be the European, and North American standars as the classification is based mostly on lighting standards for roads.

osi_environment.proto
Sheema60: I can split this category into a [20, 400] lux level and a [400 to 1000] lux level.

osi_environment.proto
```diff
@@ -110,7 +110,22 @@ message EnvironmentalConditions
...
+    // The lumen (symbol: lm) is the SI derived unit of luminous flux, a measure of the total
+    // quantity of visible light emitted by a source. The lumen is defined in relation to the candela as 1lm=1cd.sr
     //
+    //Categorization is done based on natural day/night time illuminance levels [6] and standards for required lighting levels on roads [6, 7, 8, 9]. 
+    //
+    // \par References:
+    // [5] [The NIST Reference on Constants, Units, and Uncertainty](https://physics.nist.gov/cuu/Units/units.html)
+    // [6] [National Optical Astronomy Observatory](https://www.noao.edu/education/QLTkit/ACTIVITY_Documents/Safety/LightLevels_outdoor+indoor.pdf)
+    // [7] [Standards for required street lighting in the USA](http://www.wsdot.wa.gov/research/reports/fullreports/847.1.pdf)
+    // [8] [Canadian IES-RP-8 standards for road lighting - municipality of Saint-Gédéon-de-. Beauce](http://sslnet.ca/wp-content/uploads/2011/10/LTE-RT-2011-0076-Anglais.pdf)
```
CarloVanDriestenBMW: Have you checked the doxygen output here?
It seems ok, but I would recommend to compile it once

osi_environment.proto
```diff
@@ -110,7 +110,22 @@ message EnvironmentalConditions
     }
 
     // Definition of ambient illumination states.
+    // Ambient light is any light in the vehicle's environment not emitted by the vehicle 
+    // itself. It can include sun/moon light, light from other cars, street lights, etc.
+    // 
+    // Lux is the SI unit of luminance or illumination of a one metre square area, which is 
+    // equal to one lumen per square meter[5]. 1lx = 1lm/m²
+    // The lumen (symbol: lm) is the SI derived unit of luminous flux, a measure of the total
+    // quantity of visible light emitted by a source. The lumen is defined in relation to the candela as 1lm=1cd.sr
```
CarloVanDriestenBMW: 1lm=1cd * sr
Define sr

osi_environment.proto
```diff
@@ -119,18 +134,34 @@ message EnvironmentalConditions
...
-        // Ambient illumination index is "medium".
-        // E.g. daytime with overcast sky.
-        AMBIENT_ILLUMINATION_MEDIUM = 3;
-        // Ambient illumination index is "low".
-        // E.g. during twilight (dawn/dusk) or in a tunnel.
-        AMBIENT_ILLUMINATION_LOW = 4;
-        // Ambient illumination index is "dark".
-        // E.g. night time, no sunlight.
-        AMBIENT_ILLUMINATION_DARK = 5;
+        // Illumination between [0.001, 0.01[ Lux is Level 1.
```
CarloVanDriestenBMW: ]0.001

CarloVanDriestenBMW:
after your final check, please edit the commit message so that it states (resolves #75) at the end. This will then automatically close the issue when accepting the pull request. The Jira ticket must not be included, because there is no linking to it. The link to the Github issue in Jira as a comment is enough. Thank you for this very good proposal!

osi_environment.proto
CarloVanDriestenBMW: @Sheema60 I think the interval needs to be open here?

Issue #86, opened 2017-09-11
Definition of Timestamp


AhmedMzid:
The current definition of timestamp in GroundTruth does not precise how the zero time must be defined. In this way, it can be different from the starting time point of the simulation. 
However, the zero time point of Timestamp in GT must - in certain circumstances - coincide with the starting time of the simulation in order to be able to reproduce the real behavior of sensor models (latency, update rate..). The problem occurs especially when a particular part of a scenario should be resimulated without replaying the complete scenario.
The need for a further specification on the definition of Timestamp in GT should be discussed.

haoyuanying:
A possible solution can be found at the document. 
[zerotime solution.pdf](https://github.com/OpenSimulationInterface/open-simulation-interface/files/1382566/zerotime.solution.pdf)

CarloVanDriestenBMW:
solved by #130

PR #90, opened 2017-10-13
solution for issue #86

haoyuanying:
The issue #86 is solved. The detailed solution can be found in the attachment in the comments.


CarloVanDriestenBMW:
Different timestamps in OSI:

Environment: Possible change to GPS time because of synchronization with DGPS in measurement run
Note "Schaltsekunde" in doxygen

Simulation time:
Needed: Sensor Start Offset in configuration which is always calculated from 0 on 
Simulation Starttime needs a proper definition
tbd: Reference a valid definition for Sceanario?

CarloVanDriestenBMW:
add field simulation_starttime

pmai:
I have added the relevant fields (offset and start time) to the configuration messages in PR #120. It is unclear to me whether we should conflate the idea of multi-scenario runs with simulation start and stop-times, since ideally sensor models should not be aware of a concept like scenario...

CarloVanDriestenBMW:
@pmai yes I think the sensor model should not be aware of the scenario concept. You can close this pull-request with the merge of #120

haoyuanying:
@pmai @CarloVanDriestenBMW I would say to have such a function actually turn on the possibility to actively simulate the fault cases in sensors, e.g. at a certain period of the time, due to environment / unknown conditions, the sensor is temperorilly out of order...

pmai:
@haoyuanying that would require hard-coding/configuring the times into the sensor models for each scenario, which I think is not a good idea. This should IMHO be handled either through Special fault-injection models (which will have their own Parameters), or, if needed, flags into the sensor model to enable fault, etc., which would then be triggered by the scenario/simulation, right?

haoyuanying:
@pmai Fault injection is one example. The idea to have the field also aims to help to crop a long scenario into sevaral smaller ones. I am not sure if to insert simulation start/ent time is easier than to re-create scenarios for simulator makers...

pmai:
But what is the sensor model supposed to do with these times? Anything that depends on scenario data should IMHO not be known to the sensor model?

haoyuanying:
@pmai In this case to separate these kind of information out of GT, do you think that it is valuable to make another field in OSI as SimulationConfiguration / SimulationControl?

pmai:
@haoyuanying The question would be from where to where would these messages be communicated? I.e. who sends it, who receives it for what purpose? I think a number of extensions of OSI to cover more than the Environment Simulation -> Sensor -> Function use case would be beneficial (e.g. Function <-> Driving Dynamics <-> Environment Simulation), however I have not seen them tackled for now... Daimler would be fairly interested in these additions as well, as far as I gathered at the Ulm Workshop...

haoyuanying:
@pmai I would also raise the question like this: if you use OSI only for data transfer, or you could use it for simulation toochain flow control (e.g. sensor feedback) as well, exactly like Function <-> Driving Dynamics <-> Environment Simulation...

CarloVanDriestenBMW:
@pmai and @haoyuanying maybe we postpone this discussion to the next Pegasus meeting? I removed the milestone and we keep the pullrequest open.

pmai:
@carsten-kuebler isn't this PR dead, i.e. we have the start time already in the sensor input configuration messages, and generally speaking we should be moving stuff that is static into a seperate setup message?

CarloVanDriestenBMW:
@pmai I agree and will close this PR

PR #96, opened 2017-11-15
Update LICENSE

CarloVanDriestenBMW:


CarloVanDriestenBMW:
The idea is to update the licence file in a way the licence is properly displayed in the top right corner next to the contributers. I am not sure if I have to delete the file for that?

Issue #89, opened 2017-10-09
ENABLE-S3: Ego Vehicle related data


HannesSchneiderAVL:
For precise sensor model simulation some properties of the Ego Vehicle will be relevant. This information can be calculated in a vehicle dynamics simulation and can be sent back to the simulation environment.
(in our use case: VTD -> Model.CONNECT (VSM || Carmaker) --- (ego vehicle dynamics) -> VTD)
A message in OSI containing this information would be helpful for a precise sensor simulation:

For the sensor model following properties would be relevant:

VehicleMovement              boolean
BreakOn                             boolean
HandBreakOn                     boolean

WheelVelocityFrontLeft
WheelVelocityFrontRight
WheelVelocityRearLeft
WheelVelocityRearRight         m/s        1/10

VehicleVelocity                       kp/h       1/100

Pitch
Yaw
Roll                                         degrees

PitchVelocity
YawVelocity
RollVelocity                             degrees/s

PitchAccelleration
YawAccelleration
RollAccelleration                      degrees/s2

ShockAbsorbersImmersionDepthFrontLeft
ShockAbsorbersImmersionDepthFrontRight
ShockAbsorbersImmersionDepthRearLeft
ShockAbsorbersImmersionDepthRearRight                    mm

TirePressureFrontLeft
TirePressureFrontRight
TirePressureRearLeft
TirePressureRearRight                                       Bar 1/100

haoyuanying:
Hello,

Thanks for the suggestions. In OSI::SensorData, all the vehicle / objects are converted to DetectedObject and the Ego Vehicle itself is removed from the DetectedObject. The only way to determine the Ego Vehicle is to check the ego_vehicle_id in SensorData (it is the ID that the vehicle in GroundTruth), or recursively check the ego_vehicle boolean of each vehicle in SensorDataGroundTruth. To have the Ego Vehicle information in SensorData, I would propose to have a redundant Ego Vehicle copy in SensorData, that would improve the efficiency to find Ego Vehicle in SensorData by direct accessing instead recursive searching in SensorDataGroundTruth.

Since the vehicle information in SensorData are relative to the Ego Vehicle in GroundTruth, I would propose to have the absolute information here for further calculation in other simulator.

The boolean of movement of the vehicle is very useful in general, especially in SensorData. The velocity and orientation in different units are a bit redundant, is there an issue that to make the conversion outside OSI in your toolchain? 

Regarding to the other relevant information, 

[...]

PR #101, opened 2017-11-24
issue #89 Ego Vehicle in SensorData, Vehicle has is_moving flag

haoyuanying:
Solution for #89

### DIFF
--- a/osi_object.proto
+++ b/osi_object.proto
@@ -62,6 +62,10 @@ message Vehicle
     // of the front axle under neutral load conditions. In object coordinates.
     optional Vector3d bbcenter_to_front = 11;
     
+    // The flag indicates that if the vehicle is moving. In SensorData, if the vehicle has the same velocity as the ego vehicle, it is hard to 
+    // decide if the vehicle is moving or not. Also, the flag can be used to indicate a vehicle in parking mode. Currently, the vehicle is inherited 
+    // from BaseMoving, and a parked vehicle techincally should be BaseStationary, but it is impossible to switch at the moment.
+    optional bool is_moving = 12;
 
     // Definition of vehicle types.
     //
--- a/osi_sensordata.proto
+++ b/osi_sensordata.proto
@@ -9,6 +9,7 @@ import "osi_detectedobject.proto";
 import "osi_detectedoccupant.proto";
 import "osi_groundtruth.proto";
 import "osi_modelinternal.proto";
+import "osi_object.proto";
 
 package osi;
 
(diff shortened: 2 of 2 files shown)


pmai:
I'm in the process of merging these changes... Regarding the is_moving flag, would it not be more logical to have an is_stationary flag, since the default for the missing field would be false, which is always valid (i.e. treating a stationary vehicle as moving is probably always more valid than vice versa)?

If so, I'll incorporate this into a new patchset, in order to avoid having to merge the merge commit...

osi_object.proto
```diff
@@ -62,6 +62,10 @@ message Vehicle
     // of the front axle under neutral load conditions. In object coordinates.
     optional Vector3d bbcenter_to_front = 11;
     
+    // The flag indicates that if the vehicle is moving. In SensorData, if the vehicle has the same velocity as the ego vehicle, it is hard to
```
pmai: Besides the switch to the logic needed (i.e. we need is_stationary instead of is_moving so that the default value is less dangerous), while trying to add this field, I stumbled upon the comment that in SensorData the relative velocity makes detection of a stationary vehicle more problematic. However in SensorData, no Vehicles are present, they are all in GroundTruth only.  Was sensor-relative ground truth meant with this comment? I've separated this change out to a new separate PR #121, since it is unrelated to the Ego-Vehicle duplication code...

osi_sensordata.proto
```diff
@@ -79,6 +80,9 @@ message SensorData
     // \note Should not be used by subscribers to SensorData data. 
     // Generally this field should be cleared after internal processing.
     optional ModelInternal model_internal = 12;
+
```
pmai: I'm still somewhat confused by the need for this in SensorData: Since the input of the sensor model will not have any data in SensorData, this would not help the Sensor Model, rather it would fall to the sensor model to fill this information from GroundTruth information, and the consumer would be the recipient of sensor data, and I think those functions would rather have more driving dynamics information than is contained in GroundTruth? @CarloVanDriestenBMW ?

haoyuanying: The original idea to have a Ego Vehicle in SensorData is to model a GPS sensor. The noise from GPS influences the Ego Vehicle. If we do not have the Ego Vehicle in SensorData, then the noise of a GPS sensor would be added up to other sensors output. Then, a possible solution is to separate a new field in the SensorData to indicate the Ego Vehicle noise (offset), in this case a Ego Vehicle in SensorData is unnecessary. @CarloVanDriestenBMW

pmai: I see, in that case I agree that adding a new object for sensed ego vehicle data makes more sense (sensor might not only rely on GPS, but e.g. map data and distances to landmarks, etc., and might also include sensed environment data, like e.g. visibility, etc.). Probably makes sense to introduce this as a new branch & pull request, so I can close this one?

CarloVanDriestenBMW: @pmai yes close that one and make a consolidated branch

pmai:
Closing pull request since GPS as sensor needs to be handled in seperate PR.

Issue #62, opened 2017-03-31
Traffic Signs: What do we need?


CarloVanDriestenBMW:
What kind of traffic signs do we need? What signs does a simulation deliver? What format do we need - Traffic signs are different everywhere (GB, Ger, US, EU in general)? Is there a comparable sign in every region describing the same content? How is this Issue solved in OpenDrive (only valid for Germany?).

AndreasVIRES:
OpenDRIVE allows defining a country/type/subtype combination for every signal. For Germany usually the StVO sign numbers are used.

E.g. StVO signal 274-56: 
![](https://upload.wikimedia.org/wikipedia/commons/thumb/2/2e/Zeichen_274-56.svg/120px-Zeichen_274-56.svg.png)
start of a speed limit (274) with a limit of 60km/h (56) in OpenDRIVE is a signal with 
country="DEU",
type="274" and
subtype="56".

carsten-kuebler:
To improve country specific traffic signs and other country specifics, I propose to add a country code regarding to ISO 3166/1. It was discussed at the OSI workshop 2017 in Ulm.

PR #108, opened 2018-01-17
Add Country code to GT (Traffic Signs: What do we need? issue #62)

carsten-kuebler:
To improve country specific traffic signs and other country specifics, I propose to add a country code regarding to ISO 3166/1 in GT. It was discussed at the OSI workshop 2017 in Ulm.


CarloVanDriestenBMW:
@carsten-kuebler build fails

CarloVanDriestenBMW:
@carsten-kuebler build fails

carsten-kuebler:
Tabs replaced with spaces

Issue #105, opened 2018-01-16
Restructuring LowLevelData regarding standardisation activities for real sensor interfaces


carsten-kuebler:
There are activities to standardise real sensor interfaces for sensor detections. Detections are basic processed features from raw sensor data as osi:LidarPointCloud and osi:RadarReflectionList.
To differentiate between sensor raw data (e.g. camera - image) and detection data (e.g. camera - line, corner), I propose to rename the message osi:LowLevelData, restructure Lidar/Radar and add attributes, which are relevant for real sensors. 
The aim is to use osi sensors and real sensors in a similar way.

PR #109, opened 2018-01-17
Restructuring of low level data (Restructuring LowLevelData regarding standardisation activities for real sensor interfaces - Issue #105)

carsten-kuebler:
New OSI messages which represent real sensor outputs. Real sensor outputs will be standardised in future. Changes break downward compatibility.


CarloVanDriestenBMW:
@carsten-kuebler build failes

carsten-kuebler:
@CarloVanDriestenBMW reason: renaming of files. As discussed we renamed lowleveldata to featuredata.

carsten-kuebler:
Link pull request to issue #105

carsten-kuebler:
@CarloVanDriestenBMW @pmai sorry. An other python script had to be changed. Now the CI-tests are ok.

osi_modelinternal.proto
```diff
@@ -22,7 +22,7 @@ message ModelInternal
     //
     // Low Level data should be transferred to receiving components by means of a separate data stream. Therefore they
     // are not part the top level SensorData message, as they are used only internally by some models.
-    optional LowLevelData low_level_data = 1;
+    optional FeatureData feature_data = 1;
```
pmai: If we want to at least try to stay backward-compatible to a limited degree (otherwise we need to bump the version number to 3.0.0, I think, which might be needed due to the lane changes in any case), then the field ID of the new feature data should be 2, and the old 1 field should be retained as reserved, e.g.

    reserved 1;
    optional FeatureData feature_data = 2;

osi_featuredata.proto
```diff
@@ -0,0 +1,260 @@
...
+
+//
+// \brief Interface for sensor data, in contrast to interpreted data after object hypothesis and tracking.
+//
+// All information regarding the environment is given with respect to the sensor coordinate system specified in \c DetectionHeader::mounting_position. 
+// When simulating multiple sensors, each sensor has an individual copy of FeatureData in its own reference frame. This allows an independent treatment of the sensors.
+//
+message FeatureData
+{
+     // Lidar detection list for multiple lidar sensors.
```
pmai: The indentation in this file is slightly garbled (switches between 5, 4 (the proper amount) or 3 spaces a number of times).

osi_featuredata.proto
```diff
@@ -0,0 +1,260 @@
...
+     //
+     repeated DetectionRadarList detection_radar_list = 2;
+}
+
+//
+// \brief The header attributes of each detection.
+//
+message DetectionHeader
+{
+   // Version number of detecion list messages.
```
pmai: I think this needs more documentation, since it does not become clear the version of what is meant here (in the OSI context, we already have the OSI version number completely describing the format, so without further comments the intended meaning does not get clear enough I think).

osi_featuredata.proto
```diff
@@ -0,0 +1,260 @@
...
+       // Data is not available.
+       //
+       DATA_NOT_AVAILABLE = 2;
+   }
+}
+
+//
+// \brief A list of lidar detection.
+//
+message DetectionLidarList
```
pmai: Similarly to the renaming in PR #112, while I think a good naming convention is useful, I'd really prefer names like LidarDetection and LidarDetectionList to DetectionLidar and DetectionLidarList, since it puts the most specific information in front. Not a big issue, and if the naming is derived from VDA stuff, I can definitely live with that, but from a readability point of view I'd prefer the prefix variant.

osi_featuredata.proto
```diff
@@ -0,0 +1,260 @@
...
+{
+    // List of lidar detections.
+    //
+    repeated DetectionLidar detection = 1;
+    
+   // Header attributes of lidar detection.
+   //
+   optional DetectionHeader header = 2;
```
pmai: Is there a reason that the header is not in front of the list? I can see how this can make sense in embedded network communication (you can generate the header information on the fly while transmitting the list items), however even then I'd call it a footer or at least not a header. For OSI, I'd think placing the header as the first field makes much more sense (we cannot decode partial information in any case with current technology). Also the header currently does not contain any information that summarizes the list items, so even in the embedded case I think the point is moot to a certain degree...

osi_lowleveldata.proto
```diff
@@ -1,147 +0,0 @@
-syntax = "proto2";
```
pmai: Removing the LowLevelData messages might affect the ability to faithfully interpret stored data, which we guarantee for minor releases (quoting from our README):

>Backward compatibility:
>Definition: "A newer version of code can be used to read old files"
>All files that have been recorded in the past with a specific major version are FAITHFULLY valid with all combinations of
>higher minor and patch versions of the same major version.

where faithfully is defined as
>FAITHFULLY "All recorded data is correctly interpreted by the interface"

Personally I think that for the low-level-data there was not much usage until now, so breaking this promise seems valid, however given also the lane changes we should think about just releasing a 3.0.0 version? @CarloVanDriestenBMW what do you think?

CarloVanDriestenBMW:
@pmai changed the milestone to 3.0.0

pmai:
I have rebased and restructured the changes, moved the classification that was common for Lidar/Radar to the top-level and adjusted naming to match what is currently in OSI (i.e. sensor-technology specific prefixes). Will merge based on this.

Issue #80, opened 2017-07-14
Handling Junctions with OSI::GroundTruth::Lane


JonathanSchmalhoferBMW:
I currently try to find the proper representation of junctions with OSI. With the current implementation given, I propose the handling of junction as described for following example:
![junction_osi](https://user-images.githubusercontent.com/26005950/28220639-871e4986-68c0-11e7-819c-9cd5832b958f.PNG)

- each lane is defined by an array of points along its center line which is sorted in ascending order along the allowed driving direction (in this example right-hand traffic)
- ID 1 has the antecessor ID 7, no successor, the left adjacent lane is ID 2, no right adjacent lane
- ID 2 has no antecessor, the successor is ID 8, the left adjacent lane is ID 1, no right adjacent lane
- ID 3 has no antecessor, the successor is ID 7, the left adjacent lane is ID 4, no right adjacent lane
- ID 4 has the antecessor ID 8, no successor, the left adjacent lane is ID 3, no right adjacent lane
- ID 5 has the antecessor ID 9, no successor, the left adjacent lane is ID 6, no right adjacent lane
- ID 6 has no antecessor, the successor is ID 10, the left adjacent lane is ID 5, no right adjacent lane
- ID 7 has the antecessor ID 3, the successor ID 1, the left adjacent lane ID 8, no right adjacent lane
- ID 8 has the antecessor ID 2, the successor ID 4, the left adjacent lane ID 7, **no right adjacent lane**
- ID 9 has the antecessor ID 2, the successor ID 5, the left adjacent lane ID 10, **no right adjacent lane**
- ID 10 has the antecessor ID 6, the successor ID 1, the left adjacent lane ID 9, no right adjacent lane

Especially I want to propose to handle lanes on junction as following:

- ID 7 and ID 8 are **not** in any relation (antecessor, successor, right adjacent, left adjacent) to ID 9 and **vice versa**
- ID 7, ID 8, ID 3 and ID 4 are **not** in any relation (antecessor, successor, right adjacent, left adjacent) to ID 10 and **vice versa**

I would like to put this handling of junctions to a discussion. Unless other/better suggestions are being made, I propose this handling as kind of *Best Practice* until a better Representation within OSI allows a more detailled handling of relations.

CarloVanDriestenBMW:
@AndreasVIRES @MHerrmannIPG @HendrikAmelunxen could you please evaluate this "best practice" proposal?

[...]

PR #110, opened 2018-01-17
New lane model (Handling Junctions with OSI::GroundTruth::Lane - issue #80)

carsten-kuebler:
According to the OSI workshop in Ulm (November 2017), I propose this new lane model and the corresponding OSI messages. Change break downward compatibility.

### DIFF
--- a/osi_groundtruth.proto
+++ b/osi_groundtruth.proto
@@ -70,13 +70,17 @@ message GroundTruth
     //
     repeated Lane lane = 9;
 
+    // A list of lane boundaries from lanes.
+    //
+    repeated LaneBoundary lane_boundary = 10;
+
     // A list of passengers in the (ego) vehicle(s).
     //
-    repeated Occupant occupant = 10;
+    repeated Occupant occupant = 11;
 
     // Conditions of the environment.
     //
-    optional EnvironmentalConditions environmental_conditions = 11;
+    optional EnvironmentalConditions environmental_conditions = 12;
 
     // The country code in 3 digit numeric format: ISO Code 3166/1 [1,2]. 
     // e.g. Germany = 276, USA = 840
@@ -84,5 +88,5 @@ message GroundTruth
     // \par References: 
     // [1] [ISO 3166/1 Country codes] (https://www.iso.org/obp/ui/) 
     // [2] [ISO 3166/1 Wikipedia] (https://en.wikipedia.org/wiki/ISO_3166-1) 
-    optional uint32 country_code = 12;
+    optional uint32 country_code = 13;
 }
--- a/osi_landmark.proto
+++ b/osi_landmark.proto
@@ -599,7 +599,7 @@ message TrafficLight
     optional GreenArrowPresence green_arrow = 6;
 
     // The ids of the lanes that the traffic light is assigned to.
-    // Might be multiple if the traffic light is valid for multiple lanes.
+    // Might be multiple if the traffic light is valid for multiple driving lanes.
     repeated Identifier assigned_lane = 7;
 
     // Definition of traffic light types.
@@ -819,6 +819,11 @@ message RoadMarking
     // Might be multiple if the road marking goes across multiple lanes.
     repeated Identifier assigned_lane = 6;
 
+    // Intersection orientation of the road marking.
+    //
+    // \note Defined for \c Lane::type = TYPE_INTERSECTION, otherwise zero vector
+    optional Vector3d roadmarking_orientation = 7;
+
     // Definition of road marking types.
     //
     enum Type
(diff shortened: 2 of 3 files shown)


pmai:
This looks good to me at first glance, but I think it would be good to get feedback from the other workshop participants, e.g. @HendrikAmelunxen, @AndreHildebrandt, @MHerrmannIPG, @JonathanSchmalhoferBMW, or @AndreasVIRES if this matches what they expect.

osi_lane.proto
```diff
@@ -97,52 +112,15 @@ message Lane
         
         // A normal lane.
         // E.g. lanes with ids 1, 2, 3, 4 and 7 of the highway_exit image.
-        TYPE_NORMAL = 2;
```
pmai: To retain backward compatiblity TYPE_NORMAL can be changed to TYPE_DRIVING, I think, but the new TYPE_NONDRIVING and TYPE_INTERSECTION enum values should be new unused values to avoid misinterpretation, I think...

osi_lane.proto
```diff
@@ -155,6 +133,9 @@ message Lane
 //
 message LaneBoundary
 {
+    //The id of the lane boundary.
```
pmai: Again, if we want to technically remain backward compatible, the id field need to use a new tag number and the remaining fields should retain their tag numbers.

pmai:
@CarloVanDriestenBMW I think we need to decide soonish whether it makes more sense to release as 3.0.0 rather than 2.3.0 given the changes to the lane model (and some other changes)...

CarloVanDriestenBMW:
I am starting to think that the version change to 3.0.0 is reflecting the now aligned development with VDA in a better way -> and yes it may not be backward compatible

pmai:
BTW, I think this changeset is still missing the predecessor/successor pairings that @CarloVanDriestenBMW suggested to better express valid ways to cross a junction, right?

carsten-kuebler:
@pmai no, there will be no pairings for junctions. This decision was taken at OSI workshop in Ulm (2017). Valid ways depend on road markings, traffic signs linked to predecessor lanes and road markings and free lane boundaries of the junction. For this purpose, junctions have no center lines...

pmai:
Yes the Ulm consensous was to throw out the relationships for junctions. @CarloVanDriestenBMW proposed switching from separate predecessor/successor lists to a new combined pair list afterwards, to enable the logical possibilities to be still represented after the Workshop. Just wanted to ensure that it was clear whether this changeset included that or not. Thanks, will merge as soon as someone from the Environment Simulation side has reviewed.

carsten-kuebler:
highway_exit_2

we add the lane pairing to OSI::Lanes. We have to improve the documentation with this image. 

So for lane 6 the lane pairing will be (4,7) and (4,8). 
(In the previous version of the image lane 6 will be only the right adjacent lane.)

@pmai and @CarloVanDriestenBMW: PR now with lane pairing.

carsten-kuebler:
highway_exit

In this case lane 4 has two right adjacent lanes (lane 5 and 6). 6 is no successor lane of lane 4. @pmai and @CarloVanDriestenBMW do you agree?

osi_groundtruth.proto
```diff
@@ -70,19 +70,23 @@ message GroundTruth
     //
     repeated Lane lane = 9;
 
+    // A list of lane boundaries from lanes.
```
solosito: In my opinion this is redundant. Just `A list of lane boundaries.` would be enough.

osi_landmark.proto
```diff
@@ -927,14 +932,17 @@ message RoadMarking
...
         COLOR_RED = 6;
-        
+
         // Marking with green color.
+        //
         COLOR_GREEN = 7;
-        
+
         // Marking with violet color.
+        //
```
solosito: Suggestion: the same typo might be fixed in the [line 58](https://github.com/OpenSimulationInterface/open-simulation-interface/blob/0b2c5424b080778f481a7ceec72929351d2d10cc/osi_landmark.proto#L58)

osi_lane.proto
```diff
@@ -44,47 +44,64 @@ message Lane
...
     repeated Vector3d center_line = 3;
 
-    // The boundary of the lane.
+    // Definition of the intended driving direction.
     //
-    // The lane boundaries describe the lane markings or other boundaries delimiting the lane as well as free lines that
-    // might be used e.g. in construction sites. The lane markings do not have to be parallel to the center_line, but
-    // they are the basis for calculating the width of the lane at any given distance.
+    // Defined and used for driving lanes.
+    // true means driving direction is according to ascending order of center line points.
```
solosito: true means driving direction agrees with ascending storage order of the points defining the center_line.

osi_lane.proto
```diff
@@ -44,47 +44,64 @@ message Lane
...
+    // false means driving direction is according to descending order of center line points.
```
solosito: false means driving direction disagrees with descending storage order of the points defining the center_line.

solosito:
I added some comments that, in my opinion, are making it more understandable.

PR #111, opened 2018-01-17
New detected lane model (related to Handling Junctions with OSI::GroundTruth::Lane - issue #80 and pull request #110)

carsten-kuebler:
According to the OSI workshop in Ulm (November 2017), I propose this new detected lane model and the corresponding OSI messages. Change break downward compatibility.  Changes are similar to #110 (new Lane model).

 issue #80, pull request #110

### DIFF
--- a/osi_detectedlane.proto
+++ b/osi_detectedlane.proto
@@ -24,4 +24,89 @@ message DetectedLane
     // Estimated probability that this lane really exists. Range [0,1].
     //
     optional double existence_probability = 3;
+
+    // The measurement state of the traffic sign.
+    //  
+    optional MeasurementStatus measurement_state = 4;
+
+    // Definition of measurement states.
+    // 
+    enum MeasurementStatus
+    {
+        // Measurement state is unknown (must not be used in ground truth).
+        //
+        MEASUREMENT_STATE_UNKNOWN = 0;
+        
+        // Measurement state is unspecified (but known, i.e. value is not part of this enum list).
+        //
+        MEASUREMENT_STATE_OTHER = 1;
+        
+        //Lane has been measured by the sensor in the current timestep.
+        //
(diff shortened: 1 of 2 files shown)


osi_detectedlane.proto
```diff
@@ -24,4 +24,89 @@ message DetectedLane
...
+//
+message DetectedLaneBoundary
+{
+    // The id of the original laneboundary in the ground truth.
+    // In case of a ghost laneboundary (no corresponding ground truth), this field should be unset.
+    optional Identifier ground_truth_id = 1;
+
+    // The lane boundary to which this detection is associated to.
+    // 
+    optional LaneBoundary laneboundary = 2;
```
CarloVanDriestenBMW: lane_boundary ?

osi_detectedlane.proto
```diff
@@ -24,4 +25,47 @@ message DetectedLane
     // Estimated probability that this lane really exists. Range [0,1].
     //
     optional double existence_probability = 3;
+
+    // The measurement state of the detected lane.
+    //
+    optional DetectedTrafficSign.MeasurementState measurement_state = 4;
```
pmai: I think if the message is to be reused, it should be separated out from the DetectedTrafficSign message (it probably makes sense to have this centrally defined; I can provide a patch for this if agree upon). @CarloVanDriestenBMW ?

pmai: See my comments on PR #117 , I'd propose (and do this during merging) to move MeasurementState into DetectedObjectHeader, and move DetectedObjectHeader out of DetectedObject, which should handle all of this cleanly, I think. @CarloVanDriestenBMW ?

pmai:
Further refactoring is handled in PR #117, so merging for now.

Issue #97, opened 2017-11-16
Reworking the documentation


CarloVanDriestenBMW:
The doxygen documentation has been written by the ones who initially developed the OSI 2.0.0. Therefore it is crucial that an outside person reads through the definitions carefully and extends the explanaitions accordingly. I would ask @carsten-kuebler to do this in the beginning of January after the updates end of December. The version 2.3.0 will be released with the updated documentation at the 26th of January.

PR #112, opened 2018-01-17
Standardised Message Naming (related to issue #97)

carsten-kuebler:
SensorSpecific messages had a different naming convention. Renamed in a similar way as other messages.
This changeset is related to issue #97 (Reworking documentation).

### DIFF
--- a/osi_detectedobject.proto
+++ b/osi_detectedobject.proto
@@ -54,15 +54,19 @@ message DetectedObject
 
     // Additional data that is specific to radar sensors.
     // Field need not be set if simulated sensor is not a radar sensor.
-    optional RadarSpecificObjectData radar_specifics = 10;
+    optional SpecificRadarData radar_specifics = 10;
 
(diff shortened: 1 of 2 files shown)


carsten-kuebler:
@pmai specific sensor data messages for radar, lidar etc. uses a message naming <sensor-name><message-type>. Other messages have a naming <message-type><sensor-name><...>. If you search for similar messages it is easier to use one naming convention. 
But yes it is not neccessary. Would you prefer SpecificSensorDataRadar or should we rename DetectedTrafficSign. I find it easier to use a naming convention. This change set makes only sense with the new FeatureData and the new DetectionXXX messages.

pmai:
I've split out the non-naming related changes to PR #123, leaving this PR/branch (or a new one) as a place to do a global naming cleanup and merge that at the end.

carsten-kuebler:
@CarloVanDriestenBMW I checked the actual master and I found no need to change a message name. So this PR is obsolete.

PR #115, opened 2018-01-17
Fix version number field defaults (fixes #92)

### LINKED ISSUE #92: OSI-Version mismatch possible because of being optional
The Version is set as default value for the message **InterfaceVersion** within the **setup.py** script. All fields in that message are set to **optional**. Unfortunately, when creating a message for publishing and not explicitly setting the version values, we noticed, that the default values are not being used. So if Sender and Receiver have different OSI-Versions, by using default values, a mismatch will not be noticed. Our suggestion: The Sender should be **required** to fill the OSI-Version with his default values explicitly.

pmai:
This change allows the version number field in ground truth to work as expected, i.e. a version mismatch between sender and receiver can now be found. This change requires the sender to actively set the version number, for which the sender can use the current_interface_version file option of the osi_version.proto file to always use the version the sender was compiled against.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,7 @@ endforeach()
 configure_file(osi_version.proto.in ${CMAKE_CURRENT_SOURCE_DIR}/osi_version.proto)
 
 find_package(Protobuf 2.6.1 REQUIRED)
+set(PROTOBUF_IMPORT_DIRS ${PROTOBUF_INCLUDE_DIRS})
 
 set(OSI_PROTO_FILES
     osi_version.proto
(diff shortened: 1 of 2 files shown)


pmai:
I think you mean the VERSION file? Nothing changes there, it is still used to provide the current version info as the file option, and both crake build and python build work as is, the version info just ends up in the file option and is still used for e.g. documentation and packaging.

CarloVanDriestenBMW:
Ok, I get it now

pmai:
@CarloVanDriestenBMW can you give this a review so that I can merge?

Issue #99, opened 2017-11-20
Additional parameters for sensor models


hschoen:
OSI should completely transport all information that sensor models may require. From the radar point of view I suggest to add the following parameters:

![radar_gen](https://user-images.githubusercontent.com/25342898/33018204-b77b6af6-cdf5-11e7-915f-5e5fe2424f5c.png)
_Additional parameters for a phenomenological sensor model_

![radar_phys](https://user-images.githubusercontent.com/25342898/33018513-fee0c034-cdf6-11e7-8ad0-c7e463ee363d.png)
_Additional parameters for a physical sensor model_

Some of these parameters are already part of OSI but given for completeness.

For camera and LiDAR (and other sensors such as ultra sonic) parameters have to be completed as well.

martin-fzd:
For LiDAR we need a kind of **beam matrix**, describing in which direction are rays (i.e. laser beams) to be shot. Here we define the total number of elevation layer as well as the spacing in azimuth (angular resolution). Maybe doable with a external sensor config which instantiates specific sensors, see #98 ? From here we could also read the **Beam divergence** and the **wavelength** of the LiDAR.

We also need an interface allowing for incorporating **intrinsic calibration** (important for intensity calculations!) as well as a paramters for **ray steeing**, i.e. rotation rate, immitating steeing of laser beams (e,g, from left to right).

jondo2010:
Possibly related question about the interface intentions: How should it be that a model implementing a LiDaR sensor is able to generate it's low-level point cloud with only `GroundTruth` input? Doesn't it need access to some sort of geometry to perform intersection/ray-cast queries upon?

martin-fzd:
I agree. Material assignments and geometric properties must be considered for simulation model validation. Not sure whether this will be part of OSI...?
It will be even more challenging if don't have a “newly delivered, just washed” approximation for appearance of objects, which also affects camera models. Should we think about a degree of dirtyness, similar as we have found it for rain?

Here is my proposal for now:
As a large car model cataloge is available in the tools and the "GroundTruth" Objects could be made parameterizable: Basic materials could be assigned to parts of the objects (let's start with glass, metal, and plastic for simplicity) which allows for including specific material reflection properties for different sensors.

CarloVanDriestenBMW:
@martin-fzd maybe you can get in touch because I think you are working on such a lidar module? There will be a big low-level data update with the next release btw

PMeyerDSPACE:
[...]

PR #116, opened 2018-01-17
Add undercariage clearance to vehicles

pmai:
This optional attribute allows sensor models that want to approximate
the influence of see-under without needing access to the whole vehicle
geometry.  Addresses the part of issue #99 that can be addressed
directly in ground truth data.


pmai:
I think that in the case of attached trailers we have a second vehicle, with its own underbody plane?

PR #117, opened 2018-01-19
Detected Objects adapted to real sensor structures.

carsten-kuebler:
DetectedObject is adapted to standardised real sensor output structures. Other detected objects (like traffic sign/light etc.) are prepared.

DetectedObject list is adapted to the standardisation activities for real sensors. Other detected objects are draft.

### DIFF
--- a/osi_detectedobject.proto
+++ b/osi_detectedobject.proto
@@ -14,59 +14,96 @@ package osi;
 //
 message DetectedObject
 {
+    // Header attributes of detected object.
+    //
+    optional DetectedObjectHeader header = 1;
+
     // Specific id of the object as assigned by the sensor internally. Need not match with ground_truth_id.
     //
-    optional Identifier tracking_id = 1;
+    optional Identifier tracking_id = 2;
 
     // The id of the original object in the ground truth list of vehicles / objects / ..
     // Multiple entries if detected object is a merge of multiple ground truth objects.
-    repeated Identifier ground_truth_id = 2;
+    repeated Identifier ground_truth_id = 3;
+
+    // The amount of time that this object has been currently observed/tracked.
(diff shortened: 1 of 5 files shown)


osi_detectedobject.proto
```diff
@@ -161,4 +223,99 @@ message DetectedObject
         //
         optional double prob_animal = 9;
     }
+
+    // Definition of measurement states.
+    //
+    enum MeasurementState
```
pmai: Would it make sense to move this enum also to the DetectedObjectHeader, which would make reuse simpler (see my comments in PR #111 )?

osi_detectedobject.proto
```diff
@@ -161,4 +223,99 @@ message DetectedObject
...
+        
+        // Traffic sign has not been measured by the sensor in the current timestep. Values provided by tracking only.
+        //
+        MEASUREMENT_STATE_PREDICTED = 3;
+    }
+
+    //
+    // \brief The header attributes of each detected object.
+    //
+    message DetectedObjectHeader
```
pmai: Maybe also move the DetectedObjectHeader definition out of DetectedObject, to make reuse easier?

osi_detectedobject.proto
```diff
@@ -161,4 +223,99 @@ message DetectedObject
...
+        // 
+        optional InterfaceVersion version = 1;
+
+        // Time stamp at which the measurement was taken (not the time at which it was processed or at which it is transmitted) in the global synchronized time. 
+        //
+        optional Timestamp timestamp = 2;
+
+        // Continuous up counter to identify the cycle.
+        //
+        optional uint32 cylcle_counter = 3;
```
jondo2010: `cycle_counter`

PR #120, opened 2018-01-24
Add initial draft of sensor input configuration message (fix #98)

### LINKED ISSUE #98: Concept for config data
# New concept for config data for sensor models and ray-tracer/renderers
Currently the parameters of the sensor models are hard coded (e.g. field of view of the sensor). However, it might be interesting for the user to vary these parameters.

now
_Situation now_

I suggest to define a new group of OSI-parameters called "configuration parameters". These parameters are then set either in the scenario or in a config file (or a dedicated GUI window) of the environment simulation.

future
_Suggested concept for config data_

The config data should also contain configuration parameters for physical simulation. For LiDAR and radar this would be the number of rays etc. for ray-tracing and for camera the picture resolution, dynamic range etc. for rendering.
The config data can get large in size if it contains data like antenna diagrams etc. Hence config data is transmitted only ONCE (to the ray-tracer/renderer/sensor model) at simulation start (or at low frequency) to avoid unnecessary high data load at every simulation step. 

radarconfig
_Example for config data for radar sensor model_

pmai:
This is a first draft of the message used to configure the sensor input
between the model and the simulation environment. The mechanics of the
message interchange are determined by the model packaging.

It should be noted that the sensor input configuration mechanism is
separate from any mechanisms that are used to configure the sensor
model itself, which currently are packaging dependent.

Note that this pull request is to aid reviewing and further development,
more work to flesh out this approach if adopted is still needed. This is
just the result of discussions with @raueATconti based on the input by
@hschoen and others.

### DIFF
--- a/osi_sensorinputconfiguration.proto
+++ b/osi_sensorinputconfiguration.proto
@@ -20,6 +20,16 @@ package osi;
 // configuration allowed for multiple alternatives, in which case the set
 // configuration will only contain the alternative chosen.
 //
+// It should be noted that this message is not intended to provide for
+// parametrization of a generic sensor model, but rather for the automatic
+// configuration of an environment simulation in order to supply the 
+// necessary input to it, depending on its actual configuration.
+// Mechanisms to parametrize sensor models are currently packaging-specific,
+// i.e. they depend on the packaging mechanism chosen:  For FMU-packaging
+// the parametrization can be implemented using normal FMU parameters,
+// and the requested sensor input configuration can depend on those
+// parameter values by being defined as a calculatedParameter.
+//
 message SensorInputConfiguration
 {
     // The mounting position of the sensor (origin and orientation of the sensor coordinate system)
@@ -117,21 +127,23 @@ message RadarSensorInputConfiguration
     // This represents the combined TX/RX antenna diagram
     repeated AntennaDiagramEntry combined_antenna_diagram = 1;
 
-    // Number of rays to cast across horizontal field of view
+    // Number of rays to cast across horizontal field of view (azimuth)
     optional uint32 number_of_rays_horizontal = 2;
 
-    // Number of rays to cast across vertical field of view
+    // Number of rays to cast across vertical field of view (elevation)
     optional uint32 number_of_rays_vertical = 3;
 
     // Maximum number of interactions to take into account
     optional uint32 max_number_of_interactions = 4;
 
     message AntennaDiagramEntry {
-        // Horizontal deflection of entry in sensor/antenna coordinates
+        // Horizontal deflection (azimuth) of entry in sensor/antenna
+        // coordinates
         // Unit: [rad]
         optional double horizontal_angle = 1;
 
-        // Vertical deflection of entry in sensor/antenna coordinates
+        // Vertical deflection (elevation) of entry in sensor/antenna
+        // coordinates
         // Unit: [rad]
         optional double vertical_angle = 2;
 
--- a/osi_sensorinputconfiguration.proto
+++ b/osi_sensorinputconfiguration.proto
@@ -1,0 +1,216 @@
+syntax = "proto2";
+
+option optimize_for = SPEED;
+
+import "osi_common.proto";
+
+package osi;
+
+//
+// \brief The configuration settings for the Sensor Input to be provided
+// by the environment simulation.
+//
+// This message can be provided by the sensor model to the environment
+// simulation, in which case it describes the input configuration that
+// is desired by the sensor model.  In response the environment simulation
+// will configure the input and provide a new message of this type, which
+// describes the actual configuration that it is going to employ.  The two
+// can and will differ, when either the environment simulation does not
+// support a given requested configuration, and/or when the requested
+// configuration allowed for multiple alternatives, in which case the set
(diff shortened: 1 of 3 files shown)


osi_sensorinputconfiguration.proto
```diff
@@ -0,0 +1,216 @@
...
+    // This specifies the rate at which the sensor model is provided with
+    // new input data.
+    //
+    // Unit: [s]
+    // \note In the case of FMU packaging this will correspond to the
+    // communication step size.
+    optional double update_cycle_time = 5;
+
+    // Radar-specific Sensor Input Configuration
+    optional RadarSensorInputConfiguration radar_sensor_input_configuration = 1000;
```
CarloVanDriestenBMW: It is good to start at 1000 :)

CarloVanDriestenBMW:
@haoyuanying do you have an opinion about further configurations?

pmai:
I have added the timing related fields as discussed on Wednesday to the configuration messages.

jondo2010:
Interesting first steps. What about supporting non-passive sensors? I.e., most cameras have gain/gamma control that can be simulated with HDR shaders, and some recent LiDaR modules can be beam-steered. This would require additional (optional) control messages from the sensor models back to the simulator.

pmai:
Theoretically extending this to non-passive sensor/simulator interaction would be fairly simple (e.g. adding sensor-controlled gain, and then exchanging configuration messages also during runtime), it would mostly need the ability to signal a changed configuration request to the simulator. Since the actual messaging involved is not part of OSI, but layered standards, like e.g. OSMP for FMU packaging, this would have to be addressed there (e.g. for FMUs the request is already a calculated parameter, the set configuration would then be a tunable parameter, and one would need a flag output to signal the change request).

In practice most sensor models currently will request input that comprises all possible input (i.e. full dynamic range on camera data, full envelope on lidar), and then do the active behaviour inside the sensor model...

But for a next release I think moving more of the work into the simulation environment can make sense from a convenience and performance PoV (by being able to leverage work already being done in the simulation environment). The input config messages will have to be extended to cover more needed functionality in any case (they are currently just bare minimums to demonstrate the approach).

pmai:
Hmm, but all of the SensorInputConfiguration fields are in fact used to configure the environment simulation, this is not specific to the Radar stuff (e.g. the FoV for all sensors and Channel-Config for camerasensors) . The reason we did not name them EnvironmentSimulationConfiguration is mostly to make clear that we are specifying requirements for the sensor input, and it is up to the rest of the system to determine which parts of the simulation system this affects (and how), since this is to a degree implementation-dependent... Also there will be multiple configurations for multiple sensors, thus the name EnvironmentSimulationConfiguration could be construed as misleading.

I agree that the comments should mention azimuth and Elevation, will add that...

pmai:
I've tried to clarify a few things in the documentation.

CarloVanDriestenBMW:
A feedback from the team:

> The sensor and company specific config data must be defined e.g. by overwriting the initial (standartized) config message

pmai:
@CarloVanDriestenBMW, @hschoen this needs a review to be merged...

osi_sensorinputconfiguration.proto
haoyuanying: @CarloVanDriestenBMW +1 :)

haoyuanying:
@pmai @CarloVanDriestenBMW @hschoen I have a feeling that the newly added sensorinputconfiguration message is going beyond the simulation level that the current OSI stands... Since I do not see that the current OSI GT supports raytracing based sensor model (e.g. the triangle based detailized bounding box description, even in 3D!), is it necessary to have the sensor configuration like the number of rays here? I would propose that let OSI stand at phenomenological level, and replace the physical model parameters (e.g. number of rays) to abstraction models (e.g. probability models).

pmai:
The GT-based detailed information for raytracing based models is supposed to be added in a seperate PR, i.e. the idea really is to have OSI also support this use case.

hschoen:
I agree that the ray tracing use case is of paramount importance from the sensor model point of view.

haoyuanying:
@hschoen I do think that raytracing is very important in sensor models PoV as well. The question I have is ATM the sensor model description in OSI is more advanced than the GT description (if the raytracer needed scene is completely relied on the GT description, and no further off-side fine-grained scene reconstruction), and I am not sure how necessary / applicable to use raytracing model to simulate a sensor based on the current GT description (or release 3.0)...

hschoen:
@haoyuanying : I agree, the transmission of the ray tracer scene via OSI is not solved yet in OSI release 3.0. In the meantime this will be done with a different (3D-compatible) interface. Still, ray tracer configuration via OSI should be possible also because OSI might support 3D scene description in a future release.

pmai:
@CarloVanDriestenBMW, @hschoen, @haoyuanying I've added nominal Emitter frequency to Radar/Lidar, due to ist influence on material/geometry interaction and on doppler shift calculation (the latter can actually be recalculated after the fact, but for the former this is e.g. needed to select proper RCS maps, etc.).

I'll show usage for this in a seperate pull request, that adds a sensor view which can contain the results of a simple ray tracing component in the environment simulation, for further processing in the sensor model proper.

PR #121, opened 2018-01-24
Add is_stationary convenience flag to Vehicle (#89)

pmai:
Convenience flag, needs discussion about whether we need something like this elsewhere.

### DIFF
--- a/osi_object.proto
+++ b/osi_object.proto
@@ -9,14 +9,14 @@ package osi;
 //
 // \brief A vehicle (ego or other).
 //
-// This is basically a BaseObject with additional attributes, such as type and lights. 
-// The origin of the rear (front) axis coordinate system in world coordinates is calculated as: 
-// base.position + R * bbcenter_to_rear (front) for the ego vehicle (R rotates from vehicle to world frame, i.e. inverse orientation of base.orientation). 
-// 
-// For all vehicles, including ego vehicles, the position given 
+// This is basically a BaseObject with additional attributes, such as type and lights.
+// The origin of the rear (front) axis coordinate system in world coordinates is calculated as:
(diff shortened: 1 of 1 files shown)


CarloVanDriestenBMW:
default should be moving imo

carsten-kuebler:
From standardisation (real sensors) activities point of view there is an enum DetectedObject::MovementState (see #117).

Simplified petri net for moving objects:

O --> stationary --> moving <--> stopped

pmai:
So should we support the more expressive state (distinguishing between unknown, stationary and stopped vs. moving)? Given that this is GroundTruth, I'm not even sure we need the flag at all, but what is the opinion of the requesters (see issue #89), i.e. @HannesSchneiderAVL, @haoyuanying ?

carsten-kuebler:
@pmai what is the benefit, if we specify in GT that a vehicle is temporary stationary? I see no benefit. Sensor(Output)Data can benefit from this information, because it's a local, temporary valid information.

pmai:
If noone (@HannesSchneiderAVL, @haoyuanying ) of the original requesters comments on this, I'm going to drop the PR for now...

haoyuanying:
@pmai @carsten-kuebler Would it be beneficial in terms of enhancing the performance to have the flag as a redundent index to check if the vehicle is temperorilly stationary if comparing to check velocities in X, Y, Z?

pmai:
I don't think there will be any significant performance savings (i.e. transmitting another field in protobuf is likely to be much more exprensive than any simple x, y, z comparision (i.e. at most 3 substractions, 3 abs, 2 additions and 1 compare). And utility libraries can internally add this optimization on top for clients...

HannesSchneiderAVL:
sorry for the late answer. We discussed this internally and came also to the conclusion that this could have performance issues, so we also see no need for this flag any more.

pmai:
Closing PR for now due to lessened Need.

Issue #119, opened 2018-01-22
Traffic Signs: End of minumum speed missing


ReinhardBiegelIntech:
The ```enum Type``` in ```TrafficSign``` is missing an entry for "end of minimum speed". German StVO references this sign as [no. 279](https://www.dvr.de/publikationen/downloads/verkehrszeichen.html).

If there are no objections, I'd create a PR for adding this type.

CarloVanDriestenBMW:
@carsten-kuebler if you have a pullrequst on that already @ReinhardBiegelIntech could than review and alter it?

carsten-kuebler:
@CarloVanDriestenBMW no, there is no open pullrequest. I would propose to extend pullrequest #114 (Extended Documentation) and add the sign 279.

By the way, there are other signs which are missing (no_parking begin/end 283.1-3 286.1-3 --> additional arrows on signs). 
Or the one-way road sign is missing...

carsten-kuebler:
@CarloVanDriestenBMW and @ReinhardBiegelIntech please comment/review pullrequest #122 .

pmai:
Closed with PR merge.

PR #122, opened 2018-01-24
Additional signs and road markings (issue #119)

carsten-kuebler:
Complete some signs and add new signs and road markings.

Add possibility to define addition arrows on traffic signs (eg. on no parking signs). Introduce view_normal of displayed traffic sign symbol.


ReinhardBiegelIntech:
Looks good to me. Thank you.

PR #125, opened 2018-01-29
OSI Bibtex Key added and renaming of "motor bike" to "motorbike"

martin-fzd:
OSI Bibtex Key for standardized citation.
Renaming of "motor bike" to "motorbike" following recommondation in various dictionaries (e.g. [https://dictionary.cambridge.org/dictionary/english-german/motorbike](cambridge) )

PR #127, opened 2018-01-30
BibTeX and motorbike cleanups

pmai:
Contains the BibTeX and motorbike naming fixes from PR #125 in cleaned up form.


martin-fzd:
Works for me. Please also see #125

CarloVanDriestenBMW:
@pmai my review comment was lost:

> Use the following citation for referencing the OSI interface in your scientific work:

Issue #60, opened 2017-03-31
Add "road condition"


hschoen:
Did I miss it or don't we have a parameter for the road condition?

- wet
- icy
- covered with snow
- oil
- sand
- leaves
- etc.

Also the layer thickness would be relevant (especially water layer for radar).
Coverage (in percent) of road?

TimoHankeBMW:
No such parameter exists yet as there is no exactly specified and agreed-upon list of road conditions.
Especially the 'exactly specified' part makes this quite hard, e.g. 'wet' is a very broad und unspecific term, of whichi every usage instance will likely have a different understanding.
As discussions for specifications would likely fill many days' worth of workshops, I'd suggest that the best way is to wait for a specification from the scenario definition and use that. If these don't match, we would need to adapt anyway.

PhilippRoethigBMW:
@hschoen and @TimoHankeBMW what do you think about implementing the condition of the road in a separate message? This message could define specific and measurable properties of the road rather than just determine a general category for it. Of course we would need to decide which properties would make sense.

```cmd
//
// \brief The condition of the roads surface.
//
message RoadCondition
{
    // The temperature of the roads surface in degree Celsius.
    //
    optional double surface_temperature = 1;

    // The height of the water film on top of the surface in mm.
    //
    optional double surface_water_film = 2;

    // The temperature where the water on top of the surface would start to freeze or dew in degree Celsius.
    //
    optional double surface_freezing_point = 3;

    // The percentage of ice covering the road.
    //
    optional double surface_ice = 4;

    // The coefficient presenting the friction and the grip of the road.
    //
    optional double surface_friction = 5;
}
```

hschoen:
@PhilippRoethigBMW, I think this is a good start! We should however merge this solution with the specification from the scenario definition as proposed by @TimoHankeBMW.
From a radar point of view I would like to add roughness of the road surface (clutter). Here we should use the definition used in environment simulation. @AndreasVIRES and @MHerrmannIPG , do you have a definition for roughness of the road surface?

MHerrmannIPG:
@hschoen, no, there's currently no parameter for the roughness of the road surface. It's no problem to add a value to the test specification globally for the complete road network which then is available to any sensors. Making it variable along the track/in various areas of the network would be a bigger task.

AndreasVIRES:
**roughness**
[...]

PR #128, opened 2018-01-31
RoadConditions added (Issue #60)

carsten-kuebler:
Add road conditions for every lane. Use International Roughness Index for roughness.

This pull request address issue #60.

### DIFF
--- a/osi_lane.proto
+++ b/osi_lane.proto
@@ -83,6 +83,9 @@ message Lane
     //
     optional LaneFraming lane_framing = 9;
 
+    // The condition of the lane, e.g. influenced by weather. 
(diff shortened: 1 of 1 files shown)


CarloVanDriestenBMW:
@hschoen This is the already merged proposal - is everything you need included?

hschoen:
From a radar point of view we must be able to estimate reflection and scattering from the road surface.  For this we will need information about roughness, wetness and the (relative, complex) permittivity of the road. The first two items are included in the proposal (surface_roughness and surface_water_film), which appreciate very much! The **permittivity** is not included in the proposal which is fine if it is given elsewhere - the permittivity is not a road condition (dynamic) but a road (material) property (static).

PR #130, opened 2018-02-07
Clarify and refactor timestamp field definitions

pmai:
This adds clarifying documentation to the timing-related fields, renames
the timestamp fields in DetectedObjectHeader and DetectionHeader to
indicate their measurment time related nature, adds a field named
last_measurement_time to the SensorData message to provide an upper
bound on embedded measurement times and removes legacy/ill-defined
fields.

### DIFF
--- a/osi_detectedobject.proto
+++ b/osi_detectedobject.proto
@@ -265,9 +265,13 @@ message DetectedObjectHeader
     //
     optional InterfaceVersion version = 1;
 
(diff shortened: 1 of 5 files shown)


CarloVanDriestenBMW:
@pmai aren't we missing the "simulation_start_time" ?

pmai:
@CarloVanDriestenBMW the simulation start time is already contained in the sensor input configuration message (it is constant during the simulation run, thus only needs to be transmitted once; we could probably move more config data to that message in the future).

PR #132, opened 2018-02-07
Refactor SensorDataGroundTruth, add SensorView

pmai:
This cleans up SensorData, removes ModelInternal, SensorDataGroundTruth,
and introduces the new SensorView message which contains sensor-relative
and sensor-specific views on the GroundTruth. Note that this is a first
draft for discussion.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,8 +61,9 @@ set(OSI_PROTO_FILES
     osi_object.proto
     osi_occupant.proto
     osi_sensordata.proto
-    osi_sensorspecific.proto
     osi_sensorinputconfiguration.proto
+    osi_sensorspecific.proto
+    osi_sensorview.proto
 )
 
 protobuf_generate_cpp(PROTO_SRCS PROTO_HEADERS ${OSI_PROTO_FILES})
--- a/osi_modelinternal.proto
+++ b/osi_modelinternal.proto
@@ -8,24 +8,6 @@ import "osi_object.proto";
 package osi;
 
 //
-// \brief Internal data and state flags for the sensor model, set and used during modification of the interface data.
-//
-// Content of all ModelInternal* messages may be cleared at the end of the sensor model processing chain and not sent
-// with the final output of the sensor model. Receiving components must not expect values in ModelInternal* messages to
-// be set. In particular, the message structures defined in this file must not be used by receiving components outside
-// of the sensor model to read required input data.
-//
-message ModelInternal
-{
-    // Low level data interface, for use by models performing statistical generation of such "low level data" before
-    // object hypothesis and tracking from object ground truth.
-    //
-    // Low Level data should be transferred to receiving components by means of a separate data stream. Therefore they
-    // are not part the top level SensorData message, as they are used only internally by some models.
-    optional FeatureData feature_data = 1;
-}
-
-//
 // \brief Additional internal data and state flags for SensorDataObjects.
 //
 message ModelInternalObject
--- a/osi_sensordata.proto
+++ b/osi_sensordata.proto
@@ -8,7 +8,8 @@ import "osi_detectedlane.proto";
 import "osi_detectedobject.proto";
 import "osi_detectedoccupant.proto";
 import "osi_groundtruth.proto";
-import "osi_modelinternal.proto";
+import "osi_sensorview.proto";
+import "osi_featuredata.proto";
 
 package osi;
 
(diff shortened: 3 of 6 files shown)


hschoen:
@pmai : Thank you for the proposal! From radar point of view I miss the following in osi_sensorview.proto :

1. Direction of RX-ray for post-processing of phase difference (for azimuth angle).
2. (Optional) data for validation of sensor signal processing (interaction type, -position, TX direction, etc.)

pmai:
@haoyuanying That is precisely the idea, i.e. the the sensor model requests the detailed SensorView through its SensorInputConfiguration message, and the simulation environment provides the data through an internal Ray Tracing engine; if the simulation environment does not support this, it will notify the sensor model through its changed SensorInputConfiguration message (the response). As an alternative, one can imagine a seperate pre-filter FMU, which takes a pure SensorData message with only the GT filled out by the simulation environment, and based on that recreates a scene graph and does internal Ray Tracing to generate the additional SensorView data (this might require some trickery to identify the GT objects with more detailed geometry data and texture data that is provided in the ray tracing configuration).

In the future, one could imagine ray tracing processing fragments (e.g. hit-detection codelets, etc.) to be passed into the simulation environment through the SensorInputConfiguration, to be JIT-compiled and used in the ray tracing process...

pmai:
@hschoen doesn't the RX-ray direction presuppose that the ray tracing engine is doing bidirectional ray tracing? If it only does "normal" ray tracing it would not have that information since there is no RX ray (only a TX/RX ray which has the predetermined direction based on the raster)?

Could you make a proposal for the additional data you would want to see?

hschoen:
For the calculation of the phase difference between several RX antennas, angles between the ray received by RX antenna and (normal vector to the) RX plane are required. Proposal:

// Azimuth angle between RX plane and received ray
optional RX_ray_horizontal_angle = 1;
// Elevation between RX plane and received ray
optional RX_ray_vertical_angle = 2;

The corresponding TX ray angles are only needed for validation purposes. In case of backscatter RX angles = TX angles. For multipath rays this is not true.

pmai:
@hschoen the horizontal and vertical angle of the received ray are implicit due to its position in the repeated reflection field: It contains rays in left-to-right, top-to-bottom order for all cast rays (i.e. dividing the FoV angle by the number of rays cast gives you the delta angle that you add to the beginning FoV to get the angle). The data currently does not contain the source of the ray, i.e. there is no relation to the TX side...

I don't think the current proposal really supports multiple RX or TX antennas directly, i.e. much more would be needed to support that I think (i.e. we do not have antenna locations and orientations seperately from the sensor position)...

So I think this proposal can definitely be extended, but I think more is needed if stuff like phased arrays or multi-antenna setups are to be covered.

pmai:
Which reminds me of one issue: The combined antenna diagram needs to be seperated out, i.e. seperate RX and TX diagramms, since the angle of the RX ray and the corresponding TX ray(s) do not need to be identical if we want to support multipath.

hschoen:
@pmai : **Multipath** must be supported - it is the main reason for ray tracing in a physical model. Direct backscatter without multipath can also be modelled in a phenomenological model with the knowledge of 3D geometry from GT.
**Multiple RX and TX antennas** do not necessarily need to be supported by OSI (and ray tracer), because this can be taken into account by post processing in the sensor model. If the distance between antennas and the incidence angles are known the phase difference can be calculated.
If **horizontal and vertical angles** of the rays (for _both_ RX and TX) are implicit in the repeated ray field I agree that no further parameter for the angles is required. Is it really true that the rays are related to the receiver (RX)? As the cast rays originate from the transmitter (TX) I would have expected that the angles of the scan field are related to TX and hence we have to add RX angles (which are not identical for multi-path).
I would appreciate if **RX and TX antenna diagrams are separated**. However, if RX and TX antenna characteristics are identical (which is usually the case), using a combined antenna diagram would still be possible by multiplying the value at TX angle by the value at RX angle for each ray.

pmai:
@hschoen non-multipath raytracing is also useful I'd say, since this exactly eliminates the need to transfer knowledge of the 3d geometry and textures to the sensor model, i.e. the sampling of the geometry happens in the simulation environment. That said I'm all for multipath to also be taken into account.

The current definition only contains the angles for the RX side, as one would expect for simple "normal one-way" ray tracing: The rays start at the RX antenna (the "eye") and go backwards through multiple interactions/hits until they hit the TX antenna (the light source), which then means that we have a signal. Going from TX to RX would be "backward" ray tracing in my opinion (some switch the terms around), which is also viable, but different, and some schemes combine both directions.

I think we should split the RX and TX antenna diagrams, because otherwise I don't think this will work in general for multi-path. I'll also add the TX angles to support multi-cast.

PR #133, opened 2018-02-07
Add road surface texture (or fine roughness)

pmai:
This addresses issue #60 further, by including the surface texture or
fine roughness in addition to the surface_roughness field which is
indicating unevenness.

### DIFF
--- a/osi_lane.proto
+++ b/osi_lane.proto
@@ -179,8 +179,10 @@ message Lane
         //
         optional double surface_friction = 5;
 
-        // The coefficient presenting the roughness of the road. 
-        // International Roughness Index (IRI) [1] values range from 0 = smooth ground (equivalent to driving on a plate of glass) up to > 20 mm/m (a very rough road). 
+        // The coefficient presenting the roughness or unevenness of the road.
+        // International Roughness Index (IRI) [1] values range from 0 = smooth
+        // ground (equivalent to driving on a plate of glass) up to > 20 mm/m
+        // (a very rough road). 
         //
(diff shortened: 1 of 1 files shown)


CarloVanDriestenBMW:
@raueATconti can you comment on camera models here?

hschoen:
Fine roughness and unevenness as I proposed are discussed in the following dissertation. However, the definition of unevenness there slightly deviates from IRI, but I suppose they are compatible.
**[Robert Schneider, „Modellierung der Wellenausbreitung für ein bildgebendes Kfz-
Radar“, Dissertation, Universität Karlsruhe, Mai 1998.](https://publikationen.bibliothek.kit.edu/24298)**
This quote may be used in the source code comment for fine roughness. For unevenness I agree that IRI is rather the standard.

martin-fzd:
What is the intended use of road surface? Do you think of a (piecewise) definition of the road ahead?
 I'm asking as I wonder how to handle transitions between different road conditions efficiently...

pmai:
@martin-fzd I think for now (i.e. for phenomenological models) that's likely the way to go, i.e. if you need changing road conditions (water, different texture, ...) you need to model that as a new lane. Having multiple conditions for the same lane, with switch-over points seems fairly complex, for little gain, since we need actual models using e.g. road textures and ray tracing or similar to handle road conditions more realistically.

pmai:
I'll merge this change for now, with the added quote, since the overall discussion of how to handle changing road conditions goes beyond the current PR.

PR #144, opened 2018-02-26
Add host vehicle data as sensor input data

pmai:
This changeset introduces HostVehicleData which comprises all vehicle
internal data that is available to the sensors (or in the future other
parts of OSI) as additional input. This includes e.g. GPS-sensor derived
data as well as e.g. engine ECU data, etc.  For now this only includes
GPS-sensor location data with error estimation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,6 +54,7 @@ set(OSI_PROTO_FILES
     osi_detectedoccupant.proto
     osi_environment.proto
     osi_groundtruth.proto
+    osi_hostvehicledata.proto
     osi_landmark.proto
     osi_lane.proto
     osi_featuredata.proto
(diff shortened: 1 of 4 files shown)


martin-fzd:
Rather than GPS Data, I would advise to read Steering Wheel Angle, Wheel Speeds and Yaw Rather from the host vehicle, as these are more valueable for e.g. tracking algorithms.

CarloVanDriestenBMW:
@martin-fzd I think the goal is to extend the message over time. With the current implementation it is more or less the output of a componente that estimates the host vehicles location. We can make that subject to the next pegasus meeting?

martin-fzd:
Yes, please.

CarloVanDriestenBMW:
@raueATconti can you add this to the agenda?

Issue #51, opened 2017-02-23
Compare: OSI and DIN70000


CarloVanDriestenBMW:
As mentioned in #42 there may be discrepancies between these two sources. We should talk about this in the Workshop.

CarloVanDriestenBMW:
Question: Is the bounding box dynamic?

carsten-kuebler:
@CarloVanDriestenBMW The ISO8855 and DIN 70.000 define the same vehicle coord. system and to "our disappointment" it is yaw->roll->pitch. (See Tabelle 1 Seite 16 in ISO8855). Should we change the definition in OSI? It will affect all orientations, spherical coords (issue #148) and maybe the polygons (issue #154 define polygon clockwise instead of counter-clockwise).
OSI uses the american definition of vehicle coord. systems (See https://en.wikipedia.org/wiki/Axes_conventions#/media/File:RPY_angles_of_cars.png).
OSI uses few more angles in messages, which could be defined similar (OSI::Reflections in OSI::SensorView...

martin-fzd:
I don't see any resons **not** to stick to the refered ISO / DIN norms in an interface standardization initative like this. Any other opinions?

CarloVanDriestenBMW:
@carsten-kuebler I would like the complete OSI to be checked for definitions which are not compatible with ISO8855 and all of them reworked. Please check the complete document if possible. I guess it has only 45 pages so it is ok :)

carsten-kuebler:
@CarloVanDriestenBMW There is no need to change our lane or road definitions?!

CarloVanDriestenBMW:
@carsten-kuebler Your proposal is good! Maybe we consider #14 ? I already asked @pmai for an opinion there.

We will check the lane definitions - could you talk with T. Schaller?

TimFrickeBMW:
For the record: there is no "american way of rotation". The Wikipedia article (!) cited does indeed state "roll-pitch-yaw", but **as a name, not as a definition.** What may be confusing is that extrinsic rotations around x-y-z (roll-pitch-yaw) result in the same attitude as intrinsic rotations around z-y'-x'' (yaw-pitch-roll). From my experience, the common definition around the world is the Tait-Bryan convention of intrinsic rotations z-y’-x″, whis is often incorrectly referred to as "Euler angles". Note that in the Wikipedia article cited above there is even a link to the "Tait-Bryan angles".

carsten-kuebler:
[...]

PR #158, opened 2018-03-08
Use vehicle axis system definition of DIN ISO 8855

carsten-kuebler:
This PR addresses issue #51 

Change OSI transformation between axis systems to ISO definition.
World frame -> vehicle frame -> sensor frame ...

Orientation is defined (world -> vehicle) rotate yaw -> rotate pitch -> rotate roll.

Note this is the international standard and differs from national american standard.
(OSI V2.xxx used national american standard)

**Additional Bugfixes:**
remove road marking orientation

remove camera sensor specific values. Redundant!

### DIFF
--- a/osi_common.proto
+++ b/osi_common.proto
@@ -97,18 +97,24 @@ message Dimension3d
 // 
 // The preferred angular range is (-pi, pi]. The coordinate system is defined as right-handed. 
 //  
-// The rotations are to be performed \b roll \b first, \b pitch \b second and \b yaw \b third to follow the definition according to [1]. 
+// The rotations are to be performed \b yaw \b first, \b pitch \b second and \b roll \b third to follow the definition according to [1]. 
 // 
 // Rotations are defined in the reference coordinate frame around z-axis (=yaw), y-axis (=pitch) and x-axis (=roll), 
 // not in the body frame of the object. 
-// A positive angle corresponds to a counter-clockwise rotation around the respective axis.
+// A positive angle corresponds to a clockwise rotation around the respective axis.
 //
 // Roll/Pitch are 0 if the objects xy-plane is parallel to its parent's xy-plane.
 // Yaw is 0 if the object's local x-axis is parallel to its parent's x-axis.
 //
+// Rotation_yaw_pitch_roll = Rotation_roll*Rotation_pitch*Rotation_yaw
+//
+// vector_global_coord_system := Inverse_Rotation_yaw_pitch_roll(orientation)*(vector_local_coord_system) + local_origin.position
+
+//
+// \note This definition changed in OSI version 3.0.0. Previous OSI versions (V2.xx) had an other definition.
+//
 // \par References: 
-// [1] LAVALLE, Steven M. Planning algorithms. Cambridge university press, 2006. Chapter 3.2.3.
-// See http://planning.cs.uiuc.edu/node102.html for online version of this chapter.
+// [1] DIN ISO 8855:2013-11
 //
 message Orientation3d
 {
@@ -155,7 +161,7 @@ message MountingPosition
     optional Vector3d position = 1;
 
     // Orientation offset relative to the specified reference coordinate system.
-    //
+    // Origin_sensor := Rotation_yaw_pitch_roll(MountingPosition.orientation)*(Origin_reference_coordinate_system - MountingPosition.position)
     optional Orientation3d orientation = 2;
 }
 
@@ -167,7 +173,10 @@ message MountingPosition
 // Units are [m] for radial distance and [rad] for azimuth and elevation angles. 
 // If azimuth and elevation are zero, the referenced point lies directly ahead
 // in the central viewing direction of the sensor, and the respective vector points directly in the latter.
+// A positive angle corresponds to a clockwise rotation around the respective axis, i.e. 
+// z-axis (azimuth = yaw), y-axis (elevation = pitch).
 //
+// vector_cartesian := Rotation(elevation)*Rotation(azimuth)*Unit_vector_x*distance
 message Spherical3d
 {
     // The radial distance.
@@ -188,6 +197,9 @@ message Spherical3d
 // 
 // This includes the StationaryObject, TrafficSign, TrafficLight, RoadMarking messages.
 //
+// All coordinates and orientations are relative to the global ground truth
+// coordinate system.
+//
 message BaseStationary
 {
     // The 3D dimensions of the stationary object (bounding box), e.g. a landmark:
@@ -199,7 +211,9 @@ message BaseStationary
     optional Vector3d position = 2;
 
     // The relative orientation of the stationary object w.r.t. its parent frame.
-    // There may be some constraints how to use orientation w.r.t. to some stationary object's or entity's definition.
+    //
+    // Origin_base_stationary_entity := Rotation_yaw_pitch_roll(BaseStationary.orientation)*(Origin_parent_coordinate_system - BaseStationary.position)
+    // \note There may be some constraints how to align the orientation w.r.t. to some stationary object's or entity's definition.
     optional Orientation3d orientation = 3;
     
     // Usage as ground truth:
@@ -225,6 +239,9 @@ message BaseStationary
 // 
 // This includes the MovingObject messages.
 //
+// All coordinates and orientations are relative to the global ground truth
+// coordinate system.
+//
 message BaseMoving
 {
     // The 3D dimension of the moving object (its bounding box).
@@ -237,18 +254,27 @@ message BaseMoving
 
     // The relative orientation of the moving object w.r.t. its parent frame.
     //
+    // Origin_base_moving_entity := Rotation_yaw_pitch_roll(BaseMoving.orientation)*(Origin_parent_coordinate_system - BaseMoving.position)
+    // \note There may be some constraints how to align the orientation w.r.t. to some stationary object's or entity's definition.
     optional Orientation3d orientation = 3;
 
     // The relative velocity of the moving object w.r.t. its parent frame and parent velocity.
     // The velocity becomes global/absolute if the parent frame does not move.
+    //
+    // BaseMoving.position(t) := BaseMoving.position(t-dt)+velocity*dt 
     optional Vector3d velocity = 4;
 
     // The relative acceleration of the moving object w.r.t. its parent frame and parent acceleration.
     // The acceleration becomes global/absolute if the parent frame is not accelerating.
+    //
+    // BaseMoving.position(t) := BaseMoving.position(t-dt)+velocity*dt+acceleration/2*dt^2
+    // BaseMoving.velocity(t) := BaseMoving.velocity(t-dt)+acceleration*dt 
     optional Vector3d acceleration = 5;
 
-    // The relative orientation rate of the moving object w.r.t. its parent frame and parent orientation rate in the center point of the bounding box.
-    // The orientation rate becomes global/absolute if the parent frame is not rotating.
+    // The relative orientation rate of the moving object w.r.t. its parent frame and parent orientation rate in the center point of the bounding box (origin of the bounding box frame).
+    //
+    // Rotation_yaw_pitch_roll(BaseMoving.orientation(t)) := Rotation_yaw_pitch_roll(BaseMoving.orientation_rate*dt)*Rotation_yaw_pitch_roll(BaseMoving.orientation(t-dt))
+    // \note BaseMoving.orientation(t) is \b not equal BaseMoving.orientation(t-dt)+BaseMoving.orientation_rate*dt
     optional Orientation3d orientation_rate = 6;
     
     // Usage as ground truth:
--- a/osi_common.proto
+++ b/osi_common.proto
@@ -87,21 +87,22 @@ message Dimension3d
 }
 
 //
-// \brief A 3D orientation, orientation rate or orientation acceleration (i.e. derivatives)
-// or its uncertainties denoted in euler angles.
+// \brief A 3D orientation, orientation rate or orientation acceleration (i.e. 
+// derivatives) or its uncertainties denoted in euler angles.
 // 
 // Units are 
 // \arg [rad] for orientation
 // \arg [rad/s] for rates
 // \arg [rad/s^2] for accelerations
 // 
-// The preferred angular range is (-pi, pi]. The coordinate system is defined as right-handed. 
+// The preferred angular range is (-pi, pi]. The coordinate system is defined as 
+// right-handed. 
+// For the sense of each rotation, the right-hand rule applies.
 //  
-// The rotations are to be performed \b yaw \b first, \b pitch \b second and \b roll \b third to follow the definition according to [1]. 
-// 
-// Rotations are defined in the reference coordinate frame around z-axis (=yaw), y-axis (=pitch) and x-axis (=roll), 
-// not in the body frame of the object. 
-// A positive angle corresponds to a clockwise rotation around the respective axis.
+// The rotations are to be performed \b yaw \b first (around the z-axis), 
+// \b pitch \b second (around the new y-axis) and \b roll \b third (around the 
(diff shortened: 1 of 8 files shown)


osi_common.proto
```diff
@@ -97,18 +97,24 @@ message Dimension3d
 // 
 // The preferred angular range is (-pi, pi]. The coordinate system is defined as right-handed. 
 //  
-// The rotations are to be performed \b roll \b first, \b pitch \b second and \b yaw \b third to follow the definition according to [1]. 
+// The rotations are to be performed \b yaw \b first, \b pitch \b second and \b roll \b third to follow the definition according to [1]. 
 // 
 // Rotations are defined in the reference coordinate frame around z-axis (=yaw), y-axis (=pitch) and x-axis (=roll), 
 // not in the body frame of the object. 
-// A positive angle corresponds to a counter-clockwise rotation around the respective axis.
+// A positive angle corresponds to a clockwise rotation around the respective axis.
```
TimFrickeBMW: The terms "clockwise" or "counter-clockwise" are highly ambiguous and should thus be avoided! I propose the following unambiguous scientific definition, for which a lot of illustrative examples pop up when you make a Google search:
```For the sense of each rotation, the right-hand rule applies.```

File: osi_common.proto [hunk 2]
```diff
@@ -97,18 +97,24 @@ message Dimension3d
 // 
 // The preferred angular range is (-pi, pi]. The coordinate system is defined as right-handed. 
 //  
-// The rotations are to be performed \b roll \b first, \b pitch \b second and \b yaw \b third to follow the definition according to [1]. 
+// The rotations are to be performed \b yaw \b first, \b pitch \b second and \b roll \b third to follow the definition according to [1].
```
TimFrickeBMW: Changing the order of rotations to yaw-pitch-roll makes this definition conform with various norms (DIN 8855 / DIN 70000 / DIN 9300). **HOWEVER**, this order of rotations applies to **INTRINSIC ROTATIONS**, i.e., every rotation is performed along the body axes. This means, that the statement in lines 102 and 103 **has to be removed!** All in all, I propose the following definition:

``The rotations are to be performed \b yaw \b first (around the z-axis), \b pitch \b second (around the new y-axis) and \b roll \b third (around the new x-axis) to follow the definition according to [1] (Tait-Bryan / Euler convention z-y'-x'').``

Explicitly stating the Tait-Bryan / Euler convention z-y'-x'' further reduces ambiguity and helps those who need to google this topic.
Note that the original definition of extrinsic rotations and the roll-pitch-yaw sequence **implicitly** produces the same result as the new definition of intrinsic rotations and the yaw-pitch-roll sequence. The major flaw of the old definition - apart from being non-standard - is, that the explanation of the extrinsic rotations is ambiguous: if the origin of the "reference coordinate frame" is chosen different from the body frame's origin, rotations would also result in translations!

osi_common.proto
```diff
@@ -167,7 +173,10 @@ message MountingPosition
 // Units are [m] for radial distance and [rad] for azimuth and elevation angles. 
 // If azimuth and elevation are zero, the referenced point lies directly ahead
 // in the central viewing direction of the sensor, and the respective vector points directly in the latter.
+// A positive angle corresponds to a clockwise rotation around the respective axis, i.e. 
+// z-axis (azimuth = yaw), y-axis (elevation = pitch).
```
TimFrickeBMW: This definition is ambiguous ("clockwise") if not even plain wrong ("y-axis" - I think the rotated y'-axis was intended). I propose the following definition:

``With the x-axis of the sensor frame pointing in the central viewing direction of the sensor and the senor frame being right-handed, azimuth and elevation are defined as the rotations that would have to be applied to the sensor frame to make its x-axis point towards the referenced point or to align it with the referenced vector. The rotations are to be performed \b azimuth \b first (around the z-axis) and \b elevation \b second (around the new y-axis) to follow the definition of Orientation3D. For the sense of each rotation, the right-hand rule applies.``

Moreover, I would recommend stating the definition first and the example (lines 174 and 175) second.

osi_object.proto
```diff
@@ -15,10 +15,10 @@ package osi;
...
 // orientation of base.orientation).
 // 
-// For all vehicles, including host vehicles, the position given  in
+// For all vehicles, including host vehicles, the position given in
 // base.position points to the center of the vehicle's bounding box.
 //
-// The object coordinates are defined as x-axis is the direction from rear to
+// The vehicle object coordinates are defined as x-axis is the direction from rear to
 // front of the vehicle, y-axis corresponds to rear axle and z-axis points to
 // vehicle ceiling.
```
TimFrickeBMW: The direction of the y-axis is not defined here! Please add a note specifying that the coordinate system is right-handed.

osi_detectedobject.proto
```diff
@@ -93,16 +93,20 @@ message DetectedObject
     // The estimated probabilities for the object to belong to a specific class.
     //
     optional ClassProbability class_probability = 17;
-
-    // The estimated pedestrian head pose.
+    
+    // Pedestrian head pose for behaviour prediction. Describes the head
+    // orientation w.r.t. the host vehicle orientation.
+    // x-axis of head_pose is pedestrian's straight ahead viewing direction.
```
TimFrickeBMW: Proposed extended definition for reduced ambiguity:
``The x-axis of the right-handed head frame is pointing along the pedestrian's straight ahead viewing direction and the z-axis is pointing upwards.``

osi_detectedobject.proto
```diff
@@ -93,16 +93,20 @@ message DetectedObject
...
     //
-    // \note Pedestrian specific value.
-    optional Orientation3d pedestrian_head_pose = 18;
+    // View_normal_base_coord_system = Inverse_Rotation(head_pose)*Unit_vector_x 
+    optional Orientation3d head_pose = 18;
 
-    // The estimated pedestrian upper body pose.
+    // Pedestrian upper body pose for behaviour prediction. Describes the
+    // upper body orientation w.r.t. the host vehicle orientation.
+    // x-axis of upper_body_pose is pedestrian's upper body intended moving direction.
```
TimFrickeBMW: Proposed extended definition for reduced ambiguity:
``The x-axis of the right-handed upper body frame is pointing along the pedestrian's upper body intended moving direction and the z-axis is pointing upwards.``

osi_sensorinputconfiguration.proto
```diff
@@ -64,23 +64,31 @@ package osi;
...
     // when setting the input configuration.
+    //
     optional MountingPosition mounting_position = 1;
 
     // Field of View in horizontal orientation of the sensor
     //
     // This determines the limit of the cone of interest of ground truth
     // that the simulation environment has to provide.
+    // Viewing range: [-field_of_view_horizontal/2, field_of_view_horizontal]
+    // in sensor coordinate system.
```
TimFrickeBMW: It seems that the range should be defined as:
``[-field_of_view_horizontal/2, field_of_view_horizontal/2]``
Morover, consider adding the following note to avoid confusion as to what horizontal means:
``Note: 'horizontal' does not necessarily mean 'perpendicular to the local gravitational vector', but more generally refers to the left/right viewing range of the sensor.``

osi_sensorinputconfiguration.proto
```diff
@@ -89,6 +97,8 @@ message SensorInputConfiguration
     //
     // This determines the limit of the cone of interest of ground truth
     // that the simulation environment has to provide.
+    // Viewing range: [-field_of_view_vertical/2, field_of_view_vertical]
+    // in sensor coordinate system.
```
TimFrickeBMW: It seems that the range should be defined as:
``[-field_of_view_vertical/2, field_of_view_vertical/2]``
Moreover, I would concretize the definition as follows:
``Viewing range: [-field_of_view_vertical/2, field_of_view_vertical/2]
elevation in the sensor frame at zero azimuth as defined in Spherical3d.``

osi_featuredata.proto
```diff
@@ -54,17 +54,18 @@ message DetectionHeader
...
     // The origin represents the current mounting pose to the best knowledge of
-    // the sensor. It includes the estimated 6D pose estimation given by  the
+    // the sensor. It includes the estimated 6D pose estimation given by the
     // calibration. The uncertainty of this estimation is given with the
     // corresponding 6D root mean squared error. The estimation of the current
     // origin does not include effects due to short-time dynamics, such as pitch
     // angles from braking, but includes long-time calibration values, such as
     // pitch angles from luggage in the trunk.
+    // 
     optional MountingPosition mounting_position = 4;
```
TimFrickeBMW: Even after multiple reads I do not understand the definition in lines 47 through 67.

It seems that the "mounting position" not only covers the position, but also the orientation. Consider renaming... "MountingPosition6DoF" for instance... Ist that what the term "pose" is supposed to mean here? The combination of position and orientation?

Furthermore, the definition of the reference coordinate system (the host vehicle coordinate system) is ambiguous. It could be defined as follows:
``The host vehicle coordinate system is right-handed and body-fixed. Its xz-plane is the vehicle's plane of symmetry. In static equilibrium on horizontal ground, its y-axis is collinear with the rear axle, pointing left, and the z-axis is pointing vertically upwards.``
With this definition, "mounting position" (i.e., position and orientation) can be said to be defined "with respect to the host vehicle cooridnate system" without any ambiguity.

osi_detectedobject.proto
carsten-kuebler: Is definition z-axis upwards neccessary? Or is it an unintended constraint!

carsten-kuebler: // The x-axis of the right-handed head frame is pointing along the 
    // pedestrian's straight ahead viewing direction and the z-axis is pointing 
    // upwards (cranial direction [1] i.e. to pedestrian's skull cap).
    //
    // \par References: 
    // [1] https://en.wikipedia.org/wiki/Anatomical_terms_of_location

osi_detectedobject.proto
carsten-kuebler: // The x-axis of the right-handed upper body frame is pointing along the 
    // pedestrian's upper body ventral direction [2] (i.e. usually pedestrian's 
    // intended moving direction) and the z-axis is pointing upwards (to 
    // pedestrian's head).
    //
    // \par References: 
    // [2] https://en.wikipedia.org/wiki/Anatomical_terms_of_location

osi_featuredata.proto
carsten-kuebler: I made a reference to OSI::Vehicle. So we need not to redefine the host vehicle coordinate system. 

    // Mounting position of the sensor (origin and orientation of the cartesian
    // sensor coordinate system); given relative from the middle of the rear
    // axle of the host vehicle coordinate system (see: \c OSI::Vehicle vehicle
    // reference point) to the sensor detection coordinate system's origin. The
    // sensor detection coordinate system is a spherical polar coordinate system
    // (same origin and orientation as cartesian sensor coordinate system).
    //
    // The sensor detection coordinate system's x-axis must be the angle
    // bisector of the sensor's horizontal and vertical field of view in a
    // right-handed coordinate system. The xy-plane represents the horizontal
    // field of view and the xz-plane the vertical field of view.
    //
    // The origin represents the current mounting pose to the best knowledge of
    // the sensor. The estimation of the 6D pose given by the calibration. The
    // uncertainty of this estimation is given with the corresponding 6D root
    // mean squared error. The estimation of the current origin does not include
    // effects due to short-time dynamics, such as pitch angles from braking,
    // but includes long-time calibration values, such as pitch angles from
    // luggage in the trunk.

carsten-kuebler:
@TimFrickeBMW I included your comments. I replied to some comments.

osi_detectedobject.proto
TimFrickeBMW: I think the minimum information required about the head attitude is yaw (pitch can probably be neglected and roll is irrelevant in my opinion). Now, since we define attitude by intrinsic rotations, the direction of the z-axis is important! Whether it points upwards or downwards basically determines wheter a positive yawing motion corresponds to a movement of the head towards the left or towards the right.
Good idea to mention the cranial direction!

osi_detectedobject.proto
TimFrickeBMW: Since we define attitude by intrinsic rotations, the direction of the z-axis is important! Whether it points upwards or downwards basically determines wheter a positive yawing motion corresponds to a movement of the body towards the left or towards the right.
Good idea to mention the ventral direction!

osi_featuredata.proto
```diff
@@ -44,35 +49,35 @@ message DetectionHeader
...
+    // sensor detection coordinate system is a spherical polar coordinate system
+    // (same origin and orientation as cartesian sensor coordinate system).
     //
-    // The x-axis must be the angle bisector of the horizontal field of view in
-    // a right-handed coordinate system. The xy-plane represents the
-    // horizontal field of view and the xz-plane the vertical field of view.
+    // The sensor detection coordinate system's x-axis must be the angle
+    // bisector of the sensor's horizontal and vertical field of view in a
+    // right-handed coordinate system. The xy-plane represents the horizontal
+    // field of view and the xz-plane the vertical field of view.
```
TimFrickeBMW: Do we have to define a cartesian AND a spherical sensor frame? (I would stick to the word "frame" as in OSI:Spherical3d and not use "coordinate system" here. Would be nice to have a consistent wording throughout OSI.) Furthermore, I would recommend clarifying the meaning of "horitzontal" and "vertical" as misunderstandings might arise otherwise. I propose the following formulation:


``Mounting position of the sensor (origin and orientation of the sensor frame). Both origin and orientation are given in and with respect to the host vehicle coordinate system (see: \c OSI::Vehicle vehicle reference point).``

``The sensor frame's x-axis is pointing in the central viewing direction of the sensor. It is the angle bisector of the sensor's horizontal and vertical field of view. The terms horizontal and vertical must be understood as names for the two principal axes of the sensor's field of view, which do not have to be horizontal or vertical in the strict sense of being parallel or perpendicular to the local gravitational vector. The horizontal field of view defines the sensor frame's xy-plane and the vertical field of view defines the xz-plane. The sensor frame is right-handed and the z-axis is pointing in an upward direction.``

osi_common.proto
```diff
@@ -164,10 +175,21 @@ message MountingPosition
 //
 // Used e.g., for low level representations of radar detections.
 // 
-// Units are [m] for radial distance and [rad] for azimuth and elevation angles. 
+// Units are [m] for radial distance and [rad] for azimuth and elevation angles.
+// With the x-axis of the sensor frame pointing in the central viewing direction
```
TimFrickeBMW: We may want to rephrase this definition to be consistent with the defintion of the sensor frame in OSI:DetectionHeader. I propose a reference to avoid redundancies:
``Azimuth and elevation are defined as the rotations that would have to be applied to the sensor frame (definition in OSI:DetectionHeader) to make its x-axis point towards the referenced point or to align it with the referenced vector. The rotations are to be performed \b azimuth \b first (around the z-axis) and \b elevation \b second (around the new y-axis) to follow the definition of OSI:Orientation3D. For the sense of each rotation, the right-hand rule applies.``
I am not sure, however, whether OSI:DetectionHeader is the optimal location for the original definition.

osi_featuredata.proto
carsten-kuebler: The sensor frame is cartesian. But detections may be defined with spherical coordinates. To distinguish this possible change I would propose to use "sensor frame" (cartesian) and "sensor detection frame" (spherical, cartesian, ...).

Horizontal and vertical is defined relative to the sensor (detection) frame's orientation.

osi_common.proto
carsten-kuebler: Normally detections uses spherical coordinates but it should be a common definition. I adapted your definition.

carsten-kuebler:
@TimFrickeBMW I add your comments to the documentation...

CarloVanDriestenBMW:
WOW! That is by far the most impressive review/PR until now :+1: 

THX @carsten-kuebler and @TimFrickeBMW 

Nice work!

PR #151, opened 2018-03-01
Detected... message references sensors (issue #147)

carsten-kuebler:
This PR adress issue #147 

Add a list of sensor ids to all Detected... messages. These sensors confirm this Detected... 'object'. It is a redundant info.

Issue #147, opened 2018-02-26
Fusion State


CarloVanDriestenBMW:
The OSI:SensorData needs a field which keeps track how many sensors have seen an object after the fusion. Example: The camera does a fusion of objects from the radar. The output of the camera would have the information for every detectedObject which sensor has seen it.
Something like a `repeated identifier detectedBySensorIDs = 1000;`

PR #161, opened 2018-03-16
New proposal for SensorID List

carsten-kuebler:
New proposal for PR #151

### DIFF
--- a/osi_detectedlandmark.proto
+++ b/osi_detectedlandmark.proto
@@ -15,11 +15,13 @@ message DetectedTrafficSign
 {
     // Specific ID of the traffic sign as assigned by the sensor internally.
     // Need not match with ground_truth_id.
+    //
     optional Identifier tracking_id = 2;
 
     // The ID of the original traffic sign in the ground truth.
     // In case of a ghost sign (no corresponding ground truth), this field
     // should be unset.
+    //
     optional Identifier ground_truth_id = 3;
 
     // A list of candidates for this traffic sign as estimated by the sensor.
@@ -27,7 +29,8 @@ message DetectedTrafficSign
     repeated CandidateSign candidate_sign = 4;
 
     // A list of candidates for (a) possible supplementary sign(s) as estimated
-    //     by the sensor.
+    // by the sensor.
+    //
     repeated CandidateSupplementarySign candidate_supplementary_sign = 5;
 
     // The detected geometry of the traffic sign.
@@ -39,6 +42,7 @@ message DetectedTrafficSign
     //
     // \note Use as confidence measure where a low value means less confidence
     // and a high value indicates strong confidence.
+    //
     optional double existence_probability = 7;
 
     // The measurement state.
@@ -47,6 +51,7 @@ message DetectedTrafficSign
     
     // The root mean squared error of the base parameters of the detected
     // traffic sign.
+    //
     optional BaseStationary rmse = 9;
 
     // Links to the corresponding lanes.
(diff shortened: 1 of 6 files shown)


carsten-kuebler:
@CarloVanDriestenBMW It was faster to add a new PR instead of merging the master in the old PR. I hope it is ok....
I changed the field name for sensor_id in SensorData.

carsten-kuebler:
This PR addresses issue #147.

CarloVanDriestenBMW:
I guess we have to define how the version of the ISO standard and the version of OSI is combined. E.g. an entry in the readme.md in order to clearly define to which ISO this specific OSI version relates to?

carsten-kuebler:
I agree, but we can't refer to this standard (there is no number). The standard will define the communications and if we add semantic compatibility (as a restriction to minor version changes) I think we are prepared.

carsten-kuebler:
@CarloVanDriestenBMW: last changes have no renumbering of fields. We should do this at last....

pmai:
@carsten-kuebler, @CarloVanDriestenBMW I've rebased onto master and squashed the fixup commit, so please git checkout Detected--message-references-sensors-V2 && git fetch && git reset --hard origin/Detected--message-references-sensors-V2 prior to further work on this branch.

Generally-speaking, is it wise to seperate the Header from the Objects like this? This requires a more complex traversal for everyone, AFAICS, with limited benefit? I.e. having a generic Header message and including that in all DetectedXXX seems like a more direct approach, or am I missing something?

carsten-kuebler:
@pmai The header is not for one detected entity. It is for all objects of one detected entity . Maybe it is easier to understand, if we use plurals for field names.

pmai:
@carsten-kuebler I've re-reset this branch, since a merge commit which included all the old commits reappeared: Please be sure that after a branch reset you do not do a git pull on the branch, but rather a git fetch and git reset --hard origin/branchname, to take up the rebased stuff.

The reason I do these rebases is to cleanup history prior to merging into mainline (i.e. avoid fixup commits, multiple merges between branches, conflicts, etc.), to keep development history usable going forward.

pmai:
@carsten-kuebler On the object headers: yes, the naming and old usage fooled me; if the data is really only per entity kind, we should somehow make this a bit clearer in the names and/or documentation, I think... Otherwise looks good to me...

pmai:
@carsten-kuebler, @CarloVanDriestenBMW  I have re-rebased to solve the conflicts with newly merged master as best I could; please re-check if this still matches what is wanted. I'd suggest to merge these content PRs prior to doing major changes of documentation and/or field renumberings, since this will create lots of spurious conflicts...

pmai:
@CarloVanDriestenBMW is conflict free and should be mergeable now; approve if this works for you.

osi_detectedlandmark.proto
```diff
@@ -13,53 +13,62 @@ package osi;
...
     optional Identifier ground_truth_id = 3;
 
     // A list of candidates for this traffic sign as estimated by the sensor.
     //    
     repeated CandidateSign candidate_sign = 4;
 
-    // A list of candidates for (a) possible supplementary sign(s) as estimated by the sensor.
-    //    
+    // A list of candidates for (a) possible supplementary sign(s) as estimated
+    //     by the sensor.
```
CarloVanDriestenBMW: too many spaces?

osi_sensordata.proto
```diff
@@ -26,9 +26,11 @@ package osi;
 //
 message SensorData
 {  
-    // The ID of the sensor.
+    // The ID of the sensor at host vehicle's mounting_position.
     //
-    optional Identifier id = 1;
+    // This ID can equal \c DetectionHeader::sensor_id, if SensorData hold only
```
CarloVanDriestenBMW: holds?

pmai:
@CarloVanDriestenBMW I've fixed and regularized the Doxygen comments.

PR #167, opened 2018-03-22
Review documentation and add BaseStationary to supplementary signs

carsten-kuebler:
This PR addresses issue #97.

Regularize doxygen comments.
Add links to OSI message fields and enums.
Add "plural" form to repeated field names (add _list/multiple_, add _s etc.).
Remove typos and replace special chars.

**Additional:**

Split CanditateSupplementarySign in two messages (every supplementary sign has its own list of candidates and a base_rmse).
Add BaseStationary to SupplementarySign and remove its position_supplementary field.


CarloVanDriestenBMW:
plural of abbreviations...hmmm true

I am in favor for: point -> points

It should be as simple as possible and also noted in the wiki how the rule is for OSI (guideline)

carsten-kuebler:
point -> points. 
So you would change 
base_polyline -> base_polyline_points (or outline_points) and
center_line -> center_line_points

CarloVanDriestenBMW:
I actually forgot about the polyline... 
I think there are two variants:
1) point -> points (they have no pattern or connection)
2) point -> polyline (the points together form a line, like a lane marking or a lane boundary?)

btw: sensor_id_list -> sensor_id_lists would be consistent, or?

osi_detectedlandmark.proto
```diff
@@ -159,24 +165,48 @@ message CandidateSign
 }
 
 //
-// \brief A candidate for (a) detected supplementary sign(s) as estimated by the
+// \brief Candidates for all detected supplementary signs of one traffic sign
+// as estimated by the sensor.
+//
+message CandidateSupplementarySigns
```
CarloVanDriestenBMW: @carsten-kuebler I think I do not understand this. Why is the base_rmse not in the "CandidateSupplementarySign". The name "candidate_multiple_supplementary_signs " seems not intuitive as well as I would not create two messages which sound so similar like "CandidateSupplementarySign" and "CandidateSupplementarySigns".
Should we have a call for that?

carsten-kuebler:
"btw: sensor_id_list -> sensor_id_lists would be consistent, or?"

I would say no. 
It is a list of sensor IDs --> sensor_IDs (Capital letters should not be used)
Or a sensor ID list --> sensor_id_list
But not a list of list --> sensor_id_lists

osi_detectedlandmark.proto
carsten-kuebler: I agree, how would you call a list of supplementary signs -> sign forest or sign jungle. Both sounds colloquially.

We can have a call from 11:30. Is it ok?

CarloVanDriestenBMW:
Identifier sensor_id
repeated Identifier sensor_ids

// If new type necessary
message IdentifierList
(
repeated Identifier sensor_ids
...
)

repeated IdentifierList sensor_id_lists


I think we can make it the same way with everything else eg. "detection", "detections" ....

carsten-kuebler:
> Identifier sensor_id --> repeated Identifier sensor_ids

 OK!

> // If new type necessary -> this would btw be in osi_common.proto
> message IdentifierList
> (
> repeated Identifier sensor_ids
> ...
> )

// If new type necessary -> this would btw be in osi_common.proto
message IdentifierList
(
repeated Identifier sensor_ids
)

No additional fields! If there are more fields (e.g. a header), then the message name should be different to "List"

I checked all _id_list. Everything would be ok, but what's about sensor_data_list --> there is no datas?!

CarloVanDriestenBMW:
I see it in recordings!!! OK:
SensorData is actually singular and plural at once!

"SensorData(Time)Series" contains "repeated SensorData sensor_data_series"

"SensorDataList" seems redundant to me?

"SensorDataSeriesList" contains "repeated SensorData(Time)Series sensor_data_series_list"

right?

carsten-kuebler:
@CarloVanDriestenBMW I made the changes (also removed the plural s field names). All messages in osi_datarecording.proto are necessary.)

osi_datarecording.proto
```diff
@@ -27,16 +29,20 @@ message SensorDataList
 {
     // List of sensor data for multiple sensors at a specific timestep.
     //
-    repeated SensorData sensors = 1;
+    // \note OSI uses singular instead of plural for repeated field names.
```
CarloVanDriestenBMW: This is meant for the word "data" but not for everything else I guess?

osi_datarecording.proto
```diff
@@ -27,16 +29,20 @@ message SensorDataList
...
+    //
+    repeated SensorData sensor = 1;
```
CarloVanDriestenBMW: this should be sensor_data

osi_datarecording.proto
```diff
@@ -27,16 +29,20 @@ message SensorDataList
...
+// \c SensorData messages.
 //
 message SensorDataSeriesList
 {
     // List of sensor data for multiple sensors at subsequent timesteps.
     //
-    repeated SensorDataSeries sensors = 1;
+    // \note OSI uses singular instead of plural for repeated field names.
+    //
+    repeated SensorDataSeries sensor = 1;
```
CarloVanDriestenBMW: sensor_data_series

The word series is both singular and plural. Same case as "data"

osi_detectedlandmark.proto
```diff
@@ -24,164 +24,219 @@ message DetectedTrafficSign
...
+    optional EstimatedMainSign main_sign = 4;
 
-    // The detected geometry of the traffic sign.
-    //    
-    optional Geometry geometry = 6;
+    // A list of additional supplementary sign(s) as estimated by the sensor.
+    //
+    // \note OSI uses singular instead of plural for repeated field names.
+    //
+    repeated EstimatedSupplementarySign supplementary_sign = 5;
```
CarloVanDriestenBMW: We have never used the word "estimated" before. It should be "detected"

osi_detectedlandmark.proto
```diff
@@ -24,164 +24,219 @@ message DetectedTrafficSign
...
     optional double existence_probability = 7;
 
-    // The measurement state.
+    // Links to the corresponding lanes.
     //
-    optional MeasurementState measurement_state = 8;
-
-    // The root mean squared error of the base parameters of the detected
-    // traffic sign.
+    // \note OSI uses singular instead of plural for repeated field names.
```
CarloVanDriestenBMW: When did we say that???

osi_detectedlandmark.proto
```diff
@@ -24,164 +24,219 @@ message DetectedTrafficSign
...
-        // Traffic sign that has an oversize rectangle geometry.
-        // E.g. direction plates on highway or city signs.
-        //
-        GEOMETRY_PLATE = 12;
+        // A list of candidates for this traffic sign as estimated by the
+        // sensor.
+        //
+        // \note OSI uses singular instead of plural for repeated field names.
+        //
+        repeated CandidateSign candidate = 1;
```
CarloVanDriestenBMW: candidate_signs,

I have no idea why we lost the complete naming convention now...

osi_detectedlandmark.proto
```diff
@@ -24,164 +24,219 @@ message DetectedTrafficSign
...
     //
-    optional double candidate_probability = 2;
+    message EstimatedSupplementarySign
+    {
+        // The definition of one of more supplementary signs that together
+        // define this candidate.
+        //
+        // \note OSI uses singular instead of plural for repeated field names.
+        //
+        repeated CandidateSupplementarySign candidate = 1;
```
CarloVanDriestenBMW: candidate_supplementary_signs

osi_detectedlandmark.proto
```diff
@@ -24,164 +24,219 @@ message DetectedTrafficSign
...
+        //
+        // \brief A candidate for a detected supplementary sign as estimated by
+        // the sensor.
+        //
+        message CandidateSupplementarySign
+        {
+            // The definition of one of more supplementary signs that together
+            // define this candidate.
+            //
+            optional TrafficSign.SupplementarySign sign = 1;
```
CarloVanDriestenBMW: supplementary_sign

osi_detectedlandmark.proto
```diff
@@ -225,17 +278,68 @@ message DetectedTrafficLight
...
+    // sensor.
+    //
+    message EstimatedTrafficLight
+    {
+        // A list of candidates for this traffic light as estimated by the
+        // sensor.
+        //
+        // \note OSI uses singular instead of plural for repeated field names.
+        //
+        repeated CandidateTrafficLight candidate = 1;
```
CarloVanDriestenBMW: We cannot do:
optional TrafficLight traffic_light = 1; 
and at the same time:
repeated CandidateTrafficLight candidate = 1;

The naming convention must always follow the same pattern!
repeated CandidateTrafficLight candidate_traffic_lights = 1;

osi_datarecording.proto
carsten-kuebler: Maybe I misunderstood the discussion with Pierre?!
So we use definitive PLURAL? --> No problem! All field names (which have to change are marked with the same note).

osi_datarecording.proto
carsten-kuebler: OK.

osi_datarecording.proto
carsten-kuebler: Please note that you would adress
sensor_data_series (SensorDataSeriesList) --> sensor_data (SensorDataSeries) -> ... 

often you find in OSI like:
sensor(s) (you get a message SensorDataSeries) -> data (you get a message SensorData) -> ...

osi_detectedlandmark.proto
carsten-kuebler: We get a problem, if we use the message structure for the detected messages. 

Would you prefer DetectedSupplementarySignData?

osi_detectedlandmark.proto
carsten-kuebler: see first comment

osi_detectedlandmark.proto
carsten-kuebler: Do you really think so. See some lines below. If we change consequent, then you must rename geometry --> traffic_sign_geometry...

osi_detectedlandmark.proto
carsten-kuebler: see previous comment.

osi_detectedlandmark.proto
carsten-kuebler: One option for a pattern is, to use the message name as field name. (drawback: if you use a message twice in a message the field names would be the same. There are other drawbacks..).

Second option: you choose a field name which is clearly in the message context and you would not repeat infromation several times. 

In OSI we have a mix of both...

CarloVanDriestenBMW:
@carsten-kuebler ok, let's resolve the conflict and merge. We will do this with every PR and afterwards we go over everything again for naming and index numbers

carsten-kuebler:
Ok. Should we synchronize our activities by phone?

CarloVanDriestenBMW:
right_lane_boundary_id

> The "right adjacent lane boundaries" may only be shared with the "left adjacent lane boundaries" of the "right adjacent lane"

CarloVanDriestenBMW:
@carsten-kuebler please check again. I think I solved it...

carsten-kuebler:
I reviewed the merge.

carsten-kuebler:
Removed tabs.
Merge @CarloVanDriestenBMW update

carsten-kuebler:
@CarloVanDriestenBMW PR is ok

PR #168, opened 2018-03-22
Remove RelevantLane and RelevantDirection in Detected...

carsten-kuebler:
Remove RelevantLane and RelevantDirection.
Relevant lanes and directions are defined in Candidate... assigned_lanes and traffic light's icons.


CarloVanDriestenBMW:
Question:
I think how I understand these fields now, they represent the "best guess" of the sensor on what it thinks is the assoziated lane and not the lane it is actually assigned to!

carsten-kuebler:
Yes. Would you leave the associated_lane_id(s) empty? I would use associated_lane_id(s) as a best guess of the sensor...

CarloVanDriestenBMW:
@carsten-kuebler where is the "associated_lane_id" ? So:

The sensor has a "detectedLane" and a "DetectedTrafficSign" and within the sign we have an associated_lane_id ?

carsten-kuebler:
Yes in the standard case (all referenced IDs are "local" in a SensorData message/message sequence). 

Could there be also the case that there are HD maps and these are used/referenced by the DetectedTrafficSigns?

CarloVanDriestenBMW:
hm good question!
the map data and the detected lane would be merged by the sensor and the (also) merged traffic sign in the map would be assoziated with the (newly fused) detected lane?

carsten-kuebler:
It is easier to remove the messages and avoid the merge...

CarloVanDriestenBMW:
@carsten-kuebler I think we need the relevant lane and reference the DetectedLanes in here. The relevant direction is redundant because of the icon AND the fact that the association with a specific lane gives the direction relevance.

Issue #165, opened 2018-03-22
Remove interface version in SensorDetectionHeader and SensorView


carsten-kuebler:
Remove message field version from messages SensorDetectionHeader and SensorView.

pmai:
I‘m not sure whether that is sensible for SensorView, since this could end up a top-level message, like Groundtruth.

carsten-kuebler:
With this argument, SensorData should also get a version field?! 
@CarloVanDriestenBMW what do you think of Pierre's objections. It's about versioning and compatibility.

pmai:
Yes, I think all potential top level messages should have the version field (I totally missed that it didn’t already), otherwise one would have to depend on e.g. GT being present to determine the version...

carsten-kuebler:
Do you have a definition which messages could be top-level messages (GT, SensorData, SensorView, HostVehicle, ...)

CarloVanDriestenBMW:
@carsten-kuebler for now: GT, SD, SensorView

objections @pmai ?

pmai:
I'd add SensorInputConfiguration, since this is potentially exchanged prior to any of the other messages.

carsten-kuebler:
@CarloVanDriestenBMW So we must add version to SD (not SensorDetectionHeader --> remove version field), and FeatureData?! and SensorInputConfiguration.

CarloVanDriestenBMW:
Solved in #179

PR #169, opened 2018-03-22
Interface version removed in messages (issue #165)

carsten-kuebler:
Remove OSI interface version in messages. The interface version is not necessary due to the compatibility requirements of the OSI versioning.

### DIFF
--- a/osi_featuredata.proto
+++ b/osi_featuredata.proto
@@ -32,11 +32,6 @@ message FeatureData
 //
 message SensorDetectionHeader
 {
-    // Version number of used detection list messages ( \c SensorDetectionHeader, 
-    // \c LidarDetection, \c LidarDetectionList etc.).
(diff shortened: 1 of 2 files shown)


carsten-kuebler:
This PR addresses issue #165 .

SebastianSchmidtBMW:
@CarloVanDriestenBMW Our suggestion would be to keep the interface version in all potential toplevel messages as stated by pmai in #165 
But `DetectionHeaderIntefaceVersion` should be removed in favor of the generic `InterfaceVersion` from osi_version.proto

CarloVanDriestenBMW:
I agree, also on the generic "InterfaceVersion"

carsten-kuebler:
@SebastianSchmidtBMW I add a new PR #175 with version field for all top-level messages.

PR #171, opened 2018-03-22
Definition of SensorView for multiple sensors.

carsten-kuebler:
Change from one sensor to multiple sensors (if required).

### DIFF
--- a/osi_sensorview.proto
+++ b/osi_sensorview.proto
@@ -32,19 +32,19 @@ message SensorView
     
     // Radar-Specific SensorView
     //
-    optional RadarSensorView radar_sensor_view = 1000;
+    repeated RadarSensorView radar_sensor_view = 1000;
 
     // Lidar-Specific SensorView
     //
-    optional LidarSensorView lidar_sensor_view = 1001;
+    repeated LidarSensorView lidar_sensor_view = 1001;
 
     // Camera-Specific SensorView
     //
-    optional CameraSensorView camera_sensor_view = 1002;
+    repeated CameraSensorView camera_sensor_view = 1002;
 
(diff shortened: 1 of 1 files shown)


pmai:
Is this the right way to go? SensorView is the input to the sensor model and embedded in SensorData, which contains one sensor Id, mounting position, etc. So I don‘t see how there could be multiple views at this level, but not at the level of SensorView and SensorData?

carsten-kuebler:
One SensorData can hold detections from different sensors (after sensor data fusion).

pmai:
Yes, though I think that there are still some issues with SensorData to support this; however SensorView is purely input data (like GT) so it is not merged... If one wants to track SensorView in the output, then I think SensorView should be repeated not the sub-views. And GT potentially needs to be repeatable as well (different GT subsets). And SensorData needs fixing as well (Id, MountingPosition, ...).

carsten-kuebler:
Because SensorView is used as input, SensorView could be for a virtual sensor model with multiple low-level sensors.

pmai:
In that case you would also need multiple SensorInputConfiguration messages to spec the details for the views. Currently this would entail multiple SensorData inputs (or at the very least multiple SensorViews). I think the whole model of how multi-feeds are going to be handled needs rethinkink, because just making every other field repeatable does not cut it IMHO.

CarloVanDriestenBMW:
@pmai well... I guess we need the solution until Tuesday...

CarloVanDriestenBMW:
@carsten-kuebler : I had a call with @pmai on friday and we worked on a solution which will be pushed by him today. i guess it will include your suggestion here and the top leven version information. Thx for bringing up this topic!

Issue #154, opened 2018-03-05
Refine BasePolygon


CarloVanDriestenBMW:
5 people in a workshop have no clear understanding of this -> problem!

```
    // Usage as ground truth:
    // The two dimensional (flat) contour of the object. This is an extension of the concept of a 
    // bounding box as defined by Dimension3d. The contour is the projection of the object's outline 
    // onto the z-plane in the object frame (independent of its current position and orientation). 
    //
    // Usage as sensor data:
    // The polygon describes the visible part of the object's contour.
    //
    // General definitions:
    // The polygon is defined in the local object frame: x pointing forward and y to the left.
    // The origin is the center of the bounding box.
    // As ground truth, the polygon is closed by connecting the last with the first point. Therefore  
    // these two points must be different. The polygon must consist of at least three points.
    // As sensor data, however, the polygon is open.
    // The polygon is defined counter-clockwise.
repeated Vector2d base_polygon = 4;
```

carsten-kuebler:
@CarloVanDriestenBMW It is the outline of the object's shadow cast at the bottom of the bounding box (objects xy-plane). You could use this information to get a little more free space information. 
At the moment this additional information is not part of the standadisation activities for real sensors.

SebastianSchmidtBMW:
@CarloVanDriestenBMW the definition above is unambigious so we would keep it. What has to be discussed is the necessity of the `base_ploygon` field. 
If it is not part of the standardisation plans, it should only be kept if there is a specific need for it by a manufacturer. See #15

carsten-kuebler:
Keep in mind, that the projection is on the xy-plane of the local frame. The local xy-plane is not necessarily the GT xy-plane.

CarloVanDriestenBMW:
@SebastianSchmidtBMW should we add a comment, that the hight is the same as the parents bounding box hight?

PR #172, opened 2018-03-23
base_polygon removed

carsten-kuebler:
This PR addresses issue #154. This PR is a suggestion.

Remove base_polygon in BaseStationary and BaseMoving.


CarloVanDriestenBMW:
I actually do not want to remove it, but place a better equivalent instead of the 2D polygon -> 3D Polygon

carsten-kuebler:
Really? 3D is ambiguous, isn't it? 2D projection of the object's outline is unambiguous. In robotics you would use multiple cylinders, sheres and cones to wrap the object.

CarloVanDriestenBMW:
We will close this one and update the documentation for the polygon2D

PR #175, opened 2018-03-23
Add OSI Version field to all top-level messages

carsten-kuebler:
This PR addresses issue #165.
This PR is in contradiction to PR #169. PR #169 would be obsolete.

Add OSI version message to all top-level messages and remove local version message (from SensorDetectionHeader, because FeatureData message is top-level message)


CarloVanDriestenBMW:
@pmai you would include this in your change today, right? When can I expect it?

PR #179, opened 2018-03-26
Seperate SensorView from SensorData

pmai:
This commit completely seperates SensorView as the new input interface
for sensor models from SensorData as the output interface for sensor
models. It also regularizes sensor_id and mounting_position treatment
and adds support for virtual sensors that contain multiple physical
detectors (actual physical sensors).  All potential top-level messages
now also contain a proper interface version field.

This PR should make obsolete PRs #169, #175 and #171.


osi_sensordata.proto
```diff
@@ -60,143 +68,137 @@ message SensorData
...
-    // 
-    repeated DetectedLane lane = 13;
-        
+    //
+    repeated DetectedLane lane = 17;
+
+    // The ID of the lane the host vehicle travels on in the DetectedLane data,
+    // relative to the sensor.
+    //
+    optional Identifier host_vehicle_lane_id = 18;
```
CarloVanDriestenBMW: Where is the host_vehicle_data gone? Shouldn't this be part of it?

CarloVanDriestenBMW: found it! Isn't this some "self-localization information" being part of the HostVehicleData? What do you think?

osi_sensorview.proto
```diff
@@ -29,33 +48,94 @@ message SensorView
...
+    // that where exchanged during the simulation initialization phase.
+    //
+    // \note The host vehicle is always contained in the ground truth provided,
+    // regardless of any filtering.
+    //
+    optional GroundTruth global_ground_truth = 7;
+
+    // The ID of the host vehicle in the ground_truth data.
+    //
+    optional Identifier host_vehicle_id = 8;
```
CarloVanDriestenBMW: also HostVehicleData?

pmai: I seperated host_vehicle_data and host_vehicle_id since they really have nothing in common: host_vehicle_data is proper new input, that is not duplicated in global_ground_truth, whereas host_vehicle_id really is more or less rendundant data that might really be included in ground_truth (and to some degree is since global_ground_truth already contains the host_vehicle flag).

osi_sensordata.proto
pmai: The Host Vehicle Data is only contained in SensorView, since it is data that the vehicle "knows" about itself, and not something that a sensor would output on its own (we currently have no support for modeling e.g. GPS receivers as sensor models, i.e. producing SensorData). Somewhere along the further processing tool-chain one would imagine other modules taking SensorData and GPS and other data into account to update this internal information.

If one wanted to give sensor models (or fusion models) the ability to express an opinion on the vehicle location, then this should be a new field (estimated host vehicle location, or something of this sort), since host vehicle data could at some point in time contain e.g. brake pedal position, etc., which is not something that ADAS sensors have any opinions about.

osi_sensordata.proto
```diff
@@ -60,143 +68,137 @@ message SensorData
...
-    // 
-    optional DetectedEntityHeader traffic_light_header = 102;
+    //
+    optional DetectedEntityHeader traffic_light_header = 12;
 
     // The list of traffic lights detected by the sensor
-    // 
-    repeated DetectedTrafficLight traffic_light = 11;
+    //
+    repeated DetectedTrafficLight traffic_light = 13;
```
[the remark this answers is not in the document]
pmai: See above, I tried to keep all naming stuff outside of this PR.

osi_sensordata.proto
```diff
@@ -60,143 +68,137 @@ message SensorData
...
     // The list of objects detected by the sensor as perceived by the sensor.
     //
     repeated DetectedObject object = 9;
-    
+
     // General information about the \c DetectedTrafficSign .
-    // 
-    optional DetectedEntityHeader traffic_sign_header = 101;
+    //
+    optional DetectedEntityHeader traffic_sign_header = 10;
```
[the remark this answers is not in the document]
pmai: I didn't touch the naming stuff, since I think that should be a seperate PR so as not to confuse everyone (I've included the renumberings, because that is a more pervasive problem, and I had to renumber in any case). I'm not that happy with the whole DetectedEntityHeader field naming, and semantics, since it sounds like a header for each detected entity, whereas it really is only a header for each class of entities...

osi_sensordata.proto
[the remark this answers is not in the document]
pmai: As an aside, since we discussed this on the phone: Given the code that protoc generates for repeated fields, using the singular might be slightly less misleading than the plural (see discussion in https://stackoverflow.com/questions/4121175/protobuf-naming-conventions point 4):

For

    repeated DetectedTrafficLight traffic_light = 13;

the generated methods might be

    ::traffic_light(int index), ::traffic_light(), ::add_traffic_light()

this is slightly misleading due to the ::traffic_light(int index) vs. ::traffic_light() accessors (get one or all traffic_lights), however the plural version is also at least as misleading, where every method is called traffic_lights:

    ::traffic_lights(int index), ::traffic_lights(), ::add_traffic_lights()

So basically protobufs suck...

osi_sensordata.proto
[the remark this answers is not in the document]
carsten-kuebler: @CarloVanDriestenBMW I would leave traffic_sign and traffic_sign_header. Otherwise we have to add the whole scope of a field to all field names. 
@pmai Only DetectedHeader?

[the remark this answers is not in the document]
pmai: @carsten-kuebler, @CarloVanDriestenBMW  I'm thinking of something like DetectedEntityClassHeader and traffic_sign_class_header that makes the per-class character clearer, but those names are rather unwieldy as well, so I'm inclined to leave this as is for now...

osi_sensorview.proto
```diff
@@ -116,12 +196,12 @@ message RadarSensorView
...
 // Lidar-Specific Sensor View Data
-message LidarSensorView 
+message LidarSensorView
 {
-    // Lidar Input Configuration valid at the time the data was created
+    // Lidar View Configuration valid at the time the data was created
     //
-    optional LidarSensorInputConfiguration lidar_sensor_input_configuration = 1;
-    
+    optional LidarSensorViewConfiguration lidar_sensor_view_configuration = 1;
```
carsten-kuebler: The field names of this message are:
- lidar_sensor_view_configuration
- reflections

Is it more clear, if we rename it, isn't it?
- configuration
- reflections

or (but not better)

- lidar_sensor_view_configuration
- lidar_sensor_view_reflections

This also applies to the other XXXSensorView messages.

pmai: I think shortening to just `configuration` is too short, since it is not just any configuration but the configuration of this view (there might be other configurations, e.g. the configuration of detected objects, ...). The reflections OTOH are not related to the view (they are not reflections of the view but `reflections` of the outside world).

So if one wanted to avoid the mouthful `lidar_sensor_view_configuration`, I'd propose shortening to `view_configuration`, even at the risk of some ambiguity, since there is also the more general SensorViewConfiguration.

osi_sensorviewconfiguration.proto
```diff
@@ -0,0 +1,632 @@
...
+        // point (positive dB is gain, negative dB is attenuation)
+        //
+        // Unit: [dB]
+        optional double response = 3;
+    }
+}
+
+//
+// \brief The configuration settings for the Lidar Sensor Input to be provided
+// by the environment simulation.
```
carsten-kuebler: Lidar Sensor View not Input

osi_sensorviewconfiguration.proto
```diff
@@ -0,0 +1,632 @@
...
+    // through frequency adjustments.  For material and geometry interaction
+    // purposes the frequency is also relevant.
+    //
+    // Unit: [Hz]
+    optional double emitter_frequency = 9;
+}
+
+//
+// \brief The configuration settings for the Camera Sensor Input to be provided
+// by the environment simulation.
```
carsten-kuebler: View (typo)

osi_sensorviewconfiguration.proto
```diff
@@ -0,0 +1,632 @@
...
+        // BGR, BAYER_RGGB/GBRG/GRBG/...), non-BAYER filters, non-linear
+        // encodings, ...
+    }
+
+    // TBD: Optical (and other) effects to apply to image, etc.
+}
+
+//
+// \brief The configuration settings for the Ultrasonic Sensor Input to be
+// provided by the environment simulation.
```
carsten-kuebler: View (Typo)

osi_sensorviewconfiguration.proto
```diff
@@ -0,0 +1,632 @@
...
+
+    // The root mean squared error of the mounting position.
+    //
+    optional MountingPosition mounting_position_rmse = 3;
+
+    // Field of View in horizontal orientation of the physical sensor
+    //
+    // Viewing range: [-field_of_view_horizontal/2, field_of_view_horizontal/2]
+    // azimuth in the sensor frame as defined in \c OSI:Spherical3d.
+    // Unit: [rad]
```
carsten-kuebler: Should we add information about distortion? Ist the field of view the min. or max. field of view (as example barrel distortion for cameras)?

osi_sensordata.proto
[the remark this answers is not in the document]
carsten-kuebler: If we don't use a message Message, we could use Message as a fixed expression.
DetectedMessage(s)Header

osi_sensordata.proto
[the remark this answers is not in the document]
CarloVanDriestenBMW: We stick to the plural now. At least you can read it in a better way in the .proto file

osi_sensordata.proto
CarloVanDriestenBMW: If you look in the osi_hostvehicledata.proto you find that it is defined to be the place for an "estimated position"!
It is not the Interface for the brake pedal atm but in fact the one for the GPS, assigned lane, etc.... it is opinion only? :dancer: 
I think it must be part of SensorData

osi_sensorview.proto
CarloVanDriestenBMW: see comment above -  the information on the perfect host vehicle position is in the vehicles including the flag "host vehicle"

osi_sensordata.proto
pmai: Given that I wrote that: Sure it is an estimated position, since the vehicle itself has no way of knowing the proper true position. But as stated in that file's header:

    // \brief Interface for host vehicle data that is available to sensors and
    // other functions due to host vehicle's internal communication.

I.e. this is the input side that's available to sensors due to internal communication, it was not meant for the output side of a single sensor. It is the current best knowledge of the vehicle itself, like its knowledge of its break pedal position, RPMs, etc.

I think any estimation by a single sensor in SensorData should follow the pattern for all other detected output we have (i.e. it should be added as single attributes to SensorData). I'll push a commit for that.

osi_sensordata.proto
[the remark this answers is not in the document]
pmai: I'd still leave this PR as is, since the plural/singular stuff should be dealt with in a global commit I think (it is mixed up enough as it currently stands).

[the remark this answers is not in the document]
carsten-kuebler: @pmai I changed this (plural->singular) in the review PR. I can make the merge manually.

PR #182, opened 2018-03-27
Allow repeated SensorView copies in SensorData

pmai:
This is relevant for e.g. sensor fusion models that merge multiple
SensorData inputs into one SensorData output.


CarloVanDriestenBMW:
I am actually a bit confused by the "RadarSensorViewConfiguration" copy. I thought we only need a few values from that in every time step?

pmai:
The overhead of the config view is very slight compared to the radar data itself, and without the config information the raw data does not necessarily make sense: Consider the CameraSensorView data, which is just raw binary data, with the format completely described in the config. This also aids in getting the right data for multiple views, since they contain the configs...

PR #187, opened 2018-03-27
Cmake for v2.2.1

FabeleAltran:
The cmake file from current master branch is reuse and adjusted to allow several OSI versions on the same computer.

The main idea is to install the into a directory that takes the version into account.
This draft uses the major version described inside the version file to configure the installation path.

This will install any Version 2.x.x into the following directories:
Libraries: /usr/local/lib/osi2
Headers: /usr/local/include/osi2

The "/usr/local/" path is chosen to respect the Linux file system architecture.
The OSI resources are not yet a system library, so i think the local is the better suitable installation path.
(See http://refspecs.linuxfoundation.org/FHS_2.3/fhs-2.3.html#THEUSRHIERARCHY for details)

@pmai Please comment and change anything you deem necessary


FabeleAltran:
@pmai Please wait with anything i need to check some details with @CarloVanDriestenBMW

Issue #197, opened 2018-03-28
Merge Vehicle in MovingObject


CarloVanDriestenBMW:
We have to find a clever way to integrate the Vehicle message into the MovingObject message. There no real reason, why these two should be different!

CarloVanDriestenBMW:
Important: There is currently no ClassProbability for MovingObjects in DetectedObjects with PR #176 ! (Bug)

carsten-kuebler:
We could rename ClassProbability to ClassProbabilityVehicle and add a similar ClassPropabilityMovingObjects. Is it ok?


StationaryObjects include all stationary objects except TrafficLights, TrafficSigns.
Why would you merge Vehicle and MovingObjects?

For me StationaryObjects and MovingObjects covers all "secondary" objects.

carsten-kuebler:
see also discussion #200

Issue #163, opened 2018-03-20
Major Version Todos: Renumber OSI message fields


carsten-kuebler:
Todo for next major release.

CarloVanDriestenBMW:
@carsten-kuebler I think we will do this as the last step before releasing tomorrow.

carsten-kuebler:
And plural s...

CarloVanDriestenBMW:
@carsten-kuebler  we will stick to singular now:
Reason: If we change a field from optional to repeated we do not have to cha nge the name and have compatibility! Kudos @pmai

carsten-kuebler:
So I will renumber all messages. @CarloVanDriestenBMW @pmai Do we need fixed numbers for fields or can we renumber all messages?

carsten-kuebler:
Renumbering done. 
Plural s not done.
Remove Landmarks - open PR.
Relevant xx - open PR.
Version in top-Level messages misses? FeatureData has its own Version message - change?

carsten-kuebler:
all "open issues" are included in PR #200 (the final PR 🥇 )

PR #200, opened 2018-03-28
Improvement/harmonize detectedobjects (solves #197)

CarloVanDriestenBMW:
* Harmonized proto indicies in every file #163 
* Merged #199 in this PR
* Merged MovingObjects and Vehicles
* Same structure for Detected*Moving/Stationary*Object as in DetectedTrafficSigns etc
* Removed DetectedLandmarks (cleanup)
* Removed LaneFraming
* Added limiting_structure_id to LaneBoundaries
* Removed SensorDataList

@pmai Please look into it very thoroughly! Thx!

@carsten-kuebler A review from you would also be very nice!


carsten-kuebler:
@pmai Why do you index some fields starting with 1000? Is there a reason?

pmai:
I use this approach when some fields are just sub-views/configs so that we have space to add „normal“ attribute fields to the top-level  message later on without needing to interleave them with those sub-views. Not strictly necessary but aids in keeping the fields in a somewhat readable order... Not that I think this whole field numbering business is worthy of human attention, but ... protobufs...

carsten-kuebler:
@pmai But we don't have to keep to a monotonously increasing order, do we? We can insert fields that have a higher index. A renumbering of the indices should then take place in each major release.

carsten-kuebler:
@CarloVanDriestenBMW I'm not happy with the merging of Vehicle with MovingObject and that message StationaryObject is also included in object.proto. From my point of view StationaryObject should be moved to landmarks.proto and object.proto renamed to movingobject.proto. 

For me the question arises then whether TrafficLight, TrafficSign and RoadMarking should also be implemented as extensions of StationaryObject (as Vehicle and MovingObject).

As a further question I asked myself whether there are other moving objects and there came to me variable advertising pillar and advertising display panels (movies as "2D moving" objects). Do we need these objects in order to simulate "2D moving objects" through animated "advertising films" on displays?

osi_object.proto
```diff
@@ -273,31 +274,31 @@ message MovingObject
...
+            // Definition: Hatchback car with maximum length 4 m.
             //
             TYPE_SMALL_CAR = 2;
 
             // Vehicle is a medium car.
             //
-            // Definition:
+            // Definition: Hatchback or sedan with lenght between 4.5 and 5 m.
             //
             TYPE_MEDIUM_CAR = 3;
```
carsten-kuebler: should we reorder medium and compact car?

pmai:
@carsten-kuebler yes, we don't need to use monotonically increasing field numbers, which is the other way of dealing with the interleaving; however since the interval between major releases should be fairly long (hopefully), there is a long time where you have to deal with files with non-monotonically increasing numbers, which is a bit error prone and unwieldy.

That said I don't have strong opinions one way or the other (except that we should ditch ProtoBufs in any case), feel free to renumber as wanted.

osi_lane.proto
```diff
@@ -346,6 +317,12 @@ message LaneBoundary
     //
     optional Color color = 4;
 
+    // The ids of \c StationaryObjects which limit the corresponding lane.
+    // This field must be set if the \c LaneBoundary.type is set to
+    // # TYPE_STRUCTURE
```
AhmedMzid: Is there any reason for the '#' before TYPE_STRUCTURE?
bildschirmfoto vom 2018-03-29 16-42-53

CarloVanDriestenBMW: @carsten-kuebler can you check this?

carsten-kuebler:
# is necessary for doxygen links.  But without space. 

Gesendet von IBM Verse


   AhmedMzid --- Re: [OpenSimulationInterface/open-simulation-interface] Improvement/harmonize detectedobjects (solves #197) (#200) --- 
    Von:"AhmedMzid" <notifications@github.com>An:"OpenSimulationInterface/open-simulation-interface" <open-simulation-interface@noreply.github.com>Kopie:"carsten-kuebler" <carsten.kuebler@twt-gmbh.de>, "Mention" <mention@noreply.github.com>Datum:Do. 29.03.2018 16:44Betreff:Re: [OpenSimulationInterface/open-simulation-interface] Improvement/harmonize detectedobjects (solves #197) (#200)
  
  @AhmedMzid commented on this pull request.  
 In osi_lane.proto: 
 > @@ -346,6 +317,12 @@ message LaneBoundary
     //
     optional Color color = 4;
 
+    // The ids of \c StationaryObjects which limit the corresponding lane.
+    // This field must be set if the \c LaneBoundary.type is set to
+    // # TYPE_STRUCTURE
 
 Is there any reason for the '#' before TYPE_STRUCTURE?

pmai:
@CarloVanDriestenBMW Does the LaneBoundary/DetectedLaneBoundary and Lane/DetectedLaneBoundary split look right to you? If so, I'll rebase this and merge...

PR #204, opened 2018-04-09
Correct cmake version check and install

FabeleAltran:
This commit allows the user to install version 2.x and 3.x on the same system and use it through the cmake command "find_package".

The mayor version requested will always be check first and is a required input to the command.
Example: 
```cmake
# Any cmake version installed
find_package(open_simulation_interface) # -> Produces fatal error

# only version 2.x installed
find_package(open_simulation_interface 2) # -> Select OSI version 2.x.x
find_package(open_simulation_interface 3) # -> incompatible version error 
```
 This is required due to the compatibility break between versions 2.x and 3.x.

A similar patch will be provided to the current master branch (version 3.x)

@pmai If the changes are good for you and accepted, i will create pull request for the OSI 3.x version. If you want any additional changes please comment and i will implement them.
@CarloVanDriestenBMW FYI


FabeleAltran:
@CarloVanDriestenBMW The milestone is wrong! This patch is only for the 2.x versions!
For the 3.x version a separate PR will be created.

PR #207, opened 2018-04-13
Fix graphviz link

nkm8:
http://www.graphviz.org/Download_windows.php currently leads to a 404 page; updated readme with new link.


CarloVanDriestenBMW:
THX!
:)

PR #228, opened 2018-04-27
Extend documentation center line (resolves #225)

### LINKED ISSUE #225: Adjust description of centerline in osi_lane.proto
I think the center line as currently described is a bit outdated. The description should be adjusted.

Reason 1: The center line in the Ground Truth is described as ideal trajectory a vehicle should follow. However, on crossings no centerlines exist with the new lane model. This leads to the requirement for other models to be able handle situations without a center line, i.e. the center line is only a convenience feature and should be documented as such.
Reason 2: If a lane narrows (e.g. because of a cunstruction site), the center line highly depends on physical dimensions of the vehicles (mainly the width). It is questionable how reliable a center line then becomes.

carsten-kuebler:
This PR address resolves #225

Enhance the documentation of center line.


CarloVanDriestenBMW:
Should we consider removing Centerlines from the GT and add it to SensorData only? Then it would be model dependend how the CL is calculated?

HendrikAmelunxen:
I agree to this. The "virtual" centerline information (= "the line that a typical vehicle is supposed to follow") does not fit to the groundtruth information.

AndreHildebrandt:
I also agree with the removal of the centerline. I already gave some reasons (#225 ). For the moment I don't see the benefit, but the effort of generating and transmitting point sequences, where the validity is in the best case questionable.

carsten-kuebler:
@LudwigFriedmannBMW I changed the documentation. I moved the information to a note. Please note that in case of discontinuities in the roadway width, the center line may also have discontinuities. The center line was not intended to be an exact geometric center line.

osi_lane.proto
```diff
@@ -54,56 +54,58 @@ message Lane
...
-        // direction defined by the sequence of points forming the center line.
+        // arbitrary distances. When the points are pairwise linearly connected,
+        // the lateral distance to the real ideal line (as used by the
+        // simulation environment internally) must not exceed 5cm. As shown in
+        // the following image:
+        //
+        // \image html line_approximation_error.png "Approximation error as
+        // green line."
+        //
+        // \note The center line is the line that a typical vehicle follows more
```
LudwigFriedmannBMW: I would change this paragraph as follows:

The center line is the line that a typical vehicle follows more or less (depending on the situation, a little more to the left or right of the center line). The intended direction of travel on the lane is given by the direction defined by the sequence of points forming the center line.

carsten-kuebler:
@LudwigFriedmannBMW. I also refer to centerline_is_driving_direction.

LudwigFriedmannBMW:
Hello @carsten-kuebler, can you please resolve the conflicts?

carsten-kuebler:
@LudwigFriedmannBMW merged.

PR #231, opened 2018-05-02
Extended documentation traffic sign/road markings (resolves #229) (resolves #230)

### LINKED ISSUE #229: Add Sign 297 "direction arrows" in documentation
Hello,
as far as we understood Roadmarkings Sign 297 "direction arrows` (Richtungspfeile) would be covered by roadmarkings with traffic_main_sign_type for example TYPE_PRESCRIBED_STRAIGHT_AHEAD or  TYPE_PRESCRIBED_RIGHT_TURN_AND_STRAIGHT_AHEAD etc.

If so, it would be good to add these Sign-Numbers in Proto-Files so text search runs more easily

Thanks

### LINKED ISSUE #230: Add Sign 294 "stopping line" in Documentation
Hello,

it would be helpful if the OSI documentation makes it more clear how a stoppingline (Haltlinie) has to be modelled.

As far as we understood: Lane  with TYPE_INTERSECTION and one or multiple "free_lane_boundary" which references roadmarking. Please also reference Sign-Number 294.

Thanks

carsten-kuebler:
Remove doxygen windows bug (roadmarking types are not correct).
Add StVO road markings as symbolic signs in documentation of traffic signs (resolves #229 and resolves #230 and additional road markings).

### DIFF
--- a/osi_trafficsign.proto
+++ b/osi_trafficsign.proto
@@ -259,28 +259,28 @@ message TrafficSign
                 TYPE_RIGHT_BEFORE_LEFT_NEXT_INTERSECTION = 3;
 
                 // Warning sign for a left turn
-                // (StVO 103.1). Right: #TYPE_TURN_RIGHT
+                // (StVO 103.1). Right: \c #TYPE_TURN_RIGHT
                 //
                 // See e.g.: https://www.dvr.de/bilder/stvo/gt/103-10.png
                 //
                 TYPE_TURN_LEFT = 4;
 
                 // Warning sign for a right turn
-                // (StVO 103.2). Left: #TYPE_TURN_LEFT
+                // (StVO 103.2). Left: \c #TYPE_TURN_LEFT
                 //
                 // See e.g.:
                 //
                 TYPE_TURN_RIGHT = 5;
 
                 // Warning sign for a double turn (first left turn)
-                // (StVO 105.1). Right: #TYPE_DOUBLE_TURN_RIGHT
+                // (StVO 105.1). Right: \c #TYPE_DOUBLE_TURN_RIGHT
                 //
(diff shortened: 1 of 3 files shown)


osi_trafficsign.proto
```diff
@@ -347,90 +347,144 @@ message TrafficSign
...
+                //
+                // See e.g.: https://www.dvr.de/bilder/stvo/gt/297.png
+                //
                 TYPE_PRESCRIBED_STRAIGHT_AHEAD = 22;
 
                 // Prescribed left way
-                // (StVO 211.1). Right: #TYPE_PRESCRIBED_RIGHT_WAY
-                //
+                // (StVO 211.1). Right: \c #TYPE_PRESCRIBED_RIGHT_WAY
+                //
```
CarloVanDriestenBMW: Do we have a test checking for spaces before line break?

carsten-kuebler: No. I created unit tests mainly for errors. The space at the end of a line is more of an unwanted character and rather reduces the file size. A space at the end of a line doesn't cause an error, does it? Of course, it is very easy to add such an unit test. I suspect that every initial checkin will lead to an error with this unit test. Do we want to set up these high barriers for a check-in at GitHub? 

I personally support unit tests. Should we add this unit test?

carsten-kuebler: clang-format works for proto files, clang-check doesn't work for proto files (c,c++ works fine). We could use clang-format to reformat the files. Should we add active code-modification and github checkin via travis ci?

CarloVanDriestenBMW: yes, do we need a TravisCI user for that? The commit should be on the branch of the PR.

carsten-kuebler: Yes. We can use the same mechanism as for the auto-documentation. Drawback: We could end in an infinite loop 8-) of auto-check-ins.

PR #245, opened 2018-06-04
Add proto3 support via script to map file syntax

pmai:
This change allows usage of the OSI definitions on proto3 syntax,
while remaining on-the-wire-compatible with the official proto2
syntax.

Note that Travis CI build is also amended to build both proto2
and proto3 versions so that portability fails will be detected early.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,6 +75,7 @@ script:
     - cd build
     - cmake --build . --target install
     - cd ..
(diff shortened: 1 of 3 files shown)


CarloVanDriestenBMW:
@HannesSchneiderAVL could you check if this fullfills your requirements, please?
Thank you!

PR #249, opened 2018-06-11
    This commit *was not* tested wit C++. Proceed with caution.

cszyszka:
This commit fixes compatibility with python3 for the code generated with protoc for python.
    Proto definition files were moved to osi3 direcotry.
    Proto files were modified in the way that the include stattment now specifis osi3 packge.
    e.g.
    [OLD] import "osi_version.proto";
    changed to
    [NEW] import "osi3/osi_version.proto";
    This changes fix the issue for python generted code.

    For full details see the discussion in the protobuf repository.
    https://github.com/google/protobuf/issues/1491

    More details on this work will be commented in
    https://github.com/google/protobuf/issues/1491


CarloVanDriestenBMW:
@cszyszka please crosscheck with your team in order to get the CI running in order to merge this PR, thank you!

pmai:
So may I summarize: Protobufs suck... who would've guessed.

We cannot move the protos to a subdirectory, because the CMake ProtoBuf support does not work correctly if the proto files are in a subdirectory (and other issues). And we cannot keep it in the same directory, because protobuf python3 build support does not work if its not in a subdirectory. Lovely. I should have stayed in the 80s.

Likely only solution for now: Copy *.proto to temp subdir prior to running the python side of the build.

cszyszka:
> Likely only solution for now: Copy *.proto to temp subdir prior to running the python side of the build.
This is not a problem what about changes to import statements, will that work with CMAKE/CPP ?

pmai:
Probably need to change the imports on the fly too, i.e. similar to the proto3 converter script (that we still rely on sh and sed in this day and age makes me sad, too):
```
for f in osi_*.proto
do
    sed -e 's:import "osi_:import "osi3/osi_:' $f > osi3/$f
done
```

As an intermediate build step between the c++ build and the python build...

jr7:
Don't know if this helps but this is a similar issue:  https://github.com/google/protobuf/issues/762
Unfortunately  im quiet busy at the moment, sorry about  that. BR

cszyszka:
> Don't know if this helps but this is a similar issue: google/protobuf#762


@jr7  This issue references another one which I used to solve the issue with import statements in generated code for python.  https://github.com/google/protobuf/issues/1491

Issue #243, opened 2018-05-30
OSI library import not working in Python 3


jr7:
Hello, 

the OSI Library for Python3 is not working on my computer,
was this tested before ? what is the default Python version you are using ? 

For example `import osi.osi_sensordata_pb2`  will give an Exeception :

    ModuleNotFoundError: No module named 'osi_common_pb2'

Please notice this is version 2.2.1 but this also occurs in 3.0

    In [1]: import osi3
    ModuleNotFoundError: No module named 'osi_version_pb2'

My assumption is that this is due to Python 3 not allowing relative imports.
Whats your opinion on this ? Thank you for any feedback. 
BR Jonas Ruebsam

jr7:
Maybe somebody can check this on their machine as well. 
So we can  exclude some setup related issues.

cszyszka:
@jr7 I'm checking this could you tell which version of protobuf are you using?

jr7:
@cszyszka  its 3.5.2

cszyszka:
I created new branch which solves the issue. 
https://github.com/OpenSimulationInterface/open-simulation-interface/pull/249

What was needed was:
* move the proto files to osi3 directory.
* edit import statements in definitions to include osi3 package import "osi3/osi_version.proto";
* edit setup.py instalation script 

With this version I can import and work with OSI3. With this test I cab verify it works. 

```python
import osi3.osi_sensordata_pb2
sd = osi3.osi_sensordata_pb2.SensorData()
sd.timestamp.seconds = 1
sd.timestamp.nanos = 2
sd
```

LudwigFriedmannBMW:
Hello @cszyszka, since #249 is completed, may this issue also be closed?

jr7:
I think this is still an issue, at least the last time I tried  it didn't work..

pmai:
Yes the issue is still extant, since the original pull request could not be merged, since this would break the cmake build process. I've fixed the python3 issues within setup.py in #264, so please check whether that really fixes it for you.

jr7:
Ok, I tested with the current master and it seems to work  👍

PR #264, opened 2018-08-10
Fix python3 relative import issue and setup.py

pmai:
This change builds the protobuf files inside the osi3 package path
for Python2 to ensure that relative imports work, while not breaking
the general cmake build setup.  It also moves the protobuf generation
to the build command phase for tidiness.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,10 +4,11 @@
 import os
 import subprocess
 import sys
+import re
 from distutils.spawn import find_executable
 
 from setuptools import setup
-from setuptools.command.install import install
+from setuptools.command.build_py import build_py
 
 # configure the version number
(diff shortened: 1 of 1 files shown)


pmai:
This should fix issue #243.

mshaheerm:
I am working with a module.py which is present in C:\Users\sham\newcarla\carla\PythonClient and have copied all the generated proto_to_python.py in C:\Users\sham\newcarla\carla\PythonClient\proto_generated but there is an import error. i tried running the setup.py in C:\Users\sham\Downloads\open-simulation-interface-master\open-simulation-interface-master and it created a folder osi3 which just has init.py in it. How can i use it as a package for my imports?

Issue #234, opened 2018-05-04
Speed definition in ISO 23150 vs. OSI


CarloVanDriestenBMW:
The OSI definition is the following:

> The relative velocity of the moving object w.r.t. its parent frame and parent velocity. The velocity becomes global/absolute if the parent frame does not move.
> 
> position (t) := position (t-dt)+ velocity *dt

The ISO says the following:

Parameter | Content | Data format | Unit
-- | -- | -- | --
X   velocity | Absolute velocity of the object   along the x-axis of the ego-vehicle coordinate system. |   | [m/s]
Y   velocity | Absolute   velocity of the object along the y-axis of the ego-vehicle coordinate system. |   | [m/s]
Z   velocity | Absolute   velocity of the object along the z-axis of the ego-vehicle coordinate system. |   | [m/s]
X acceleration | Absolute   acceleration of the object along the x-axis of the ego-vehicle coordinate   system. |   | [m/s²]
Y   acceleration | Absolute   acceleration of the object along the y-axis of the ego-vehicle coordinate   system. |   | [m/s²]
Z   acceleration | Absolute acceleration of the   object along the z-axis of the ego-vehicle coordinate system. |   | [m/s²]

TimFrickeBMW:
It is important to note that the cited OSI definition relates to "optional Vector3d velocity" and that there is a corresponding "optional Vector3d acceleration". The OSI definitions clearly state that "velocity" and "acceleration" are both "relative".

The fact that the ISO (- which one are you referring to? I cannot find these definitions in ISO8855. - ) defines velocities and accelerations as "absolute" does not bother me. In fact, I find the term "absolute" too vague: absolute w.r.t. which frame? Which frame am I supposed to assume "inertial"?

Strictly speaking, two aspects need to be specified: the frame w.r.t. which each derivation d/dt is to be performed (e.g., acceleration is the position derived twice w.r.t. the host vehicle frame) and the frame in which the vector is noted. I think, both aspects are covered by the OSI definition, but I would nonetheless consider rephrasing in future releases for better clarity:

"The relative velocity of the moving object **w.r.t. the parent frame, noted in the parent frame**. The velocity becomes global/absolute if the parent frame does not move."

Apart from that, I wonder why the OSI definition continues with the following line:

"position (t) := position (t-dt)+ velocity *dt"

[...]

PR #267, opened 2018-08-17
Update Definition solves #234

carsten-kuebler:
This PR refers to issue #234


carsten-kuebler:
@LudwigFriedmannBMW I updated the documentation and checked the other definitions (acceleration and orientation rate). I add notes to sensor data and detected objects.

osi_common.proto
```diff
@@ -346,7 +345,7 @@ message BaseMoving
 
     // The relative orientation rate of the moving object w.r.t. its parent
     // frame and parent orientation rate in the center point of the bounding box
-    // (origin of the bounding box frame).
+    // (origin of the bounding box frame), noted in the parent frame.
```
LudwigFriedmannBMW: The orientation rate becomes global/absolute if the parent frame is not rotating.

osi_common.proto
```diff
@@ -325,18 +325,17 @@ message BaseMoving
     //
```
LudwigFriedmannBMW: Please add to line 265 & 319: The relative orientation of the stationary object w.r.t. its parent frame, noted in the parent frame. The orientation becomes global/absolute if the parent frame is inertial .

carsten-kuebler:
@LudwigFriedmannBMW I also add (all parent frames up to ground truth)

PR #275, opened 2018-10-01
Add proj4 string to GroundTruth fixes #276

### LINKED ISSUE #276: Add proj4 string to GroundTruth in order to geo-reference the global coordinate frame
A colleague of mine started a PR (#275) concerning the extension of the GroundTruth by a geo-reference string. Please feel free to discuss this idea.

PhilippHamerBMW:
Using field number 16 because the projection string will not be a small field, so `14,15` can be used in future releases for `repeated` or smaller fields.

### DIFF
--- a/osi_groundtruth.proto
+++ b/osi_groundtruth.proto
@@ -131,4 +131,14 @@ message GroundTruth
     // - [2] [Wikipedia ISO 3166/1] (https://en.wikipedia.org/wiki/ISO_3166-1)
     //
     optional uint32 country_code = 13;
+
+    // Projection string that allows to transform all coordinates in GroundTruth
+    // into a different cartographic projection.
+    //
+    // The string follows the proj.4 project rules [1].
+    //
+    // \par References:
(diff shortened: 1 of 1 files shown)


CarloVanDriestenBMW:
@pmai would you expect this in the sensor view or the ground truth?

pmai:
I think groundtruth is the proper place, or it might in the future be something to be moved to setup messages if it’s constant over time.

LudwigFriedmannBMW:
@pmai and @CarloVanDriestenBMW: What do you think about the field number? Does it make sense to choose a non-consecutive value due to performance reasons? I'd like to finish this PR in order to be able to discuss new ones (i.e. #277)

pmai:
I don't think we should think too much about performance for this, but generally speaking this PR could be changed to use id 14, since 16 does not seem to buy us anything...

PhilippHamerBMW:
Reserving field numbers below 16 for frequently used fields is recommended by the protobuf language guide due to the resulting smaller byte size of the message.
But using consecutive numbers is also fine for me.

PR #288, opened 2018-11-06
Update osi_sensorview.proto

marcin-piatek:
Small change in definition of time of flight


pmai:
All units in OSI are SI base units, unless domain-specific precedent exists (mostly in the environmental stuff).  That said a change would only be possible in OSI 4.x or later since this is completely non-backward compatible.

pmai:
Closing for now for reasons above.

PR #298, opened 2018-12-19
Camera detecions

carsten-kuebler:
@CarloVanDriestenBMW: Proposal for camera detections.

### DIFF
--- a/osi_featuredata.proto
+++ b/osi_featuredata.proto
@@ -707,7 +707,7 @@ message CameraDetection
     // The probability for this classification is at least
     // \c #shape_classification_probability.
     //
-    optional bool shape_classification_possible_moving_object = 16;
+    optional bool shape_classification_moving_object = 16;
 
     // The defined shape is a landmark.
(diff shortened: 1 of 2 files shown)


osi_featuredata.proto
```diff
@@ -557,6 +563,398 @@ message UltrasonicIndirectDetection
...
+    // The probability for this classification is at least
+    // \c #shape_classification_probability.
+    //
+    optional bool shape_classification_stationary_object = 15;
+
+    // The defined shape is a possible moving object.
+    // The probability for this classification is at least
+    // \c #shape_classification_probability.
+    //
+    optional bool shape_classification_possible_moving_object = 16;
```
CarloVanDriestenBMW: I woukd remove the "possible" here because of the probabilities everything is "possible"?

carsten-kuebler:
I sent Thomas the proposal. It equals the ISO proposal (except of OSI specific terms, objects and so on, and a second documentation).

PR #300, opened 2019-01-04
Use position-independent code build for linux.

jr7:
This basically adds the build option for position-independent code under *nix systems.
The option is required to build shared libraries  on these platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,10 @@ endif()
 set(INSTALL_LIB_DIR lib CACHE PATH "Installation directory for libraries")
 set(INSTALL_INCLUDE_DIR include CACHE PATH "Installation directory for header files")
 
+if(UNIX)
+    set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+endif(UNIX)
+
 if(WIN32 AND NOT CYGWIN)
   set(DEF_INSTALL_CMAKE_DIR CMake/${PROJECT_NAME}-${VERSION_MAJOR})
 else()


pmai:
The CMakeLists.txt file already provides PIC static libraries for building shared objects (shared libraries if you will) where PIC is required through the open_simulation_interface_pic target for exactly this purpose:

Link against open_simulation_interface_static if you are building an executable and want to link against the static lib, link against open_simulation_interface_pic if you are building a shared object and you want to link against the static code, and link against open_simulation_interface if you want to dynamically link in the OSI code.

So I don't think this change is needed. However looking at master, it seems to me that commit ce92c01340262e70674be02a65fbdcc4e983a96b by @nikolavasiljevski just prior to the 3.1.0 release broke the build by making open_simulation_interface a static library, which it is not intended to be.

So I think we need to reverse that change and re-issue 3.1.0 (as 3.1.1 likely), since 3.1.0 is broken.

jr7:
Thank you for the info, somehow overlooked it ... it will check it out and give feedback.
BR Jonas Ruebsam

jr7:
OK seems to work ! I will close the pull request.

pmai:
Thanks, BTW on Linux, due to C++/ProtoBuf brain-damage, you will likely want to link against the shared library (i.e. the open_simulation_interface target, once 3.1.0 is fixed) when using OSI inside a shared library (like an FMU), since there is a great likelihood of the process that your shared library is loaded into also using OSI, and on Linux (contrary to Windows), this will lead to clashes in the protobuf registry.

See the example code in https://github.com/OpenSimulationInterface/osi-sensor-model-packaging/blob/master/examples/OSMPDummySensor/CMakeLists.txt of how to switch between the static and shared libs, which can be automated with conditionalization on Linux.

pmai:
Specifically the

    set(LINK_WITH_SHARED_OSI OFF CACHE BOOL "Link FMU with shared OSI library instead of statically linking")

switch, the value of which you could set to ON conditionalized on Linux.

jr7:
Thank you for the information I will consider this in the future.

jr7:
@pmai  have you considere this https://stackoverflow.com/questions/51795326/static-linking-of-protocol-buffers-results-in-conflicts-with-existing-symbols   ?

pmai:
Yes, changing the namespace in your own copy of OSI is also another known, hackish, work-around, which however involves a bit of complexity and invasive scripting.

I think the proper solution would be for either ProtoBuf to get their act together (seems unlikely), or us ditching ProtoBuf (much more likely, given that ProtoBuf was never a very good encoding for OSI in the first place). I'd like to concentrate our efforts there.

Issue #274, opened 2018-09-26
Canceled/crossed-out traffic signs


ghost:
"Disabled" traffic signs are not yet implemented. This might be the case e.g. when a highway exit ist closed or inaccessible (which is illustrated for instance with red bars over the sign). We have to find a way to model that.

CezarySzyszkaAltran:
Both MainSign and SupplementarySign in osi_trafficsign.proto have field 
classification.variability of type TrafficSign::Variability.

My suggestion for solution of this issue would be to add to enum Variability 
```
 // Temporary or variable sign, e.g. on a sign bridge.
 //
VARIABILITY_DISABLED = 4;
// or VARIABILITY_SIGN_INACTIVE = 4;

```
My concern over this solution is if it is not abusing enum Variability type?

CezarySzyszkaAltran:
I believe we can apply the same principle to the RoadMarking (file osi_roadmarking.proto) 

In case there is a road marking crossed over (example 1. ) we can indicate it with Boolean flag or similar enum type as in osi_trafficsign.proto. I'm not sure about the scenario when road markings are conflicting with each other (example 2.). Should we leave this for interpretation of OSI user to infer which road marking supersedes which? 

Possible solution would be to ass to osi_roadmarking.proto, message RoadMarking.Classification field 
```
// Flag indicating if a road marking is temporarily disabled 
// (e.g. due to construction site proximity )
//
optional bool is_disabled
// alternatively is_valid, is_inuse or  is_active

```

example 1.
![img_0170](https://user-images.githubusercontent.com/44769837/50151747-0f975e00-02c2-11e9-8d9c-e004aadead8d.jpg)

example 2.
![img_0171](https://user-images.githubusercontent.com/44769837/50151756-1c1bb680-02c2-11e9-9731-e6458e46be2d.jpg)

CarloVanDriestenBMW:
@LudwigFriedmannBMW what do you think? if @CezarySzyszkaAltran proposes a PR until Thursday evening we can integrate it in the 3.1.0 release?

LudwigFriedmannBMW:
Hi @CarloVanDriestenBMW , if it's not too late - I'd be fine with the solution proposed by @CezarySzyszkaAltran.

CarloVanDriestenBMW:
@LudwigFriedmannBMW sorry I have to release now - we will do this in the next release!

CezarySzyszkaAltran:
[...]

PR #303, opened 2019-01-11
Added is_out_of_service flag to some road infrastructure messages

CezarySzyszkaAltran:
This Pull Request adds possibility to indicate that certain markings, traffic lights and and traffic signs are not to be taken into account. 

This is achieved by introducing boolean **flag is_out_of_service**.

The naming of the flag can be discussed, I wanted to propose something what is clear and apply to different ways the infrastructure can be crossed-out, covered or painted over, while keeping the same name for TrafficSing, TrafficLight, RoadMarking

If merged this PR closes Issue #274 .


osi_trafficlight.proto
```diff
@@ -70,6 +70,11 @@ message TrafficLight
         //
         repeated Identifier assigned_lane_id = 5;
 
+        // Boolean flag to indicate that the traffic light is taken out of service.
+        // This can be achieved by visibly crossing the light or covering it completely.
```
LudwigFriedmannBMW: I'd propose to add that the traffic light might also be turned off.

CezarySzyszkaAltran: Good point. This is fixed now.

PR #309, opened 2019-03-12
Extension by HMI-, ADAS- and Dynamic-Values (Part II)

ThomasNaderBMW:
So this the continuation of the PR "Extension by HMI-, ADAS- and Dynamic-Values", but now based on a branch. The open points from part I are appended.
The description:

Think about a simulation-construct consisting of a human driver, an AD-function and also a vehicle/dynamic-model.
All these pieces can come from different authors, but should work together.
This is a requirement the driving-simulation-sector has to deal with:

A test person wants to activate an external highly automated driving function in a mock-up. So the function has to listen to the mock-up. The other way around the mock-up has to understand commands from the function (Take-over-request?). So this is an interface that should be standardized.
Furthermore, it is the same with an external dynamic-model. How can this listen to the simulation-environment to get values like the friction of the street needed for dynamic-calculations? And also here, the other way around: The results of the dynamic-model describe the state of a vehicle regarding kinematics. Additionally, calculated values for the engine or the steering can be part of. Both datasets are worth to be standardized.

Dear OSI-Community, please help to enhance the appended .proto-signals.

An architectural indent is appended.
Thank you!

[OSI_Extension_Architectural_Intent_V5.pptx](https://github.com/OpenSimulationInterface/open-simulation-interface/files/3063597/OSI_Extension_Architectural_Intent_V5.pptx)


osi_common.proto
```diff
@@ -428,3 +428,210 @@ message BaseMoving
...
 }
+
+//
+// \brief A description for the steeringwheel.
+//
+message Steeringwheel
+{
+    // Angle of the steeringwheel. 
+    //
+    // Unit in rad: 0=Central (Straight); Left>0; 0>Right.
```
ThomasNaderBMW: @CarloVanDriestenBMW: Right should be clockwise, if you take a look to the "Einheitskreis" or am I wrong here?

osi_common.proto
```diff
@@ -428,3 +428,210 @@ message BaseMoving
...
+}
+
+//
+// \brief A description for the positions of the pedals.
+//
+//
+message Pedalry
+{
+    // Position of the acceleration-pedal.
+    // 0-100 (percentage of position: Unpressed - fully pressed)
```
ThomasNaderBMW: @CarloVanDriestenBMW : Here you commented:
// Position of the acceleration-pedal.
//
// Unit: [ ]
//
// Range: [0.0, 1.0]
//
// \note 0.0 := unpressed; 1.0 := fully pressed

For me this would be ok.

osi_common.proto
```diff
@@ -428,3 +428,210 @@ message BaseMoving
...
+
+        // Change to the left.
+        //
+        INTERPOLATION_METHOD_CUBIC = 1;
+    }
+}
+
+// The actual gear of the car.
+//
+enum Gear
```
ThomasNaderBMW: ppannusch on 11 Dec 2018  Member
What about M/S? Or is this what you mean by GEAR_MID?
Also the enum mixes up actual gear of the powertrain and the gear lever state. What about shift paddles?

osi_driver_inputs.proto
```diff
@@ -0,0 +1,126 @@
...
+
+        // State of the doors. It is often an initial condition to start an ADAS-Function.
+        // 0=Open; 1=Closed
+        //
+        optional bool doors = 2;
+
+        // Hands-On-Detection.
+        // 0=HandsOff; 1=HandsOn
+        //
+        optional bool hands_on_detection = 3;
```
ThomasNaderBMW: ppannusch on 11 Dec 2018  Member
OSI already has SteeringControl for that, which is even more specific (which it should be)

  @ThomasNaderBMW
ThomasNaderBMW on 11 Dec 2018  Author Member
Ou yes, it is part of osi_occupant. Thanks for the hint, this is sth we have to pay attention to.

  @ThomasNaderBMW
ThomasNaderBMW on 13 Dec 2018  Author Member
@CarloVanDriestenBMW & @pmai :
In my opinion the definition of SteeringControl should move from osi_occupant to osi_common, so it can be used in osi_driver_inputs also.
// Definition of hands related to the steering wheel (mostly driver).
//
enum SteeringControl
{
// Hands state is unknown (must not be used in ground truth).
//
STEERING_CONTROL_UNKNOWN = 0;

        // Other (unspecified but known) hand positioning related to the
        // steering wheel.
        //
        STEERING_CONTROL_OTHER = 1;

        // Hands are not on the steering wheel.
        //
        STEERING_CONTROL_NO_HAND = 2;

        // One hand is on the steering wheel. Whether it is the left or right
        // hand is unspecified or unknown.
        //
        // \note If there is no differentiation between one or both hands on the
        // steering wheel, this value should be used.
        //
        STEERING_CONTROL_ONE_HAND = 3;

        // Both hands are on the steering wheel.
        //
        STEERING_CONTROL_BOTH_HANDS = 4;

        // Only left hand is on the steering wheel.
        //
        STEERING_CONTROL_LEFT_HAND = 5;

        // Only right hand is on the steering wheel.
        //
        STEERING_CONTROL_RIGHT_HAND = 6;
    }

osi_common.proto
ThomasNaderBMW: The ISO 7401 describes tests in degrees. SI would be rad.
==> Open Point

osi_common.proto
ThomasNaderBMW: Changed it to 0-1

osi_driver_inputs.proto
```diff
@@ -0,0 +1,126 @@
...
+        //
+        optional LongitudinalControl longitudinal_control = 2;
+
+        // Wished states of the driver regarding the lateral control.
+        //
+        optional LateralControl lateral_control = 3;
+
+        // Wished states of the driver regarding the function Emergency-Brake-Assistant.
+        //
+        optional EmergencyBrakeAssistant emergency_brake_assistant = 4;
```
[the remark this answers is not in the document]
ThomasNaderBMW: ==> Open Point

osi_common.proto
ThomasNaderBMW: ==> Open Point

osi_driver_inputs.proto
ThomasNaderBMW: ==> Open Point

ThomasNaderBMW:
- Today @CarloVanDriestenBMW  and I improved some points of the extension. Now there are 4 points left which need some deeper research.
- In general a check for coding style is need (e.g. Unit: [xxx] instead of In xxx)
- Furthermore there is a need for a research of references

Marius Reis will go on with these points.

osi_vehicle.proto
```diff
@@ -0,0 +1,237 @@
...
+        //
+        optional double engine_consumption = 4;
+
+        // Consumption in liters per 100 km.
+        //
+        optional double fuel_consumption = 4;
+
+        // Consumption of electrical or hybrid vehicle 
+        //
+        optional double electrical_energy_consumption = 5;
```
ainar: What is the unit of this field?

ThomasNaderBMW: kWh/100km

RobertMengerBMW:
I would like to suggest a few more modifications for **osi_vehicle**:

as the current implementation requires the receiving system to simulate the vehicle kinematics again, it would make it easier, if the following informations would be transmitted for every wheel as well:
 - wheel position (3D position relative to car-center)
 - wheel orientation (Euler-rotation relative to car-center)
 - absolute rotation-angle of wheel
 - TireContactPoint with x, y, z so that the walking / torsion / deflation of the tire could be shown correctly
 - Slip (in percent)
 - Slip-angle (in degrees)

with these informations, a connected visualisation could simply render the wheel without any additional efforts to implement joints/hinges and so on. Some of the informations might be used to control a sound generation as well (Slip, TireContactPoint and rotation speed).

mohammadbahram:
@RobertMengerBMW : Could you perhaps upload your ideas directly as a .proto file?

mbencik:
Hy what kind of angle relations to the center are we talking about?

image



or 

image

This is the standar way how the angle car relations a expressed.

RobertMengerBMW:
@mbencik as a matter of fact, in your drawings the angles are relative to the car center. And from a technical point of view, the only availabe reference space is anchored to the center of the car (most likely the middle of the front axle). 
And this is handled that way by every simulation software which I know. Do you have any other examples, where this is handled differently?

mbencik:
image

On the image the Point A has a angle relative to the center, and then the Point A moves to B. On the car that is not the case.

The wheel has a position relative to some point on the car an in the  wheels have a outside relative center for steering.

image

mbencik:
@RobertMengerBMW Well I thing modeling the data so that the detailed status is covered ( camber, toe, rotation  ) and that just the generic information can be shown is the way to go. It would give to all the people that are having the need to develop kinematics features possibilities to use the items.

RobertMengerBMW:
@mbencik 
thats the reason, why I didnt remove the original data fields with camber, toe and so on, but only added my proposal to the end


about your angles:
now I know what youre trying to describe. What you show is the result of the complex mechanical simulation from the kinematics, and in fact, the "real" center of any rotation of the wheels is not in the same place when the wheels move.  But as these are the resulting effects of an upright connected to at least 5 hinges per wheel, you need the correct mechanical model to simulate these angles. But you cant just apply those rotations to a 3D-object, as its pivot point is stationary.

And I was not clear about what I mean with "center of car". I do not mean, that the wheels should rotate around the center of the car, but as the center of the car is in most software the main anchor, its orientation is the also the zero-orientation for all components of the car. You still need the correct translation to the center of the wheelhub (which results in a transform matrix with the center of the car as origin). But as the upright and wheelhub could be oriented in every direction, its is not suitable as a reference point.


Let me explain my background: Im visualizing cars in game engines and professional 3D graphics suites, for real-time driving experience or high end rendering. In the moment I have my own framework to connect a physical simulation like CarMaker to a visualizer like Unity or Cinema 4D. And latter ones are not suited to simulate the mechanics of a complex kinematic system, but CarMaker does and CarMaker (or Adams for example) know the resulting rotation of the wheels and they are relative to the center of the car, as the car is the reference space for every translation and orientation (which is true for game engines and other 3D software as well).

I admit, that Im highly focused on connecting a physics-simulation engine to a visualizer, but Im open for suggestions and other requirements, so that we can find the best possible solution for different usecases.

mbencik:
@RobertMengerBMW  This makes sense now. But what you need is a derived value from the image below.

image

The alpha angle gives the relative vector of the front, and this need to be translated to the vehicle orientation, which is applied to the center of the vehicle. Basically the point Sa in the next image.

image


So the thing would be actually to describe what is the alpha angle representing and how does it impact the model, and the result of the vehicle is the orientation. 

The question is should that derived value be in OSI or not. The reason it is calculated, and the question is are the calculation changing for each vehicle. The problem with such things is the rules that are to be applied to calculate such values could change, so this would have to be defined (the rules), also the constraints what is possible and what not, or what is considered valid for a modeled car and what is invalid.

RobertMengerBMW:
@mbencik 
your drawings are extremly simplified versions just for learning purposes. But it is impossible do describe the real motion of a wheel with your representation. So thats not a sufficient way.

As already written, you have at least five ball-joints / hinges only on the wheel-side, no to mention the influence of rockers, pushrods, antirollbar and so on (which are at least 8-9 hinges on the car-side). So there are exactly two ways to describe the behaviour of the wheel:
1. simulate the mechanis/kinematics on the reciever side, for that you need camber, spring, damper, toe, steering and the complete mechanical model of the kinematics (on that way you can calculate a lot more than just the alpha angle)
2. simply position the wheel based on a ready to use transform-matrix. The transform-matrix could simply be generated from the position and orientation vectors. And a transform-matrix needs a origin, which should be the same for every component, and as all software/devs I know use the center of the car as origin, why shouldnt we do it the same way?

If someone knows a reasonable third way, please tell us.

pmai:
Given that this PR is now fairly huge, and that it is not always clear which information is intended to travel in which direction (e.g. wheel state itself is likely coming from vehicle model, however friction of the wheel is likely sampled from environment, so it seems unlikely that they belong in the same message), might it make sense to split this PR into smaller parts, that are more easily understood and refined?

E.g. the SETLevel4to5 project is currently working on model architectures, and e.g. needs an interface from a traffic participant (ego, other traffic) back to the environment simulation, which would transport the current state of a vehicle. This seems like a subset of the OsiVehicle message (which BTW should not have OSI in the name) might fit the bill.

So my suggestion would be to start splitting up the changes into seperate pull requests, each adding something atomically useful, and clearly defining intended message/information direction of flow.

If that seems palatable, I'd try to start doing this for the SETLevel4to5 kind of things, and we could merge more stuff incrementally from there.

ThomasNaderBMW:
@pmai: Thank you for your answer and excuse my late reply.
I will split up this PR within the next two weeks. 
So that osi_vehicle is a standalone-PR regarding the interface between the Simulation-Environment and the Vehicle-Model. Furthermore it will be used for graphic-rendering (this is a new topic, that came up recently).
For both use-cases there is a proof-of-concept running.
Also Set4to5-Input can become part of it.

TheDraguun:
When compiling osi_driver_inputs.proto from this branch I get:
osi_adas_function.proto: File not found.
osi_driver_inputs.proto: Import "osi_adas_function.proto" was not found or had errors.
osi_driver_inputs.proto:94:18: "HadPilot" is not defined.
osi_driver_inputs.proto:98:18: "LongitudinalControl" is not defined.
osi_driver_inputs.proto:102:18: "LateralControl" is not defined.
osi_driver_inputs.proto:107:18: "EmergencyBrakeAssistant" is not defined.

What am I missing?

jdsika:
> 
> 
> When compiling osi_driver_inputs.proto from this branch I get:
> osi_adas_function.proto: File not found.
> osi_driver_inputs.proto: Import "osi_adas_function.proto" was not found or had errors.
> osi_driver_inputs.proto:94:18: "HadPilot" is not defined.
> osi_driver_inputs.proto:98:18: "LongitudinalControl" is not defined.
> osi_driver_inputs.proto:102:18: "LateralControl" is not defined.
> osi_driver_inputs.proto:107:18: "EmergencyBrakeAssistant" is not defined.
> 
> What am I missing?

@TheDraguun I think it is building correctly now?

ThomasNaderBMW:
At the moment two Proof-of-Concepts with these new formats are running. 
As long as the implementation is not finished, this PR has to pause. It will last about two weeks ;)

ThomasNaderBMW:
Outdated.
Actual Request: osi_vehicle: A deeper description of a vehicle #350

Issue #282, opened 2018-10-18
Unable to import generated.py from a folder


mshaheerm:
I am working with a module.py which is present in `C:\Users\sham\newcarla\carla\PythonClient` and have copied all the generated proto_to_python.py in `C:\Users\sham\newcarla\carla\PythonClient\proto_generated` but there is an import error. i tried running the setup.py in `C:\Users\sham\Downloads\open-simulation-interface-master\open-simulation-interface-master`  and it created a folder `osi3` which just has __init__.py in it. How can i use it as a package for my imports? Thanks in advance

pmai:
I can't tell you anything about carla/PythonClient, but the python files for osi can be generated normally using the python setup-tools approach, i.e. python setup.py build will build everything in build/lib/*, python setup.py sdist will create a software distribution in dist, and/or python setup.py install will do a full local install.

pmai:
I'm not sure what you are trying to do: If you just want to import osi3 stuff, then once it has been installed via python setup.py install, it will already be on your python path, and available normally. If you want to create built distributions that you can distribute/install on other machines, it works the same way for all python packages using setuptools: python setup.py bdist_egg for an egg archive or python setup.py bdist_wheel for the newer whl archive format.

Maybe this is more of a question for carla/PythonClient (whatever that is)?

mshaheerm:
The import problem still persists i am trying to import `osi_grountruth_pb2` in my `module.py` but still there's a problem in importing. even after adding `..\osi3` to the environment variable

pmai:
Which version of Python are you using, how do you import the module, and what error message do you get? (Useful info to provide if something does not work...)

```
import osi3.osi_groundtruth_pb2
```

should work correctly if the package is installed and hence on the python path.

mshaheerm:
@pmai  sorry for the late reply. this is how i am importing from `osi3`

```
import osi3.osi_groundtruth_pb2
import osi3.osi_grpc_pb2_grpc
```
and these are the errors generated when i try to run `osi_grpc_communication_module.py`
```
Traceback (most recent call last):
  File "C:/Users/sham/newcarla/carla/PythonClient/osi_grpc_communication_module.py", line 27, in <module>
    import osi3.osi_groundtruth_pb2
  File "C:\Users\sham\newcarla\carla\PythonClient\osi3\osi_groundtruth_pb2.py", line 18, in <module>
[...]

PR #312, opened 2019-03-14
Correction of the Python installation, the importation were not corre…

ainar:
Actually the correction of the issue #282 : simplification of the installation, work with Python 3 and should work with 2.7


ainar:
This is actually also a fix to a bug. Try to import osi3.osi_groundtruth_pb2 on master after setup.py install to notice it.

PR #317, opened 2019-04-16
Improving coding style along with units

ainar:
Still need to be checked:

Do we let a redundancy in the comment? For example:

```
        // Consumption in liters per 100 km.
        //
        // Unit: [l]
```

I would say "too much is better than not enough", and it is not very annoying or disagreeing with any coding style.

### DIFF
--- a/osi_environment.proto
+++ b/osi_environment.proto
@@ -57,8 +57,10 @@ message EnvironmentalConditions
     optional double wind_speed = 8;
 
     // Contains the direction of the wind (global weather).
-    // In [Degree]. Wind direction is measured in degrees clockwise from due north.
-    // Like on a windrose. E.g.: A wind blowing from the north has a wind direction of 0 Degree.
+    // Wind direction is measured in degrees clockwise from due north.
+    // Like on a windrose. E.g.: A wind blowing from the north has a wind direction of 0°.
+    //
+    // Unit: [deg]
     //
     optional double wind_direction = 9;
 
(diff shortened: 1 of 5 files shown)


osi_environment.proto
```diff
@@ -57,8 +57,10 @@ message EnvironmentalConditions
     optional double wind_speed = 8;
 
     // Contains the direction of the wind (global weather).
-    // In [Degree]. Wind direction is measured in degrees clockwise from due north.
-    // Like on a windrose. E.g.: A wind blowing from the north has a wind direction of 0 Degree.
+    // Wind direction is measured in degrees clockwise from due north.
+    // Like on a windrose. E.g.: A wind blowing from the north has a wind direction of 0°.
```
CarloVanDriestenBMW: ° lets the build fail.

CarloVanDriestenBMW:
> 
> 
> Still need to be checked:
> 
> Do we let a redundancy in the comment? For example:
> 
> ```
>         // Consumption in liters per 100 km.
>         //
>         // Unit: [l]
> ```
> 
> I would say "too much is better than not enough", and it is not very annoying or disagreeing with any coding style.

It is not redundant but it has to be explained why the meaning is "Liter per 100km" and the unit is only "liter" because we are using SI units. Maybe add a reference on why consumption is measured like that ."ISO XY states that consumption in cars is measured in l/100km this my be different in countries like the US....".

Thank you for going through this in detail. This PR has high priority and needs a lot of attention :)

CarloVanDriestenBMW:
@ainar bump

Issue #321, opened 2019-05-03
Date included in OSI Environment


ndunningbmw:
### Current problem:
It is currently not possible to accurately calculate, and thus fill, the ambient light level value in the OSI Environment message. This is because the data of the Ground Truth message is not available has an effect on lighting levels, especially in or morning or evenings.

### Suggestion:
Add a date member of to `osi_environment` which is compliant with ISO 8601. A PR for this will be opened and mentioned here: #323

PR #323, opened 2019-05-06
Added Date to EnvironmentConditions (solves #321)

ndunningbmw:
Issue: #321 

Changes made:
* Created Date message based on UNIX time
* Added Date to EnvironmentConditions and adjusted Field Numbers

### DIFF
--- a/osi_environment.proto
+++ b/osi_environment.proto
@@ -287,7 +287,7 @@ message EnvironmentalConditions
         //
         // \note Although it is possible to extract the \c #time_of_day from this
         // message, it is still recommended that this is taken from the explicit
-        // \c #TimeOfData message.
+        // \c #TimeOfDay message.
         //
         optional int64 seconds_since_epoc = 1;
     }
--- a/osi_environment.proto
+++ b/osi_environment.proto
@@ -21,20 +21,20 @@ message EnvironmentalConditions
     
     // The date at the host vehicles location.
     //
-    optional Date date = 3;
+    optional Date date = 8;
 
     // Atmospheric pressure in Pascal at z=0.0 in world frame (about 101325
     // [Pa]).
     //
     // Unit: [Pa]
     //
-    optional double atmospheric_pressure = 4;
+    optional double atmospheric_pressure = 3;
 
     // Temperature in Kelvin at z=0.0 in world frame.
     //
     // Unit: [K]
     //
-    optional double temperature = 5;
+    optional double temperature = 4;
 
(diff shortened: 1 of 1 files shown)


CarloVanDriestenBMW:
@ndunningbmw can you add the proper refernce to ISO 8601, please? why do you think should we keep the TimeOfDay message? "It is recommended" is kind of missing the "why", or?

pmai:
If this should go into the next minor release of OSI (i.e. OSI 3.2) then the fields cannot be renumbered, since this breaks backward compatibility. Only for OSI 4.0 would renumbering be possible.

The description of the date message seems a bit unclear, since it is not totally clear whether the time expressed is the UTC time, or some local time. Theoretically speaking it should probably use UT1 instead of UTC, but practically speaking UTC should be near enough at all times for these kinds of calculations.

The near duplication with time of day is somewhat unlucky, so maybe time of day should be deprecated, and enough information in environment should be provided so that local time can be derived from the provided UTC (e.g. providing timezone offset).

CarloVanDriestenBMW:
@ndunningbmw better like this?

ndunningbmw:
> If this should go into the next minor release of OSI (i.e. OSI 3.2) then the fields cannot be renumbered, since this breaks backward compatibility. Only for OSI 4.0 would renumbering be possible.

I don't see why we can't wait until OSI 4.0.0, as this is not a urgent change but it is an necessary one.

> The description of the date message seems a bit unclear, since it is not totally clear whether the time expressed is the UTC time, or some local time. Theoretically speaking it should probably use UT1 instead of UTC, but practically speaking UTC should be near enough at all times for these kinds of calculations.

Can you explain why UT1 would be a better alternative?

> The near duplication with time of day is somewhat unlucky, so maybe time of day should be deprecated, and enough information in environment should be provided so that local time can be derived from the provided UTC (e.g. providing timezone offset).

I agree that it is not the best situation, but if the global position of the parent frame cannot be derived from the positional data, being able to still capture the local time is vital to some of the functionallity OSI is enabling.

CarloVanDriestenBMW:
@ndunningbmw like this? I removed the Date message because there is just one global time UTC and there might me multiple TimeOfDay (in theorie). Is the note ok?

CarloVanDriestenBMW:
@pmai I am getting the feeling that the unix_timestamp could also be part of the GroundTruth?

CarloVanDriestenBMW:
@ainar You opinon is also welcome :)

PR #334, opened 2019-07-10
Improvement on Gear handling (HMI-ADAS extension)

ainar:
I tried to clarify its state and to differentiate power train states and lever states.

Any comment would be welcomed.


ThomasNaderBMW:
I took a look at your suggestion and think the gear-systematic is better now. From my point of view it can be merged!

PR #350, opened 2019-10-23
osi_vehicle: A deeper description of a vehicle

ThomasNaderBMW:
This is the continuation of Extension by HMI-, ADAS- and Dynamic-Values (Part II) #309.

#### Add a description
**osi_vehicle is a deeper description of a vehicle.**
The logic is, that it categorizes the vehicle through different messages. The categories are:
Vehicle-Basics, Vehicle-Kinematics, Vehicle-Powertrain, Vehicle-SteeringWheel, Vehicle-Wheels, Vehicle-Light-State, Vehicle-Automated-Driving-Functions.

As it is an description of the whole vehicle it **can be used as interface for various reasons.**
Three proof-of-concepts are still running, they use osi_vehicle as interface to
- the vehicle-model for kinematics calculation (Driver-Inputs are included in the categories)
- the graphic-engine for the visualisation
- a real car for measurement reasons.

**Some questions to ask**:
What is this change?
It is an extension that enables new use-cases for osi and helps to modularize simulation-components.
What does it fix?
Proprietary connections.
As it is an additionaly .proto-file it hasn't any effect on the rest.
How has it been tested?
In three proof-of-concepts (see above).

#### Mention a member
@jdsika @pmai @ThomasSchloemicherAVL 

and @all:
Thanks for the many contributions so far. Go on with that! 👍


ThomasSchloemicherAVL:
Maybe merge with osi_vehicle.proto with osi_hostvehicledata.proto. Personally i like the term HostVehicleData.
Furthermore maybe adding localization info fields plus the method of localization behind. Was the localization method based on pure reading GNSS data or some more advanced localization methods behind such as GNSS-RTK and/or fusing (U)HD-Map info with CV feature extraction etc...

osi_vehicle.proto
```diff
@@ -0,0 +1,692 @@
...
+        // top.
+        // For more information: https://en.wikipedia.org/wiki/Camber_angle.
+        //
+        // Unit: [rad]
+        //
+        optional double camber = 5;
+
+        // Contains the tirepressure of each tire.
+        //
+        // Unit: [Pa]
```
jdsika: I there the need to have a temperature reference here?

osi_vehicle.proto
```diff
@@ -0,0 +1,692 @@
...
+        // 
+        optional double tirepressure = 6;
+
+        // Contains the springdeflection in z-direction for each wheel.
+        //
+        // Unit: [mm]
```
jdsika: no SI

osi_vehicle.proto
```diff
@@ -0,0 +1,692 @@
...
+    
+        //Contains the relativ orientation of the wheel to the center of the car
+        //As the rotation of the wheel is also controlled by this value, 
+        //
+        //Unit: [rad]
+        //
+        optional Orientation3d orientation = 9;
+
+        //Contains the absolute (longitunal) slip of the tire
+        //0-100 percent
```
jdsika: Look at the comment rules here: https://opensimulationinterface.github.io/osi-documentation/osi/commenting.html

osi_vehicle.proto
```diff
@@ -0,0 +1,692 @@
...
+    // to change some vehicle states. E.g. a desired acceleration(-pedal-position), but effective is the one in
+    // the powertrain message.
+    //
+    message VehicleAutomatedDrivingFunction
+    {
+        //
+        // Here the name of the function can be chosen.
+        // E.g.: Hadpilot, longitudinal_control, lateral_control or emergency_brake_assistant.
+        //
+        optional string function_name = 1;
```
jdsika: hm... A string is always the easy solution... "Carlos super duper function" <- valid standard name according to OSI

osi_vehicle.proto
```diff
@@ -0,0 +1,692 @@
...
+            // E.g.: At the point of activation, the actual speed could be 80 km/h, 
+            // but the function tries to accelerate to 130 km/h.
+            //
+            // Unit: [km/h]
+            //
+            optional double targeted_speed = 2;
+
+            // The timegap describes the minimumdistance to the next vehicle in front.
+            //
+            // Unit: [s]
```
jdsika: The minimum distance in seconds? This is also a strange definition

osi_vehicle.proto
```diff
@@ -0,0 +1,692 @@
...
+            }
+        }
+    }
+}
+
+
+//
+// \brief A description for the steering wheel.
+//
+message SteeringWheel
```
jdsika: This definition already exists in osi_ common, right?

osi_vehicle.proto
```diff
@@ -0,0 +1,692 @@
...
+    // Unit: [N*m]
+    //
+    optional double torque = 3;
+}
+
+//
+// \brief A description for the positions of the pedals.
+//
+//
+message Pedalry
```
jdsika: Double defintion

osi_vehicle.proto
```diff
@@ -0,0 +1,692 @@
...
+        //
+        INTERPOLATION_METHOD_CUBIC = 3;
+    }
+}
+
+//
+// \brief The actual gear of the car.
+//
+//
+message GearLeverState
```
jdsika: double definition

osi_vehicle.proto
```diff
@@ -0,0 +1,692 @@
...
+        //
+        optional string reference_string = 4;
+
+        // The weight of the car. 
+        //
+        optional double weight = 5;
+
+        // The license plate of the car. 
+        //
+        optional string license_plate = 6;
```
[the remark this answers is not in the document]
ThomasNaderBMW: Is there a need for a check? The value is used for visualisation-reasons so far..

osi_vehicle.proto
```diff
@@ -0,0 +1,692 @@
...
+    }
+
+    //
+    // \brief So this is the interface, that describes how the vehicle is
+    // moving.
+    // All coordinates and orientations are relative to the global ground truth
+    // coordinate system. 
+    //
+    message VehicleKinematics
```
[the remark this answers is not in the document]
ThomasNaderBMW: Im open to Thomas and your idea for merging. Let's discuss this next time in our meeting.

osi_vehicle.proto
```diff
@@ -0,0 +1,692 @@
...
+
+        // Torque in Nm.
+        //
+        // Unit: [N*m]
+        //
+        optional double engine_torque = 3;
+
+        // Consumption in liters per 100 km.
+        //
+        // Unit: [l]
```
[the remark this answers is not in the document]
ThomasNaderBMW: Sry, dumb mistake. I would take the value that is used in the public, so there is no need to calculate and change the value.. But of course this is a "philosophical" question regarding OSI.

osi_vehicle.proto
ThomasNaderBMW: Really? I couldn't find it: https://github.com/OpenSimulationInterface/open-simulation-interface/blob/master/osi_common.proto
In the PR #309 I had it in the osi_common. , but I changed that.

osi_vehicle.proto
ThomasNaderBMW: see above.

osi_vehicle.proto
ThomasNaderBMW: see above.

osi_vehicle.proto
jdsika: Now I saw that you are doint the PR on master and not on the other branch... You are right the master does not contain the definition!

osi_vehicle.proto
ThomasNaderBMW: E.g. ACC works with a distance in seconds, because meters depends on the velocity you drive

osi_vehicle.proto
ThomasNaderBMW: A enum with the examples as enumeration would be better right?

osi_vehicle.proto
```diff
@@ -0,0 +1,692 @@
...
+        //
+        optional double fuel_consumption = 4;
+
+        // Consumption of electrical or hybrid vehicle per 100 km
+        //
+        // Unit: [kWh]
```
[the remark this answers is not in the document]
ThomasNaderBMW: done.

osi_vehicle.proto
```diff
@@ -0,0 +1,692 @@
...
+    //
+    message VehicleBasics
+    {
+        // The id of the car in the simulation. 
+        //
+        optional Identifier id = 1;
+
+        // The manufacturer of the car. 
+        //
+        optional string manufacturer = 2;
```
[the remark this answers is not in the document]
ThomasNaderBMW: brand is better.
I think about sth. like "Handelsregistereintrag", but that is country-spec I think

osi_vehicle.proto
```diff
@@ -0,0 +1,692 @@
...
+    }
+
+    //
+    // \brief Interface to the vehicle-model.
+    // The focus here is on the wheels.
+    // It is made usage of the wheel-message to shorten the code.
+    //
+    message VehicleWheels
+    {
+        // Contains the physical description of the front-left wheel.
```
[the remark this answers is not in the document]
ThomasNaderBMW: car or vehicle? A question I couldn't answer so far. Let's discuss it next time.

jdsika:
@ThomasNaderBMW [Moore](https://en.wikipedia.org/wiki/Moore_machine)

jdsika:
OSI_Vehicle {
 MovingObject { }
 WhatEverYouNeed { }
}

osi_vehicle.proto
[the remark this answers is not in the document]
ThomasNaderBMW: Ok, there is no international standardization existing. Therefore I would propose:
- If single-line: maximum of 8 Characters
- two-line:        9-13 Characters.
This proposal comes out of DIN 74069:1996-07 and should fit to most license-numbers

osi_vehicle.proto
```diff
@@ -0,0 +1,692 @@
...
+    // The focus here is on the steering wheel.
+    //
+    message VehicleSteeringWheel
+    {
+        // Angle, angle-speed and torque.
+        // See osi_common_extension.
+        //
+        optional SteeringWheel steering_wheel = 1;
+
+        // Spring-stiffness of the steering in Nm/deg.
```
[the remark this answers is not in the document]
ThomasNaderBMW: Ergonomische Untersuchung des
Lenkgefühls an Personenkraftwagen
Dipl.-Ing. (Univ.) Hagen J. Wolf
Pages: 13-15
Here you will see these parameters affecting the steering feel.

osi_vehicle.proto
```diff
@@ -0,0 +1,692 @@
...
+    //
+    // \brief Interface to the vehicle-model.
+    // The focus here is on the physical description of a wheel.
+    //
+    message Wheel
+    {
+        // Contains the friction-coefficient of each wheel.
+        // Dimensionless.
+        //
+        // Unit: []
```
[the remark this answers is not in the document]
ThomasNaderBMW: https://en.wikipedia.org/wiki/Friction#Coefficient_of_friction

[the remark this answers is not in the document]
jdsika: ok, just pick the original reference from the wiki entry. no second level citations like in master thesis etc.

osi_vehicle.proto
ThomasNaderBMW: It can be helpful.

osi_vehicle.proto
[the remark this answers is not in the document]
ThomasNaderBMW: It took me some time, but I think now it is better.

osi_vehicle.proto
```diff
@@ -0,0 +1,692 @@
...
+        //
+        optional double friction_coefficient = 1;
+
+        // Contains the x, y and z-coordinate of the contact point of 
+        // the wheel, so that walking, torsion and deflation can be visualized.
+        // relative to the center of the wheel.
+        //
+        // Unit: [m]
```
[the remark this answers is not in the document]
ThomasNaderBMW: I think it is from ainar.. let's discuss this next time.

osi_vehicle.proto
```diff
@@ -0,0 +1,829 @@
...
+// As it is an description of the whole vehicle (focus on cars) it can be used as interface for various reasons, 
+// e.g. to the vehicle-model for kinematics calculation or even for the visualisation in a graphic-engine.
+//
+message Vehicle
+{
+    //
+    // The interface version used by the sender (i.e. the simulation
+    // environment).
+    //
+    optional InterfaceVersion version = 1;
```
jdsika: I would move this into a root message like TrafficUpdate or similar and use OSI_Vehicle as a class

osi_vehicle.proto
```diff
@@ -0,0 +1,829 @@
...
+    
+    //
+    // The data timestamp of the simulation environment. The zero time point is
+    // arbitrary but must be identical for all messages.
+    // Recommendation: Zero time point for start point of the simulation.
+    //
+    optional Timestamp timestamp = 2;
```
jdsika: move to root message

osi_vehicle.proto
```diff
@@ -0,0 +1,829 @@
...
+    {
+        // The id of the car in the simulation. 
+        //
+        optional Identifier id = 1;
+
+        // The brand of the car. 
+        //
+        // \note It is implementation-specific how the string is filled.
+        //
+        optional string brand = 2;
```
jdsika: brand and model name may better be a part of the reference_string. If the specific model is not available one should use a generic type like the types in MovingObejct

osi_vehicle.proto
```diff
@@ -0,0 +1,829 @@
...
+            // \note It is implementation-specific which information should be sent.
+            //
+            optional Notification notification = 4;
+
+            // Definition of possible function-states.
+            //
+            enum FunctionState
+            {
+                // The function is not working.
+                //
```
jdsika: Unknown=0 , Other=1 is missing

osi_vehicle.proto
```diff
@@ -0,0 +1,829 @@
...
+
+    // Defines the type of the notification.
+    //
+    optional NotificationType notification_type = 2;
+
+    // Definition of possible notification-types.
+    //
+    enum NotificationType
+    {
+        // An error occurred.
```
jdsika: Unknown=0, Other=1

osi_vehicle.proto
ThomasNaderBMW: Done.

osi_vehicle.proto
ThomasNaderBMW: Done.

osi_vehicle.proto
ThomasNaderBMW: I deleted brand and modelname. So far it is not really necessary.

osi_vehicle.proto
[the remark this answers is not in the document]
ThomasNaderBMW: It is in!

osi_vehicle.proto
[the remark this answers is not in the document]
ThomasNaderBMW: Deleted.

osi_vehicle.proto
ThomasNaderBMW: Done.

osi_vehicle.proto
```diff
@@ -0,0 +1,926 @@
...
+option optimize_for = SPEED;
+
+import "osi_version.proto";
+import "osi_common.proto";
+import "osi_object.proto";
+import "osi_occupant.proto";
+
+package osi3;
+
+message Traffic
```
jdsika: The idea was to have a "root message" with version and timestamp and the specific "VehicleClass". Now it has to be clarified what the exact purpose of this root message is. What model uses it as input/output? Consequently a OSMP example should be defined to clarifiy the usage.

osi_vehicle.proto
```diff
@@ -0,0 +1,829 @@
...
+    // Unit: [m/s]
+    //
+    optional double velocity = 7;
+
+    // Contains the acceleration of the vehicle at the timestamp.
+    //
+    // Unit: [m/s^2]
+    //
+    optional double acceleration = 8;
+
```
ThomasSchloemicherAVL: Isnt that redundant information to mention here accleration and velocity again? See VehicleKinematics --> Base Moving

osi_vehicle.proto
```diff
@@ -0,0 +1,829 @@
...
+    // Unit: [1/m]
+    //
+    optional double curvature = 5;
+
+    // Contains the curvature change at the timestamp.
+    //
+    // Unit: [1/(m*s)]
+    //
+    optional double curvature_change = 6;
+
```
ThomasSchloemicherAVL: out of doubles 3, 4, 5, and 6 one can formulate a polyline. us same data type here as for lane boundary polyline (still need to commit these changes, though)

osi_vehicle.proto
```diff
@@ -0,0 +1,829 @@
...
+            optional SteeringWheel steering_wheel = 2;
+
+            // Factor to scale the steeringtorque of the function output.
+            //
+            // \note 0 = no force of the function, 0.5 = half the force, 1 = 100% torque.
+            //
+            // Range: [0, 1] 
+            //
+            optional double steering_override_factor = 3;
+
```
ThomasSchloemicherAVL: Maybe pedal (accelerator and/or brake) override here also?

osi_vehicle.proto
```diff
@@ -0,0 +1,829 @@
...
+        // Contains the camber of each wheel.
+        // 
+        // Negative camber if the bottom of the wheel is farther out than the
+        // top.
+        // For more information: https://en.wikipedia.org/wiki/Camber_angle.
+        //
+        // Unit: [rad]
+        //
+        optional double camber = 5;
+
```
ThomasSchloemicherAVL: Wheel **toe** for completness?

osi_vehicle.proto
```diff
@@ -431,6 +439,73 @@ message VehicleClass
...
+            FUNCTION_NAME_LEVEL_0_LANE_DEPARTURE_WARNING = 12;
+            
+            // Monitoring of the blind spot.
+            //
+            FUNCTION_NAME_LEVEL_0_BLIND_SPOT_MONITORING = 13;
+            
+            // Warning by the danger of a collision.
+            //
+            FUNCTION_NAME_LEVEL_0_FORWARD_COLLISION_WARNING = 14;
+        }
```
ThomasSchloemicherAVL: Well done, thats a great list. Is it accounted for that some of these funtions can occur in parallel?

osi_vehicle.proto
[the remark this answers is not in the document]
TheDraguun: BTW: MovingObject supports vehicles with a variable count of wheels ...

osi_vehicle.proto
TheDraguun: localization has some  overlap with position and position_rmse in osi_hostvehicledata.proto

TheDraguun:
How will the overlap with MovingObject and HostVehicleData be solved? Duplicate and potentially inconsistent places for data might cause maintenance and compatibility nightmares...

pmai:
> 
> 
> Is is wise to change the field ids of an existing message? This will break backward compatibility ...
> Gaps in the list of field ids are not a problem.

Since moving the field to the classification is likely to only happen in a major release anyway, the id change is probably immaterial. That said, I think the lane_id stuff should be aligned with ISO 26150 DIS once that is available/forseeable at the same time as the changeover to the classification.

osi_vehicle.proto
```diff
@@ -0,0 +1,971 @@
...
+    repeated VehicleAutomatedDrivingFunction vehicle_automated_driving_function = 8;
+
+    //
+    // \brief The absolut base parameters of the vehicle.
+    //
+    message VehicleBasics
+    {
+        // The id of the car in the simulation. 
+        //
+        optional Identifier id = 1;
```
ManuelKugelmann: Identifier id is maybe better placed into the root  of the VehicleClass message to mirror MovingObject etc..
That it is nested forced me to add some extra code in my framework ...

BTW: what is the reasoning behind the name "VehicleClass" ?

osi_vehicle.proto
ThomasNaderBMW: at least it is the same like the steering-angle right?

osi_vehicle.proto
ThomasNaderBMW: Values I would say on the "waiting-list". Is there a use-case?

osi_vehicle.proto
ThomasNaderBMW: maybe it makes sense to use your polyline-definition? Let's discuss this next time.

osi_vehicle.proto
ThomasNaderBMW: meant is the targeted velocity and acceleration at the planned point.. but I am not sure here

ThomasNaderBMW:
Topic is integrated in updating hostVehicleData, see here:
https://github.com/OpenSimulationInterface/open-simulation-interface/pull/441

PR #357, opened 2019-11-20
Lfs storage and OSI file format definition

vkresch:
There is an [issue](https://github.com/OpenSimulationInterface/osi-visualizer/issues/30) on how to define files which can be read by the visualizer. Also there is the current issue on how to save large files (osi traces, images, ...) in the OSI repo. 

#### Add a description
This PR defines the OSI format for storing multiple messages in one file based on the length of the message. It also provides a script to convert from currently used separated trace files by `$$__$$` to to the OSI format. This is made so we avoid confusion on how to correctly define OSI file formats and to use [Google official suggestion](https://developers.google.com/protocol-buffers/docs/techniques?csw=1#streaming) on how storage of multiple messages should be handled. Furthermore this PR adds large file storage functionality to files defined in the `.gitattributes` file.

For tests the osi-validator can be used that can now validate both type of formats but warns the user that the old format will be deprecated in the near future.

PS: The huge file deletions for the images appear because they are being moved the the lfs cloud database.

#### Mention a member
@jdsika let me know you thoughts, suggestions and improvements ;)

#### Check the checklist


pmai:
This commit seems to move the StVO image files to lfs storage; I'm not sure whether that makes much sense, given that the files are individually fairly small and unlikely to change and as such not a good match for LFS I would assume. The use of LFS for large binary files like traces is indicated, but for the smallish image files I'm not sure that this will improve things.

I'd also split the file format definition and LFS switch into seperate PRs ideally...

vkresch:
@pmai good point then I will for now pause the git lfs implementation and use it just for trace file in the osi-visualizer and osi-validator repo.
@jdsika you can close this PR. I will create a new one.

PR #358, opened 2019-11-26
Added osi file format converter and read converter

vkresch:
There is an [issue](https://github.com/OpenSimulationInterface/osi-visualizer/issues/30) on how to define files which can be read by the visualizer.

#### Add a description
This PR defines the OSI format for storing multiple messages in one file based on the length of the message. It also provides a script to convert from currently used separated trace files by `$$__$$` to to the OSI format. This is made so we avoid confusion on how to correctly define OSI file formats and to use [Google official suggestion](https://developers.google.com/protocol-buffers/docs/techniques?csw=1#streaming) on how storage of multiple messages should be handled. It is optional to convert into lzma compressed files since this compression algorithm does a good job on compressing ~1.3gb *.txt into 2,4mb *.txt.lzma (540x smaller, compression time 2min, decompression time 0.5s).

I also provide a osi2read.py script which converts serialized trace files to human readable files for plausibliity checks of the trace files. It should be avoided for now to use it on to huge trace files since the readable files could not be opened anyway :).

For tests the osi-validator can be used that can now validate both type of formats but warns the user that the old format will be deprecated in the near future.

#### Mention a member
@jdsika @pmai let me know your thoughts, suggestions and improvements ;)

#### Check the checklist

### DIFF
--- a/doc/fileformat.rst
+++ b/doc/fileformat.rst
@@ -59,7 +59,6 @@ If you want to generate a valid OSI trace file which can be used as an input for
             moving_object.base.orientation.pitch = 0.0
             moving_object.base.orientation.yaw = 0.0 
             
-
             """Serialize"""
             bytes_buffer = sensorview.SerializeToString()
             f.write(struct.pack("<L", len(bytes_buffer)) + bytes_buffer)
--- a/doc/fileformat.rst
+++ b/doc/fileformat.rst
@@ -32,34 +32,37 @@ If you want to generate a valid OSI trace file which can be used as an input for
         sv_ground_truth.version.version_minor = 0
         sv_ground_truth.version.version_patch = 0
 
-        sv_ground_truth.timestamp.seconds = 4
-        sv_ground_truth.timestamp.nanos = 54999999
+        sv_ground_truth.timestamp.seconds = 0
+        sv_ground_truth.timestamp.nanos = 0
 
-        stationary_object = sv_ground_truth.stationary_object.add()
-        stationary_object.id.value = 114
+        moving_object = sv_ground_truth.moving_object.add()
+        moving_object.id.value = 114
 
-        for i in range(11):
+        # Generate 10 OSI messages for 9 seconds
+        for i in range(10):
+
+            # Increment the time
+            sv_ground_truth.timestamp.seconds += 1
+            sv_ground_truth.timestamp.nanos += 100000
+
+            moving_object.vehicle_classification.type = 2
             
-            stationary_object.base.dimension.length = 3
-            stationary_object.base.dimension.width = 0.5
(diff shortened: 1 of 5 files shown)


doc/fileformat.rst
```diff
@@ -0,0 +1,74 @@
...
+
+To read to content of a serialized txt/osi trace file we also provide a converter ``osi2read.py``.
+See the usage below:
+
+.. code-block:: bash
+
+    python3 osi2read.py -d trace.osi -o readable_trace
+    python3 osi2read.py -d trace.txt -f separated -o readable_trace
+
+which outputs a ``readable_trace.txth`` which can be opened by any text editor.
```
jdsika: txth?

Please add a "warning" for large files as in the PR description

doc/fileformat.rst
```diff
@@ -0,0 +1,74 @@
...
+        sv_ground_truth.version.version_minor = 0
+        sv_ground_truth.version.version_patch = 0
+
+        sv_ground_truth.timestamp.seconds = 4
+        sv_ground_truth.timestamp.nanos = 54999999
+
+        stationary_object = sv_ground_truth.stationary_object.add()
+        stationary_object.id.value = 114
+
+        for i in range(11):
```
jdsika: I do not understand the loop here ... Don't you have to increment the time as well? And the <L does not stand for a Long type?

doc/fileformat.rst
vkresch: txth just stands for "(**t**)e(**x**)(**t**) (**h**)uman readable" for distinguishing between serialized txt and readable txth. Will add the warning good point.

doc/fileformat.rst
vkresch: Yes the time increment should be also in the loop, will add it. The '<L' stands for little endian long type which in python is defined as a 32bit int (see pmai's[ comment](https://github.com/OpenSimulationInterface/osi-visualizer/issues/30#issuecomment-557580286)). Since OSMP are restricted to 32bit I thought we would use than the 32bit int.

doc/fileformat.rst
```diff
@@ -0,0 +1,76 @@
+OSI File Format
+----------------
+
+To save multiple serialized OSI messages into one trace file we use the length of each OSI message and save it before the actual OSI message.
```
jdsika: There should be a list of the possible file formats:
1) *.osi explanation
2) *.txt explanation, deprecated
3) *txth ...

doc/fileformat.rst
jdsika: He says "LE"?

vkresch: LE: **L**ittle **E**ndian int 32bit

PR #366, opened 2020-01-14
Extension osi lane

ThomasSchloemicherAVL:

#### Add a description
A polyline model was added to the osi_lane.proto to describe the lane boundaries with a polyline up to 3rd order.

**Some questions to ask**:
What is this change?
What does it fix?
It is a feature.

How has it been tested?

#### Mention a member
@PhRosenberger @ThomasNaderBMW @jdsika 

#### Check the checklist


clemenshabedank:
Hello @ThomasSchloemicherAVL let's discuss at your next meeting if this is postponed for major release or introduced in V3.3. I think there is nothing in it that would break backwards compatibility (besides that you can't just switch the enumeration values). But we should discuss if it might be a useful and reasonable field within SensorView/GroundTruth or if we only want to introduce it in osi_detectedlane.proto. The way you implemented it doesn't really make clear which or if both models (boundary points + BoundaryPolyLineModel) should be used.

My opinion would be: Definitely useful for detected lane boundaries, especially with the harmonization with ISO 23150 aspect. For lane boundaries in GroundTruth I am sceptical because it is actually already processed data. But on the other hand it could make calculations easier for Traffic Participants... But I would then tend to take this model as a center line extension because the centerline is already "imaginative data" and is used for guiding traffic participants.

PR #368, opened 2020-01-15
Move assigned_lane_id in osi_object.proto to VehicleClassification

ThomasSchloemicherAVL:

#### Add a description
Add the assigned lane id to the vehicle classification as it is the case with other entities (e.g. traffic signs)

**Some questions to ask**:
What is this change?
What does it fix?
Add the assigned lane id to the vehicle classification as it is the case with other entities (e.g. traffic signs)
It is a bug fix. Fields got altered, therefore minor version upgrade.

How has it been tested?
Documentation was build

Vehicle Classification without assigned lane id
image

Vehicle Classification with assigned lane id
image

#### Mention a member
@jdsika @ThomasNaderBMW 

#### Check the checklist

### DIFF
--- a/osi_detectedobject.proto
+++ b/osi_detectedobject.proto
@@ -190,7 +190,7 @@ message DetectedMovingObject
     //
     // Percentage value of the object width in the corresponding lane.
     //
-    // \note DEPRECATED: Use assigned_lane_percentage in VehicleClassification
+    // \note DEPRECATED: Use assigned_lane_percentage in MovingObjectClassification
     // instead.
     //
     // \rules
@@ -204,7 +204,7 @@ message DetectedMovingObject
     //
     // Percentage value of the object width in the corresponding lane.
     //
-    // \note DEPRECATED: Use assigned_lane_percentage in VehicleClassification
+    // \note DEPRECATED: Use assigned_lane_percentage in MovingObjectClassification
     // instead.
     //
     // \rules
--- a/osi_detectedobject.proto
+++ b/osi_detectedobject.proto
@@ -315,6 +315,10 @@ message DetectedMovingObject
         // [1] Patton, K. T. & Thibodeau, G. A. (2015). <em>Anatomy & Physiology</em>. 9th Edition. Elsevier. Missouri, U.S.A. ISBN 978-0-323-34139-4. p. 1229.
         //
         optional Orientation3d upper_body_pose = 5;
+
+        // Specific information about the classification of a moving object.
+        //
+        optional MovingObject.MovingObjectClassification moving_object_classification = 6;        
     }
 
     // Definition of available reference points. Left/middle/right and
--- a/osi_detectedobject.proto
+++ b/osi_detectedobject.proto
@@ -190,6 +190,9 @@ message DetectedMovingObject
     //
     // Percentage value of the object width in the corresponding lane.
     //
+    // \note DEPRECATED: Use assigned_lane_percentage in VehicleClassification
+    // instead.
+    //
     // \rules
     // is_greater_than_or_equal_to: 0
     // is_less_than_or_equal_to: 100
@@ -200,6 +203,9 @@ message DetectedMovingObject
     // Percentage side lane right.
     //
     // Percentage value of the object width in the corresponding lane.
+    //
+    // \note DEPRECATED: Use assigned_lane_percentage in VehicleClassification
+    // instead.
     //
     // \rules
     // is_greater_than_or_equal_to: 0
--- a/osi_object.proto
+++ b/osi_object.proto
@@ -572,6 +572,15 @@ message MovingObject
         //
         optional Identifier trailer_id = 4;
 
+        // The IDs of the lanes that this object is assigned to.
+        //
+        // \note Might be multiple if the object is switching lanes or moving from
+        // one lane into another following lane.
+        //
+        // \note OSI uses singular instead of plural for repeated field names.
+        //
+        repeated Identifier assigned_lane_id = 5;
+
         // Definition of vehicle types.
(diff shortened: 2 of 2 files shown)


ThomasSchloemicherAVL:
@jdsika @pmai Would it be an option to only copy the field for now to the vehicle classification message in order to be downwards compatible and delete the assigned_lane_id then later?

pmai:
That would definitely be a better option for now, and we could deprecate the original field (or even delete without renumbering). That said, I'd also think that this should merge the percentage fields in DetectedMovingObject into this new field as well, e.g. something like:

```
repeated Identifier assigned_lane_id = 5;
repeated double assinged_lane_percentage = 6;
```

ThomasSchloemicherAVL:
@pmai ok, have reverted the last commit.
What is the assigned lane percentage? Where can i find that? Or it doesnt exist yet?
`repeated double assinged_lane_percentage = 6;`

pmai:
These are the changes I had in mind.

However looking at the overall situation it seems to me that assigned lanes also should apply to non-vehicle moving objects potentially (e.g. pedestrians, animals, ...); which might have been the reason why the fields were not initially put into the VehicleClassification.

So a cleaner solution might be to introduce a MovingObjectClassification, and move those fields there...

Any opinions on this?

ThomasSchloemicherAVL:
@pmai. I agree but I would suggest to make this bugfix now and do a complete rework later (for osi v4 maybe).

What do you think @pmai, @kmeids ?

clemenshabedank:
Discussed in WP "OSI as a measurement file" from @ThomasSchloemicherAVL (2020-12-02) that it is Ready for CCB review.

clemenshabedank:
I have thought about this topic again. Isn't the information semantically different depending on where it is located. So currently it is inside DetectedMovingObject which means that it is the **output** of the sensor model (so how the sensors see the world, which might be different to Ground Truth). VehicleClassification is part of GroundTruth which is provided by the environment simulator. So we could have both!? @ThomasSchloemicherAVL can you check if it is part of ISO-23150? If yes, I would not depreciate it in DetectedMovingObject in the major release. I think it makes sense to also include it in VehicleClassification.

clemenshabedank:
> These are the changes I had in mind.
> 
> However looking at the overall situation it seems to me that assigned lanes also should apply to non-vehicle moving objects potentially (e.g. pedestrians, animals, ...); which might have been the reason why the fields were not initially put into the VehicleClassification.
> 
> So a cleaner solution might be to introduce a MovingObjectClassification, and move those fields there...
> 
> Any opinions on this?

Or we introduce PedestrianClassification and AnimalClassification, in case the MovingObject is one of those? This would be backward compatible and fit the current logic.

kmeids:

Actions:
1. Further technical discussion should be done on this topic, by @ThomasSchloemicherAVL @clemenshabedank and @pmai.

clemenshabedank:
> I have thought about this topic again. Isn't the information semantically different depending on where it is located. So currently it is inside DetectedMovingObject which means that it is the **output** of the sensor model (so how the sensors see the world, which might be different to Ground Truth). VehicleClassification is part of GroundTruth which is provided by the environment simulator. So we could have both!? @ThomasSchloemicherAVL can you check if it is part of ISO-23150? If yes, I would not depreciate it in DetectedMovingObject in the major release. I think it makes sense to also include it in VehicleClassification.

This comment of mine is wrong after our discussion and looking into the class diagram again. VehicleClassification is part of SensorView and (in another intepretation) in SensorData, which makes sense for the two fields. So I think it's a good idea to move it. So to recap already your suggestions: 

1. Keep it like it is (Thomas suggestion)
2. Introduce MovingObjectClassification and put it there (Pierres suggestion) 

@ThomasSchloemicherAVL and @pmai what do you think is better? 
I think introducing MovingObjectClassification would make it more reusable, and we wouldn't have to double the fields if we introduce e.g. PedestrianClassification in the future, where the fields would also make sense... Should I make the according changes?

ThomasSchloemicherAVL:
@clemenshabedank I think we should go for @pmai 's suggestion.

clemenshabedank:
@stefancyliax @pmai: We discussed in @ThomasSchloemicherAVL 's WP that this PR would be readyforCCBReview. We were wondering if just one of us is allowed to do the signoffs for all the commits where it is missing (contributions from Thomas, Pierre and me). Else, could you please help how it would work for each one to do it separatedly for the single commits?

ThomasSchloemicherAVL:
@clemenshabedank mustn't the MovingObjectClassification also be included in DetectedMovingObject?

pmai:
> @clemenshabedank mustn't the MovingObjectClassification also be included in DetectedMovingObject?

More specifically into DetectedMovingObject.CandidateMovingObject, yes...

clemenshabedank:
@ThomasSchloemicherAVL and @pmai thanks you are right. I changed it in https://github.com/OpenSimulationInterface/open-simulation-interface/pull/368/commits/3f6db80258777fcd3455278a2a21ac293daf8286 

I also added the field in MovingObject because the information of the assigend lane id is currently already in the GroundTruth. When it will be depreciated, we have it then under the new submessage. --> https://github.com/OpenSimulationInterface/open-simulation-interface/pull/368/commits/b412f896a2d1fc5f87b5b4567565b1617e6ed2c7

Does that make sense for you?

kmeids (approved):
Approved during the CCB meeting 11.02.2021

kmeids:
**Output of CCB meeting 11.02.2021**
Actions:
1.@pmai to do the merge (all agreed for @pmai to do the sign.off )

pmai (approved):

PR #401, opened 2020-07-21
WIP: Proposed TrafficUpdate changes

caspar-ai:
These are proposed changes to the traffic update, and some small changes to the ground truth.  They build on top of the existing work in the combined traffic participants branch.

The original requirements are here: 

https://code.asam.net/simulation/proposal/osi/-/blob/master/subchapters/1-4_requirements.adoc#user-content-provide-interface-for-traffic-participants

 These changes satisfy all requirements from (1), (2), (3), and (4).

 It doesn't address 5-7 as they are largely orthogonal.


pmai:
First, thank you for pushing this to move forward on the discussions.

As a meta-comment, I think the wheel data additions could and should be moved into a separate PR sometime in the future, since those changes are largely unrelated/orthogonal to the other changes, I think? As a side-note I would think the wheel data should probably contain all information on the whole wheel carrier (the whole unsprung mass), which would usually lead to a full 6DOF representation of the rotation axis, plus dimensions, plus rotational stuff, etc.) This would also clarify the description I think, since it currently would be unclear how steering angle is supposed to be defined, for example.

As to the future trajectory data, should this not instead also be added inside MovingObject (like the WheelData), rather than beside it? With the current approach, we would have no way to propagate the data to any other recipients, which I would assume is the object of providing it? I.e. other traffic participants might want to react to that data in some form, etc.

caspar-ai:
Thanks for the comments!

On the wheel data:
* Yes, I agree, the wheel data is orthogonal and when we're looking to actually merge it should go in a separate MR.  I put them together as this was everything I had and it made review a bit easier (as well as keeping all the stuff for those requirements together).
* Yes, a full 6DOF could be useful.  I do agree that it was quite hard to describe what I meant by the `steer_angle` in the comments which is perhaps a sign :)  Would potentially make the interface more complicated, but if the information is likely to be available in the simulator perhaps little difference.

With regards the trajectory, sounds like there are two different approaches:
1. Each traffic participant broadcasts its own update, and other traffic participants (and the simulator) can subscribe to those as needed
2. Each traffic participant sends its update to the simulator, the simulator aggregates them and includes them in the ground truth message which all traffic participants then receive

I think I had model (1) in my head, and I think from your comments you are thinking of model (2).  Does that sound right?

(no strong preference, think both work, but worth deciding on the expected architecture - or to support both)

clemenshabedank:
Thanks @caspar-ai for making a start in this respect! 

I think the fact that suggests an integration of future trajectories into OSI::MovingObject is that this is already nested in OSI::GroundTruth, so it would be sent via OSI::SensorView to other TrafficParticipants after TrafficUpdate has updated the World state in the simulator (-> OSI::SensorView is what TrafficParticipants currently consume). This would make model (2) a favorite. Probably that is what Pierre already meant.

In addition to Pierre's comments on the Wheel data I would suggest to include units for every (sub)message, and maybe a **counter**-clockwise direction of orientation (because I think that is convention in OSI, isn't it?).

ThomasNaderBMW:
Some time has passed already, nevertheless I would like to catch up this valuable PR:
- WheelData for MovingObject: I like the axle/index concept. I agree on 6 DOF would be better than steering angle. Maybe even the z-contactpoints could be argued into it, what can help for later dynamic-calculations (as it can be transported via osi-gt). Then just the friction is missing..
- FutureTrajectory: Shouldn't the future_point be repeated? And in total I would like to see it in WP15 (https://github.com/OpenSimulationInterface/open-simulation-interface/pull/405) as this is an internal value that is not perceiveable from the outside. To be included in osi-gt we could think about adding the HostVehicleDataClass -Submessage to osi-gt or better optional to moving-object. I invited you to our next meeting :)

Splitting up this branch is a good way to start!

caspar-ai:
@ThomasNaderBMW thanks for the comments.

What you are suggesting with regards the `WheelData` sounds sensible.  My PR should definitely be viewed as a starting point which captures everything are currently using and require within our setup.  Anything which adds more detail or precision sounds fine.  (provided it doesn't become too complicated to implement!)

With regards the future trajectory, the `future_point` is within the repeated `future_trajectory` as every future (positional point) also needs to be captured alongside the time that the point will be reached.  Does that make sense?

caspar-ai:
I have now updated the PR as per the various comments:

- @pmai and @ThomasNaderBMW the wheel data is now removed as that is covered by https://github.com/OpenSimulationInterface/open-simulation-interface/pull/430
- @pmai I have moved the future trajectory inside the moving object as you suggested - there has been some discussion about whether `MoveObject` is "allowed" to hold non-perceivable data, but if people are happy with that I think the solution is cleaner
- @clemenshabedank I've pulled `StatePoint` out into a common file for re-use (and then used it!)

Let me know if there is anything else I've missed.

clemenshabedank:
Thanks @caspar-ai, I like that you are explaning that the future trajectory should not be used for the stack under tests but for other traffic participants which are not under test! (just a minor comment: maybe you could change "agent" to "traffic participant", because we want to stay consistent with the naming. 
As discussed, also pulling out Statepoint is fine to me, thanks for already committing in osi_trafficcommand.proto. 
Regarding osi_trafficupdate.proto: I think the following comment should be rewritten, because we also want to include your newly introduced `repeated StatePoint future_trajectory`, right? :
  ```
  // \note Only the id, base member (without dimension and base_polygon),
    // and the vehicle_classification.light_state members are considered in
    // updates, all other members can be left undefined, and will be
    // ignored by the receiver of this message.
```

caspar-ai:
@clemenshabedank thanks for the comments, have fixed up both points.
https://github.com/OpenSimulationInterface/open-simulation-interface/commit/d45a5bfbde4d0785ccbad6cf32c36d5a80983d95

kmeids:
CCB Meeting 25.11.2020

Actions:
@paagkame to perform documentation check on "File changed".
@clemenshabedank make sure that https://github.com/OpenSimulationInterface/open-simulation-interface/pull/409 is merged before this one (https://github.com/OpenSimulationInterface/open-simulation-interface/pull/401).

paagkame:
documentation review is done.

kmeids:
Actions:
1. Ready to be merged, it should be done after https://github.com/OpenSimulationInterface/open-simulation-interface/pull/409, by @pmai

caspar-ai:
Thanks for the update @kmeids.

@pmai as a heads-up, this PR is currently targeting a non-master branch, so will need rebasing (or equivalent) before merge.  I'm happy to do that once #409 has merged, otherwise let me know if you're going to handle.

pmai:
> 
> 
> Thanks for the update @kmeids.
> 
> @pmai as a heads-up, this PR is currently targeting a non-master branch, so will need rebasing (or equivalent) before merge. I'm happy to do that once #409 has merged, otherwise let me know if you're going to handle.

Will handle the rebase/conflict stuff right now.

pmai:
I have rebased and cleaned up the branch (resolving conflicts etc.) for merging into master, which technically should now work.

One question that has popped up for me is the renaming of the repeated MovingObject in TrafficUpdate from update to traffic: While this change is unproblematic from the PoV of ASAM OSI (since TrafficUpdate was never released in the old form), the for SETLevel internal usage this would make the current update non-backward compatible, which would not be nice since it seems to be the only non-backward compatible change we are expecting for 3.3.0. Now that is not a pressing consideration for ASAM OSI, but if there is no strong feeling one way or the other, I'd definitely prefer the update naming just to make our lives easier.

So my question would be whether the name change is considered important by WP10/the rest of us, or whether update would also be good enough?

caspar-ai:
@pmai thanks for sorting out the PR.  I have no strong feelings.  I do think `traffic` is a better - in particular more descriptive - name than `update`, but I don't feel strongly enough to cause back-compatibility issues so I'm OK with you changing it back to `update`.

caspar-ai:
@pmai is there anything else that needs doing before we can merge this to master?  Anything I can help with?

caspar-ai:
@pmai I've updated the field name as you requested to be back-compatible with SETLevel project implementation.

I'm not authorised to merge this PR, are you able to do that?

pmai (approved):
Ready to be merged as per the CCB decisions, just approving for organisatorial reasons.

Issue #400, opened 2020-07-20
Missing Vehicle-Signals


ThomasNaderBMW:
Hello everyone!

In this issue I would like to do a collection of signals regarding the vehicle, which are missing so far in OSI.
The issue comes out of the ASAM-Project and is mainly driven bei WP 14/15, but of course everybody can contribute!
Which interface is being touched should be left open and something we want to think about - based on this collection.
The focus should be on internal/external vehicle states and even the ad-function.
Road information until a specific depth (so please no trafficsign optimizations).

### Example:
_Name_: pedal_position_acceleration
_Description_: Position of the acceleration-pedal. 
_Unit_: [0-1] (Unpressed - fully pressed)
_Type_: Double
_Sender_: From Host-Vehicle (Mock-Up)
_Receiver_: Dynamic-/Vehicle-Modell

Thanks in advance for contribution! :)

Krishna626:
Hello Everyone, I would like to have the following interfaces defined to simulate external dynamics in addition to the interface defined above:

Name: pedal_position_brake
> Description: Position of the brake-pedal.
> Unit: [0-1] (Unpressed - fully pressed)
> Type: Double
> Sender: From Host-Vehicle (Mock-Up)
> Receiver: Dynamic-/Vehicle-Model

Name: angle_steering_wheel
> Description: angular position of steering-wheel.
> Unit: [-pi-pi] (rotation)
> Type: Double
> Sender: From Host-Vehicle (Mock-Up)
> Receiver: Dynamic-/Vehicle-Model

Name: rpm_engine/E-motor
> Description: current engine/e-motor rpm
> Unit: [0-8000] (rotations per minute)
> Type: uint16
> Sender: From Host-Vehicle (Mock-Up)
> Receiver: Dynamic-/Vehicle-Model

Name: position_gear_lever
> Description: current gear lever position
> Unit: [-1to 8] (R: -1, D:5, N:4 etc.)
> Type: int8
> Sender: From Host-Vehicle (Mock-Up)
> Receiver: Dynamic-/Vehicle-Model

Name: z_position_wheel_contactpoints
> Description: z-value of wheel contact points from road surface
> Unit: [0-1000+] (meters)
> Type: uint8
> Sender: From Host-Vehicle (Mock-Up)
> Receiver: Dynamic-/Vehicle-Model

Name: friction_wheel_contactpoint
> Description: friction values of wheel contact points at road surface
> Unit: -
> Type: uint8
[...]

PR #405, opened 2020-08-17
Update osi_hostvehicledata.proto

Krishna626:
- suggestion for protobuf description
- replaced engine with motor to avoid explicitly defining engine for interfaces
- reorganised wheel arragement--> anti clockwise starting from front left wheel
- Naming suggestion for VehicleLocalization--> renamed as Vehicle GEOPosition

https://github.com/OpenSimulationInterface/open-simulation-interface/issues/400

#### Add a description
Minor modifications to the proposal

#### Mention a member
@pmai @ThomasNaderBMW @clemenshabedank @ThomasSchloemicherAVL : For further discussion

### DIFF
--- a/osi_hostvehicledata.proto
+++ b/osi_hostvehicledata.proto
@@ -14,8 +14,7 @@ package osi3;
 
 // \image html OSI_HostVehicle.svg
 //
-// All coordinates and orientations are relative to the global ground truth
-// coordinate system.
+// All coordinates and orientations are relative to the global ground truth coordinate system.
 //
 message HostVehicleData
 {
@@ -73,7 +72,7 @@ message HostVehicleDataClass
 
     // Interface regarding the navigation.
     //
-    optional VehicleGEOPosition vehicle_geo_position = 6;
+    optional VehicleLocalization vehicle_localization = 6;
 
     //
     // \brief The absolute base parameters of the vehicle.
@@ -82,10 +81,10 @@ message HostVehicleDataClass
     {
         // The total mass of the vehicle (curb weight). 
         //
-        // Unit: [kg]
+        // Unit: kg
         //
         // \par Reference:
-        // Paragraph 42 of the StVZO (ger. Straßenverkehrs-Zulassungs-Ordnung; engl. Road Traffic Regulations)
+        // Paragraph 42 of the German Road Traffic Admission Regulations (StVZO).
         //
         optional double curb_weight = 1;
     }
@@ -109,13 +108,13 @@ message HostVehicleDataClass
 
         // Rounds per minute of the engine. RPM can be from E-Motor/ Engine.
         //
-        // Unit: [1/min]
+        // Unit: 1/min
         //
         optional double motor_rpm = 2;
 
         // Torque in Nm. It can either be from Engine/E-Motor or combined Torque values.
         //
-        // Unit: [N*m]
+        // Unit: N*m
         //
         optional double motor_torque = 3;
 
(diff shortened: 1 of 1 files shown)


osi_hostvehicledata.proto
```diff
@@ -72,7 +72,7 @@ message HostVehicleDataClass
 
     // Interface regarding the navigation.
     //
-    optional VehicleLocalization vehicle_localization = 6;
+    optional VehicleGEOPosition Vehicle_geo_Position = 6;
```
ThomasNaderBMW, line 75: Minor: the vehicle has to be written in a small letter, so vehicle_geo_Position

osi_hostvehicledata.proto
```diff
@@ -106,17 +106,17 @@ message HostVehicleDataClass
         //
         optional Pedalry pedalry = 1;
 
-        // Rounds per minute of the engine.
+        // Rounds per minute of the engine. RPM can be from E-Motor/ Engine.
```
ThomasNaderBMW, line 109: Better now! Thx :)

osi_hostvehicledata.proto
```diff
@@ -319,7 +320,7 @@ message HostVehicleDataClass
     //
     // \brief This message contains all the necessary information of the localization solution.
     //
-    message VehicleLocalization
+    message VehicleGEOPosition
```
ThomasNaderBMW, line 323: Could it be, that xyz will be also part of this message, so georeferenced is not the only option in the future? Look to depricated optional BaseMoving location_rmse = 2;
So localization would be more open..

osi_hostvehicledata.proto
raue: since you brought the E-Motor stuff up; how to deal w/ vehicles, which have more than one of those (e.g. Tesla S 4x4)?

ThomasNaderBMW: Looks to me like we will have to do a repeated submessage "Engine". Good point, i will pay attention to that at the next steps ::)

Krishna626: @ThomasNaderBMW in yesterday's TrafficParticipants discussion @pmai has also mentioned that we may need to consider how to set this signal in cases where more than one e-motor is available?
Idea for further discussion: It can be a consolidated signal going to wheels from all sources. Naming of the signal to be brainstormed in the follow-up meetings.

osi_hostvehicledata.proto
Krishna626: @ThomasNaderBMW : as discussed, I would keep xyz and geo-position messages separately for flexibility purposes.

osi_hostvehicledata.proto
ThomasNaderBMW: Topic for next WP15-Meeting.

osi_hostvehicledata.proto
adac-nakanishi: I would like xyz will be keep too, as there are many simulator models that working with local xyz coordination systems. 
There are currently three geo-reference formats available in OpenDRIVE, I think the z-axis basically refers to ellipsoidal height, and the other coordinate system the z-axis refers to sea lavel elevation.

As an additional question, I can't know the NDS Map specifications, but are the geo-reference specifications used fixed in NDS?
Does the newly definition of VehicleGEOPosition in the OSI message have a goal that the referenced geo-reference specification is fixed? (e.g use NDS's geo-reference?)

ThomasNaderBMW: Sry for the late reply.
I gave it a new structure, can you both have a look?
To the questions:
- NDS information shall be possible to be transported
- proj4-String of OpenDrive has to be discussed as it is used to calculate the WGS84-Position (End-Result) as an Input-Parameter. Is this information necessary beside the simulation-tool, that is parsing the OpenDrive?

PR #409, opened 2020-08-18
Introduction of Traffic Participants and releated Top-Level Messages

pmai:
The basis of this PR is the consolidated traffic participant feature branch from the SETLevel4To5 project, which provides for traffic participant modeling through the newly introduced TrafficUpdate and TrafficCommand top-level messages. Note that this initial contribution has already been enhanced inside the ASAM project as well.

This draft PR is created in order to highlight this proposal and allow coordinated further work on it inside the relevant ASAM WPs.

#### Check the checklist

### DIFF
--- a/osi_trafficcommand.proto
+++ b/osi_trafficcommand.proto
@@ -90,11 +90,11 @@ message TrafficAction
     // A SpeedAction
     //
     optional SpeedAction speed_action = 5;    
-	
+
     // An AbortActionsAction
     //
     optional AbortActionsAction abort_actions_action = 6; 
-	
+
     // An EndActionsAction
     //
     optional EndActionsAction end_actions_action = 7; 
@@ -475,7 +475,7 @@ message AbortActionsAction
     // The Action Header
     //
     optional ActionHeader action_header = 1;
-	
+
     // Actions which must be aborted immediately.
     // 
     // These fields hold the action ids of the actions that must be 
@@ -497,11 +497,11 @@ message EndActionsAction
     // The Action Header
     //
     optional ActionHeader action_header = 1;
-	
+
     // Actions which are regarded as successfully executed.
     // 
     // These fields hold the action ids of the actions that are regarded 
-    // as successfully executed and shall be terminated gracefully.	
+    // as successfully executed and shall be terminated gracefully.
     //
     repeated Identifier target_action_id = 2; 
 }
--- a/osi_trafficcommand.proto
+++ b/osi_trafficcommand.proto
@@ -80,6 +80,10 @@ message TrafficAction
     // A LaneChangeAction
     //
     optional LaneChangeAction lane_change_action = 4; 
+
+    // A SpeedAction
+    //
+    optional SpeedAction speed_action = 5;    
 }
 
 //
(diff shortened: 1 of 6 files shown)


osi_trafficcommand.proto
```diff
@@ -0,0 +1,507 @@
...
+    // An AbortActionsAction
+    //
+    optional AbortActionsAction abort_actions_action = 6; 
+
+    // An EndActionsAction
+    //
+    optional EndActionsAction end_actions_action = 7; 
+}
+
+//
```
clemenshabedank, line 103: As discussed with @caspar-ai and @Krishna626 this might be a candidate for osi_common.proto (which is imported by this message), because the definition could be re-used in other messages (e.g. TrafficUpdate). TBD prior merging.

osi_trafficcommand.proto
```diff
@@ -0,0 +1,507 @@
...
+// \brief The StatePoint definition
+//
+// \note The StatePoint definition does not define mandatory fields.
+// The context defines how and what fields are used.
+// For example: FollowPathAction does not use timestamp in contrast to
+// FollowTrajectoryAction. 
+//
+message StatePoint
```
clemenshabedank, line 111: As discussed with @caspar-ai and @Krishna626 the SteatePoint definition might be a candidate for osi_common.proto (which is imported by this message), because the definition could be re-used in other messages (e.g. TrafficUpdate). TBD prior merging.

pmai: It depends: For the actual state update in TrafficUpdate more fields are needed. For reuse as the future trajectory points as proposed by @caspar-ai I think this makes potential sense. We could however do the move in the pull request that adds the future trajectory points, since it already is a global message that should not break anything...

pmai:
Two additional notes:

- OpenSimulationInterface/osi-sensor-model-packaging#56 would need to be reviewed and merged as well, to actually have traffic participant models.

- I'll rebase the PR with DCO and updated against master prior to actual merging (if the CCB decides to merge), but will only do so at the last minute, to not confuse people.

osi_trafficcommand.proto
clemenshabedank: Yes, that's what @caspar-ai did here https://github.com/OpenSimulationInterface/open-simulation-interface/pull/401#issuecomment-722533543. So I mark this conversation as resolved.

clemenshabedank (commented):
Discussed in WP11. Ready for CCB-Review (in case no more blocking comments until next CCB-meeting).

clemenshabedank (approved):

kmeids:
**Output from CCB meeting - 11.11.2020**

Actions:
1. Documentation review to be performed before the minor release (by next week).
2. Pull request to be merged by @pmai after documentation review.

paagkame:
@pmai I have done a language review on the inline code documentation (see commit 9d51c019b944ad942011953db39a29a31d38fc49). I still have a few things I would like to clarify:

**trafficcommand.proto**
1. "Recommended is the starting time point of the simulation.": What exactly does that refer to? The start time is somehow without context, but is it recommended. For what is it recommended? Set timestamp = start time?
2. Use of "shall", for example, "only certain combinations of atomic traffic actions shall be transmitted within certain time intervals.": Is this a fact? If so, why not say "are transmitted"?
3. "The orientation fields can be set depending on the constrain_orientation field being true.": Is there a hidden condition in the sentence? If so, here is my suggestion to make it more explicit: "The orientation fields can be set if the constrain_orientation field are true."

**trafficupdate.proto**
1. "[...] members are considered in updates": What does "consider in updates" exactly mean? Do they receive updates? Why not express it that way?

I haven't change anything else for the user manual. I guess the proto files will be included at [File Hierachy](https://opensimulationinterface.github.io/osi-documentation/open-simulation-interface/library_root.html) automatically.

Do I close the pull request after clarification or do you?

osi_trafficupdate.proto
```diff
@@ -13,13 +13,12 @@ package osi3;
...
+// \note For reasons of convenience and consistency, the
 // updated information is provided as a MovingObject.  Certain fields
-// of this sub-message need not be set and will be ignored by the
-// simulation environment, because they are in fact static information.
+// of this sub-message are not required to be set and will be ignored by the
+// simulation environment, because they are static information.
 // Instead of creating a seperate message type for only the non-static
-// information, re-use of the existing message was considered more
-// appropriate.
+// information, re-use existing message.
```
pmai, line 21: I actually think that this whole sentence should be dropped, since it gives rationale and is not part of the definition.

pmai:
> **trafficcommand.proto**
> 
> 1. "Recommended is the starting time point of the simulation.": What exactly does that refer to? The start time is somehow without context, but is it recommended. For what is it recommended? Set timestamp = start time?

The sentence suggests that it would be a good idea (but not required) to have the 0 timestamp coincide with the actual start of the simulation, whatever that turns out to be (OSI does not specify overall simulation semantics and timing issues, hence we can't be much more specific).

> 2. Use of "shall", for example, "only certain combinations of atomic traffic actions shall be transmitted within certain time intervals.": Is this a fact? If so, why not say "are transmitted"?

Since nearly nothing in a standard like OSI is a fact: Everything is just prescriptive, i.e. this and this should be the case (putting the burden of actually doing it on the implementations, since OSI itself cannot enforce this). Hence the frequent use of SHALL, MUST, MUST NOT, etc.

We should however be more consistent in the wording there, i.e. it would IMHO be a good idea to switch to RFC 2119 (https://tools.ietf.org/html/rfc2119) for the normative wording. This might be something to discuss in one of the next CCB/Architecture meetings.

> 3. "The orientation fields can be set depending on the constrain_orientation field being true.": Is there a hidden condition in the sentence? If so, here is my suggestion to make it more explicit: "The orientation fields can be set if the constrain_orientation field are true."

Yes, that seems better.

> 
> **trafficupdate.proto**
> 
> 1. "[...] members are considered in updates": What does "consider in updates" exactly mean? Do they receive updates? Why not express it that way?

The traffic participant provides updates of its state back to the simulation engine; the simulation engine will only consider (i.e. take into account) the specified fields, and ignore all other fields. So the simulation engine might receive more updated fields, but will only take the given subset into account. Not sure how to best phrase that...

> Do I close the pull request after clarification or do you?

I'll do the final merge, because I need to provide the missing sign-offs.

Thanks for the rewrites!

osi_trafficupdate.proto
```diff
@@ -0,0 +1,50 @@
...
+    // the notional simulation time the data applies to and the time it was sent.
+    // There is no inherent latency for moving object update data, as opposed
+    // to sensor data.
+    //
+    optional Timestamp timestamp = 2;
+
+    // Updated moving object
+    //
+    // \note Only the id, base member without dimension and base_polygon
+    // and the vehicle_classification.light_state members are considered in
```
paagkame: Just checking if the period between the "vehicle_classification.light_state is correct". If so, I am done with the review from documentation point of view.

kmeids:
Actions:
1. Documentation review is done and pull request is ready to be merged, by @pmai

PR #429, opened 2020-10-15
Fix message iteration by index and range

### LINKED ISSUE #426: Python OSITrace method get_messages_in_index_range() ignores arguments for *.osi format
#### Describe the bug
In the Python OSI Trace reader (`format/OSITrace.py`), when using the binary OSI trace format (`type_name=None`), the method `get_messages_in_index_range(bening, end)` ignores the arguments `begin` and `end`. Instead, it always returns an iterator over *all* messages. Because of that, the method `get_message_by_index(index)` always returns the first message regardless of the `index` argument.

#### Describe how to reproduce the bug
```Python
tr = OSITrace()
tr.from_file("MySensorViewTrace.osi")
l = list(tr.get_messages_in_index_range(10, 12))
m = tr.get_message_by_index(5)
```
`l` contains all message instead of messages 10, 11 and 12, `m` is message 0 instead if message 5.

`l` contains only the messages 10, 11 and 12 (or at least a documentation that it does not work as expected).

#### Describe the OS you are using
 - OS: Windows 10
 - Python 3.6.8 64-bit

#### Additional context
Possible fix:

1. In the "OSI"-branch of the method, replace `self.scenario_file.seek(0)` with `self.scenario_file.seek(self.message_offsets[begin])`
2. in the `while` loop condition, end with `min(len(serialized_message), self.message_offsets[end+1])` (requires additional index range check for `self.message_offsets`)

vkresch:
For message iteration there is currently a bug in which the range cannot be specified (https://github.com/OpenSimulationInterface/open-simulation-interface/issues/426).

#### Add a description
What is this change?
This PR replaces hte while loop iteration with the for loop iteration over the messages inspired from the osi-validator.

What does it fix?
It fixes the index specification of OSI trace files. (Fixes #426)

How has it been tested?
I ran manual test with an osi trace file to get messages by index. I also tested the conversion from *.txt -> *.osi and made the *.osi file also readable.

#### Mention a member
@tpajenkamp-dspace let me know if this PR fixes your issue. Please test it with your trace file and let me know if it works.

#### Check the checklist

### DIFF
--- a/format/OSITrace.py
+++ b/format/OSITrace.py
@@ -196,7 +196,7 @@ class OSITrace:
             for rel_index, rel_message_offset in enumerate(rel_message_offsets):
                 rel_begin = rel_message_offset + self._int_length
                 rel_end = (
-                    rel_message_offsets[rel_index + 1] - self._int_length
+                    rel_message_offsets[rel_index + 1]
                     if rel_index + 1 < len(rel_message_offsets)
                     else message_sequence_len
                 )
@@ -209,8 +209,6 @@ class OSITrace:
         else:
             self.scenario_file.close()
             raise Exception(f"The defined format {self.format_type} does not exist.")
-
-        self.scenario_file.close()
 
     def make_readable(self, name, interval=None, index=None):
         self.scenario_file.seek(0)
--- a/format/OSITrace.py
+++ b/format/OSITrace.py
@@ -44,6 +44,7 @@ class OSITrace:
         self.type_name = type_name
         self.timestep_count = 0
         self.retrieved_scenario_size = 0
+        self._int_length = len(struct.pack("<L", 0))
 
     def from_file(self, path, type_name="SensorView", max_index=-1, format_type=None):
         """Import a scenario from a file"""
@@ -161,19 +162,18 @@ class OSITrace:
         """
         Yield an iterator over messages of indexes between begin and end included.
         """
+        self.scenario_file.seek(self.message_offsets[begin])
+        abs_first_offset = self.message_offsets[begin]
+        abs_last_offset = self.message_offsets[end] \
+            if end < len(self.message_offsets) \
+            else self.retrieved_scenario_size
+
+        rel_message_offsets = [
+            abs_message_offset - abs_first_offset
+            for abs_message_offset in self.message_offsets[begin:end]
+        ]
 
         if self.format_type == "separated":
-            self.scenario_file.seek(self.message_offsets[begin])
-            abs_first_offset = self.message_offsets[begin]
-            abs_last_offset = self.message_offsets[end] \
-                if end < len(self.message_offsets) \
-                else self.retrieved_scenario_size
-
-            rel_message_offsets = [
-                abs_message_offset - abs_first_offset
-                for abs_message_offset in self.message_offsets[begin:end]
-            ]
-
             message_sequence_len = abs_last_offset - \
                 abs_first_offset - SEPARATOR_LENGTH
             serialized_messages_extract = self.scenario_file.read(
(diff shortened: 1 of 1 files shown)


format/OSITrace.py
```diff
@@ -190,17 +190,20 @@ def get_messages_in_index_range(self, begin, end):
...
-            serialized_message = self.scenario_file.read()
-            INT_LENGTH = len(struct.pack("<L", 0))
-            message_length = 0
+            message_sequence_len = abs_last_offset - abs_first_offset
+            serialized_messages_extract = self.scenario_file.read(message_sequence_len)
+
+            for rel_index, rel_message_offset in enumerate(rel_message_offsets):
+                rel_begin = rel_message_offset + self._int_length
+                rel_end = (
+                    rel_message_offsets[rel_index + 1] - self._int_length
```
tpajenkamp-dspace, line 199: `self._int_length` must not be subtracted

tpajenkamp-dspace:
@vkresch Thanks for the update, now `get_message_by_index` works without problems.

However, there is a bug  in `get_messages_in_index_range` if more than one message is extracted. I created a comment directly at the offending code line. Currently, I get:
```
google.protobuf.message.DecodeError: Truncated message.
```
This also affects other methods like `get_messages` which call `get_messages_in_index_range`.

Also, I noticed another possible problem in this method which is technically another issue and may be somewhat intended. When the generator returned by `get_messages_in_index_range` is consumed (reaches `StopIteration`), the underlying file is closed so that following calls to most methods cause `ValueError: seek of closed file` or similar errors.

tpajenkamp-dspace (commented):
@vkresch Thanks for the update, now `get_message_by_index` works without problems.

However, there is a bug  in `get_messages_in_index_range` if more than one message is extracted. I created a comment directly at the offending code line. Currently, I get:
```
google.protobuf.message.DecodeError: Truncated message.
```
This also affects other methods like `get_messages` which call `get_messages_in_index_range`.

Also, I noticed another possible problem in this method which is technically another issue and may be somewhat intended. When the generator returned by `get_messages_in_index_range` is consumed (reaches `StopIteration`), the underlying file is closed so that following calls to most methods cause `ValueError: seek of closed file` or similar errors.

format/OSITrace.py
vkresch: For some reason I cannot reproduce the bug with more than one message extraction with my osi file. Can you maybe share your osi file if possible? Try to validate your osi file with osi visualizer or the osi validator. 

The other issue is the generator property of python. Generators generate the elements on the fly and throw them away. To prevent it would be creating a deep copy of the generator as an iterator.

vkresch: Maybe an error exceptions needs to be added so that the user knows whats going on. Good point!

tpajenkamp-dspace: First of all, I would argue with logic. The first `message_offset` is `0`. Every message offset indicates the first byte of the message length indicator of its following serialized message. Thus, it is the byte right after the previous binary message (first offset excluded). The length indicator offset is already taken into account by the computation of `rel_begin`. By also subtracting `self._int_length` from the next message offset to compute `rel_end`, an incomplete serialized message is read missing the last 4 bytes. The next operation starts at the next `message_offset` entry so that these 4 bytes are skipped and never processed.

The osi-validator suffers from the same problem.

I have no appropriate means for file uploads, but this simple script generates an OSI trace file. It uses the same mechanism to pack the osi messages as `txt2osi.py`. The bug is triggers in the last line. If I apply my proposed fix, it works.
```python
import struct
from osi3.osi_version_pb2 import InterfaceVersion, current_interface_version
from osi3.osi_sensorview_pb2 import SensorView

import OSITrace

filePath = "osi_SensorView_tmp.osi"

sv = SensorView()
sv.version.CopyFrom(InterfaceVersion.DESCRIPTOR.file.GetOptions().Extensions[current_interface_version])
with open(filePath, "wb") as fp:
    for i in range(15):
        sv.timestamp.seconds = 0
        sv.timestamp.nanos = i * 5
        msgBin = sv.SerializeToString()
        filePath = r"_osi_SensorView_tmp.osi"
        fp.write(struct.pack("<L", len(msgBin)))
        fp.write(msgBin)

osiTrace = OSITrace.OSITrace()
osiTrace.from_file(filePath, type_name="SensorView")
msgs = list(osiTrace.get_messages())
```

vkresch: Yes you are right! I checked with different files. But I don't get the "Truncated error message". When decoding there are missing fields when I keep the `self._int_length` subtraction happening see output below:

image

I will adjust it and will also add a meaningful error message when the message generator reaches the end! Thanks for reporting this issue!

tpajenkamp-dspace:
With the latest commit, it works for me. Thank you.

vkresch:
@jdsika please merge. Thanks ;)

jdsika (approved):

PR #430, opened 2020-10-16
Extension by wheel data

ThomasNaderBMW:
#### Add a description
So here is a pull request that adds information about the wheels of a vehicle.
It should be a good basement, but for sure there is potential to discuss especially the reference systems.
Feel free to give input! :)

**Some questions to ask**:
What is this change?
It is about the perceivable information of the wheels.
What does it fix?
It can be used for sensor models or simply for visualization reasons.
Minor Change, it is a feature.
How has it been tested?
Not yet, but discussed with implementors of visualization interfaces.

#### Mention a member
Add @mentions of the person or team responsible for reviewing proposed changes.
@caspar-ai , @pmai 

#### Check the checklist


kmeids:
Hello @ThomasNaderBMW ,
Considering that in some cases not all wheels have the same radius (e.g. agriculture tractor, front wheels are smaller) would it make sense to move **optional double radius_wheel = 2;** under your newly proposed message **WheelData** ?

ThomasNaderBMW:
@kmeids : Thx for the feedback. I also was thinking about to move it. Now I did it!
What do you think about the number_wheels? It is also redundant if you simply count the fields of wheelData..

kmeids:
@ThomasNaderBMW: I would recommend to keep the number of wheels in case some simulation environments can't/won't provide the detailed WheelData message. (or maybe in case osi implementation is taking place at an incremental level)

kmeids:
@ThomasNaderBMW: Based on our discussion today here are some additional comments:

1. on _rotation_rate_, instead of _Positive means, the vehicle moves forward_ (since wheels can move in one direction and the vehicle can be moving in the opposite direction, i.e. slipping) let's go with the counterclockwise and clockwise convention where counterclockwise is usually associated with +v values and clockwise with -v ones. (we still need to define from which perspective, the axle can be a good reference point)
2. I also suggest to extend the _radius_ to include data about the _rim_  radius (additional field), the initial _radius_ definition can still remain the same.

raue:
As mentioned in todays discussion. The purpose/use-case and of course the limitations of having the wheel information should be somehow precised. Otherwise people might look at it from different perspectives and expectations may arise, which could go beyond to what is intended - e.g. ABS/ESP folks.

If the purpose is more focused on visualization, then I would suggest taking the so-called [wagon-wheel-effect](https://en.wikipedia.org/wiki/Wagon-wheel_effect) into account. This nasty behaviour might ruin the whole visual appearance.

osi_object.proto
```diff
@@ -436,6 +437,69 @@ message MovingObject
...
+            optional double radius_wheel = 4;
+
+            // Radius of the rim measured from the center to the outer, visible part of the rim.
+            //
+            // Unit: m
+            //
+            optional double radius_rim = 5;
+
+            // Orientation of the wheel.
+            // The reference system is the vehicle frame (center of bounding box).
```
pmai, line 487:
```suggestion
            //
            // The reference system is the vehicle frame (center of bounding box).
```

osi_object.proto
```diff
@@ -436,6 +437,69 @@ message MovingObject
...
+            // Right-handed, x-axis shows into driving direction "forward".
+            // Rough Approximation: The z-value shows the damping.
+            //
+            optional Vector3d position = 3;
+
+            // Radius of the wheels measured from the center to the outer part of the tire.
+            //
+            // Unit: m
+            //
+            optional double radius_wheel = 4;
```
pmai, line 478:
```suggestion
            optional double wheel_radius = 4;
```

osi_object.proto
```diff
@@ -436,6 +437,69 @@ message MovingObject
...
+            optional uint32 index = 2;
+            
+            // Position of the wheel (geographical center of the wheel).
+            // The reference system is the vehicle frame (center of bounding box).
+            // Right-handed, x-axis shows into driving direction "forward".
+            // Rough Approximation: The z-value shows the damping.
+            //
+            optional Vector3d position = 3;
+
+            // Radius of the wheels measured from the center to the outer part of the tire.
```
pmai, line 474:
```suggestion
            // Median radius of the wheels measured from a center of the wheel to
            // the outer part of the tire.
            //
```

osi_object.proto
```diff
@@ -436,6 +437,69 @@ message MovingObject
...
+            //
+            // Unit: m
+            //
+            optional double radius_wheel = 4;
+
+            // Radius of the rim measured from the center to the outer, visible part of the rim.
+            //
+            // Unit: m
+            //
+            optional double radius_rim = 5;
```
pmai, line 484:
```suggestion
            optional double rim_radius = 5;
```

osi_object.proto
```diff
@@ -436,6 +437,69 @@ message MovingObject
...
+            // Orientation of the wheel.
+            // The reference system is the vehicle frame (center of bounding box).
+            // Right-handed, x-axis shows into driving direction "forward".
+            // Rough Approximation: The yaw-value shows the steering angle.
+            //
+            optional Orientation3d orientation = 6;
+
+            // Rotation of the wheel around the axle.
+            //
+            // Unit: Rad/s.
```
pmai, line 495:
```suggestion
```

osi_object.proto
```diff
@@ -436,6 +437,69 @@ message MovingObject
...
+            // Counterclockwise is positive, clockwise negative.
+            // The reference point / view perspective is from the mid of the axle.
+            // If there is just one wheel centered in the mid of the axle (e.g. bicycle), 
+            // the wheel is treated as a right wheel.
+            //
```
pmai, line 500:
```suggestion
            //
            // Unit: Rad/s.
            //
```

pmai (changes requested):
Looking good, with these proposed changes for cleanup + @kmeids and @raue comments.

ThomasNaderBMW:
@kmeids  and @pmai:
Please have a short look. I made most of the corrections, but I am still no sure if the description of the orientation rate is sufficient.. What do you think?

osi_object.proto
```diff
@@ -436,6 +437,86 @@ message MovingObject
...
+            // The reference system is the vehicle frame (center of bounding box).
+            // Right-handed, x-axis shows into driving direction "forward".
+            // Rough Approximation: The z-value shows the damping.
+            //
+            optional Vector3d position = 3;
+
+            // Median radius of the wheel measured from the center of the wheel to
+            // the outer part of the tire.
+            //
+            //
```
caspar-ai, line 479:
```suggestion
```

osi_object.proto
```diff
@@ -436,6 +437,86 @@ message MovingObject
...
+            // For example, on a standard 2-axle, 4-wheel car, the rear-right
+            // wheel would be (axle=1, index=0).
+            // This concept works also for twin tires.
+            //
+            optional uint32 index = 2;
+            
+            // Position of the wheel (geomatric center of the wheel).
+            // The reference system is the vehicle frame (center of bounding box).
+            // Right-handed, x-axis shows into driving direction "forward".
+            // Rough Approximation: The z-value shows the damping.
```
caspar-ai, line 472: To clarify here, do you mean that the z-value should take into account things like suspension damping?  Rather than the position being fixed, e.g. while the car is at rest?

(is there a particular reason to highlight what the z-value is for?  Is it to help explain the convention?  If so, I think "x forward" from the previous line is sufficient, but don't feel strongly about it)

osi_object.proto
```diff
@@ -436,6 +437,86 @@ message MovingObject
...
+            //
+            // Unit: m
+            //
+            optional double width = 6;
+
+            // Orientation of the wheel.
+            //
+            // The reference system is the vehicle frame (center of bounding box).
+            // Right-handed, x-axis shows into driving direction "forward".
+            // Rough Approximation: The yaw-value shows the steering angle.
```
caspar-ai, line 500: There is (technically) no x-axis for an `Orientation3d` object.  Perhaps:

```suggestion
            // Right-handed relative to the vehicle's forward motion.  The yaw-value shows the steering angle.
```

osi_object.proto
```diff
@@ -436,6 +437,86 @@ message MovingObject
...
+            //
+            optional Orientation3d orientation = 7;
+
+            // Rotation rate in the wheel's frame of reference (rotation around the wheel axle).
+            //
+            // Unit: rad/s.
+            //
+            // The sign convention is defined using the right-hand rule with respect to the vehicle's coordinate system (DIN ISO 8855:2013-11):
```
caspar-ai, line 508: This might be a very noddy question, so sorry in advance!

This is defined as (counter-)clockwise rotation around the wheel axle, is that always when considered from the y-positive direction of the vehicle's co-ordinate system?  i.e. both wheels on an axle will typically have the same sign on their rotation rate.

Might be worth adding in a basic example to clarify.  (or let me know if this is so commonly understood by people who know about these things it isn't worth clarifying!)

caspar-ai (commented):
Looks good, a few suggestions / questions but only around the commenting.

osi_object.proto
```diff
@@ -436,6 +437,86 @@ message MovingObject
...
+            // The index of the wheel on the axle, counting in the direction
+            // of positive-y, i.e. right-to-left.
+            //
+            // For example, on a standard 2-axle, 4-wheel car, the rear-right
+            // wheel would be (axle=1, index=0).
+            // This concept works also for twin tires.
+            //
+            optional uint32 index = 2;
+            
+            // Position of the wheel (geomatric center of the wheel).
```
caspar-ai, line 469:
```suggestion
            // Position of the wheel (geometric center of the wheel).
```

osi_object.proto
kmeids: @caspar-ai , yes that is true, this means that both wheels on an axle will typically have the same sign on their rotation rate.
We also discussed adding pictures and examples, it was left for @ThomasNaderBMW  to do.

ThomasNaderBMW: Yes the "forward" concept was in my first proposal. But the definition is no very scientific, so we changed to the right hand rule.. But to be honest I still think that it is even more complicated now and not self-explaining..
@kmeids , what do you think. Shall we change to an easier convention?

osi_object.proto
ThomasNaderBMW: Yes, I would see the damping representation in these coordinates (mostly z, but also x & y).
The line could be deleted from my side, it is just a hint, what you can do with the position.. @CCB

osi_object.proto
kmeids: @ThomasNaderBMW, no need to change it, I think with a suitable pictogram there would be no problem understanding it.

ThomasNaderBMW: @kmeids ah ok. Could that be a task for the service provider? That would help a lot, especially to demonstrate it in an understandable way!

kmeids: @ThomasNaderBMW, yes it can be however in this case someone should still clarify the issue to the service provider.
Just added the "Documentation" label to it.

kmeids:
**Output from CCB meeting - 11.11.2020**

Actions:

1. @clemenshabedank will follow-up with @ThomasNaderBMW  on the following comment : _Rough Approximation: The z-value shows the damping_.
2. Documentation review to be performed before the minor release.
3. Pull request to be merged by @ThomasNaderBMW **after** documentation review.

osi_object.proto
paagkame: Hi everyone, using an image to introduce and explain complex issues is a good idea from my point of view. Shall that image already be included in the next minor release?

osi_object.proto
```diff
@@ -436,6 +437,84 @@ message MovingObject
...
+            // The index of the wheel on the axle, counting in the direction
+            // of positive-y, i.e. right-to-left.
+            //
+            // For example, on a standard 2-axle, 4-wheel car, the rear-right
+            // wheel would be (axle=1, index=0).
+            // This concept works also for twin tires.
+            //
+            optional uint32 index = 2;
+            
+            // Position of the wheel (geometric center of the wheel).
```
paagkame, line 469: Does this line mean the position is measured from the center of the wheel?

osi_object.proto
```diff
@@ -436,6 +437,84 @@ message MovingObject
...
+            // The reference system is the vehicle frame (center of bounding box).
```
paagkame, line 470: Is that mean the vehicle frame is also the center of the bounding box?

osi_object.proto
```diff
@@ -436,6 +437,84 @@ message MovingObject
...
+            optional double width = 6;
+
+            // Orientation of the wheel.
+            //
+            // The reference system is the vehicle frame (center of bounding box).
+            // Right-handed relative to the vehicle's forward motion.  The yaw-value shows the steering angle.
+            //
+            optional Orientation3d orientation = 7;
+
+            // Rotation rate in the wheel's frame of reference (rotation around the wheel axle).
```
paagkame, line 502: This is a bit tricky for me to understand. Did I got it right: 
There is a wheel frame that you use as a reference. The wheel frame used in the simulation may vary in size. You "measure" the distance of the wheel frame used in the simulation for one complete rotation and relate this to the reference wheel frame. That means the value describes how often the reference wheel frame rotates around its axle to equal the distance that the wheel frame used in the simulation would cover in one complete rotation.

osi_object.proto
kmeids: yes, more precisely when there is no load on the wheel, since that could shift it's center position and shape.

osi_object.proto
kmeids: This does not mean that the vehicle frame is also **the** center of the bounding box.
According to the following [Detailed Description](https://opensimulationinterface.github.io/open-simulation-interface/structosi3_1_1MovingObject_1_1VehicleAttributes.html):
_For all vehicles, including host vehicles, the position given in MovingObject::base . BaseMoving::position points to the center of the vehicle's bounding box._

Looking at the picture below, the bounding box consists of the doted 3D lines and the reference coordinate system is the one circled in red, which lies in the center of the bounding box.

_// The reference system is the vehicle frame (center of bounding box)._ is referring to the same coordinate system.

image

osi_object.proto
kmeids: I am afraid I don't really get you point, let's discuss this offline.

osi_object.proto
clemenshabedank: I think it would be great, because there was a lot of discussion already going on in the CCB meeting!

osi_object.proto
clemenshabedank: @paagkame good point, so I think it would be the vector from center of bounding box to the center of the wheel. I think it would make sense to write a vector description like in line 418ff. of osi_object.proto and maybe explain the center of the wheel better and the implications as @kmeids attempts. I guess also steering might change the center of the wheel position. So I think it has to be clear that this vector will not have a constant value. (so I hope I have the correct understanding here).

osi_object.proto
paagkame: @ThomasNaderBMW Can you provide an image that I can include in the documentation?

kmeids: CCB Meeting 25.11.2020:

Actions:
@ThomasNaderBMW to update the coordinate system reference point.
@paagkame to upload a pictogram illustrating the sign convention for rotation.

osi_object.proto
ThomasNaderBMW: I simplified that, please have a check @paagkame and @kmeids

osi_object.proto
ThomasNaderBMW: I changed the description. Please give some feedback @kmeids , @clemenshabedank , @paagkame

osi_object.proto
ThomasNaderBMW: I reworked that, please have a short check @kmeids  and @paagkame .
There is also a line for the picture @paagkame . Thank you very much in advance!! :)

osi_object.proto
ThomasNaderBMW: I reworked the open actions with the last commit. I hope it is clearer now! :)

osi_object.proto
clemenshabedank: Ok I think the description should be precise now, if it is supposed to be a not-changing field.

osi_object.proto
kmeids: Good from my side!

paagkame: Good from my side, too.

osi_object.proto
paagkame: I have created a new svg. Please have a look at it at open-simulation-interface\doc\images\OSI_RotationRate.svg
I am not sure though if the use of color is allowed and where to place the image best. The .proto file does not seem to be the right place.

osi_object.proto
ThomasNaderBMW: We discussed the point now and did a small adaption. It's ok now.

osi_object.proto
kmeids: @paagkame 
I assume that by "F" you are referring to the moving direction (Forward).
This can be confused by F as in Force.
To evade any confusions, we should either state that F is referring to forward or just extend the arrow a bit linger and write "moving direction"

image

kmeids: @paagkame 
Could you also remove the "V" after the "+" and "-" sign?

kmeids:
Output from CCB meeting - 09.12.2020

Actions:
1. @paagkame could you please follow up on this?  https://github.com/OpenSimulationInterface/open-simulation-interface/pull/430#discussion_r536631673

osi_object.proto
paagkame: SVG is updated with proposed changes.

kmeids:

Actions:
1. Pull request ready to be merged, by @ThomasNaderBMW.

kmeids:
Hello @ThomasNaderBMW, is it possible to have this pull request merged before Christmas ? :D

pmai (approved):
Approved by CCB, ready for merge after build

PR #441, opened 2020-11-06
Extension of the hostVehicleData

ThomasNaderBMW:
As this is a bigger PR, feedback is welcome! :)

Out of WP15. Integration in traffic_update has to be done.
-> Interface between dynamics (incl. restbussimulation) and simulation environment. 

#### Add a description
Host vehicle data is about the perception of the vehicle about it's own, internal states.
It describes data that the host vehicle knows about itself,
e.g. from location sensors, internal sensors, board net etc.
A dynamic model can serve as input provider.
Sensors, mockups or other modules can make usage of the host vehicle data.
It consists of different messages categorizing the vehicle in:
Vehicle-Basics, Vehicle-Powermanagement, Vehicle-Powertrain, Vehicle-SteeringWheel, Vehicle-Wheels, Vehicle-Localization.

#### Mention a member
@all

#### Check the checklist


osi_hostvehicledata.proto
```diff
@@ -2,31 +2,352 @@ syntax = "proto2";
...
+        }
+
+        //
+        // \brief Current estimated location based on the output of the GPS-Unit.
+        //
+        message GeoreferencedInformation
+        {
+            // Longitude in decimal degrees regarding WGS84.
+            //
+            // Unit: Degree
```
ThomasSchloemicherAVL, line 319: Degrees in WGS84 datum, range: [-180; 180]

osi_hostvehicledata.proto
```diff
@@ -2,31 +2,352 @@ syntax = "proto2";
...
+            //
+            optional double longitude = 1;
+
+            // Latitude in decimal degrees regarding WGS84.
+            //
+            // Unit: Degree
+            //
+            optional double latitude = 2;
+
```
ThomasSchloemicherAVL, line 328: Degrees in WGS84 datum, range: [-90; 90]

osi_hostvehicledata.proto
```diff
@@ -2,31 +2,352 @@ syntax = "proto2";
...
+            // Height above sea level regarding EGM96.
+            //
+            // Unit: m
+            //
```
ThomasSchloemicherAVL, line 332: Range [-300;10000]

osi_hostvehicledata.proto
```diff
@@ -2,31 +2,352 @@ syntax = "proto2";
...
+            optional double altitude = 3;
+
+            // To be discussed as it is in CartesianInformation. 
+            // Heading in decimal degrees.
+            //
+            // Unit: Degree
+            //
```
ThomasSchloemicherAVL, line 339: Range [0;360]

ThomasSchloemicherAVL (commented):
Please add also to SensorData
And please add AD-Function states

osi_hostvehicledata.proto
ThomasNaderBMW: Yes, it's really limited to these values?

osi_hostvehicledata.proto
```diff
@@ -2,31 +2,356 @@ syntax = "proto2";
...
+            // Unit: %
+            //
+            optional double slip = 6;
+        }
+    }
+
+    //
+    // \brief This message contains all the information the vehicle knows about its kinematic states.
+    //
+    message VehiclePositionAndKinematics
```
u21n80, line 292: Is it possible to provide a timestamp to the VehiclePositionAndKinematics?

osi_hostvehicledata.proto
```diff
@@ -2,31 +2,356 @@ syntax = "proto2";
...
+    {
+        //
+        // \brief Current kinematic data based on the output of the Intertial Measurement Unit (IMU).
```
pmai, line 295: This position, and likely also the Georeferenced Information, are unlikely to be the direct output of any IMU or GPS directly, but are the output of the overall position estimation function of the host vehicle, which might combine multiple location sources (IMU, multiple GPS receivers, other sensors, cross-correlation etc.). Raw sensor outputs are travelling via SensorData, or potentially needed further messages. I think this should be made clearer in the descriptions...

osi_hostvehicledata.proto
```diff
@@ -2,31 +2,356 @@ syntax = "proto2";
...
+
+            // Accuracy of localization measurement in percentage of the units.
+            //
+            // Unit: %
+            //
+            // optional double localization_accuracy = 5;
+
+            // Number of satellites.
+            //
+            // optional int32 number_of_satellites = 6;
```
pmai, line 354: are these fields supposed to be commented out?

osi_hostvehicledata.proto
```diff
@@ -2,31 +2,356 @@ syntax = "proto2";
...
     //
     optional BaseMoving location_rmse = 2;
+
+    // The ID of the object. Can be discussed.
+    //
+    // \rules
+    // is_globally_unique
+    // \endrules
+    //
+    optional Identifier id = 3;
```
pmai, line 47: probably should have a longer name like host_vehicle_id, to make the semantics clearer (cf. some of the identifier discussions). Or should be dropped, since it is, s you allude to, not really known to the vehicle itself, but comes from the simulation context (i.e. SensorView and TrafficUpdate already carry the host vehicle ID).

osi_hostvehicledata.proto
```diff
@@ -2,31 +2,356 @@ syntax = "proto2";
...
 
-    // Current estimated location error based on GPS- and related navigation
+    // Deprecated: Will be removed in next major release. Moved to VehiclePositionAndKinematics.
+    // Current estimated location error based on GPS- and related navigation 
     // sensors.
     //
     // \note Note that dimension and base_polygon need not be set.
     //
     optional BaseMoving location_rmse = 2;
+
```
pmai, line 40: We might want to add fields for the current time of day, as known to the vehicle, and a simulation-time current timestamp to indicate for which point in simulation time the fields of this message are considered valid (which indicates the time when the car networks would have transmitted this data, which is different from individual measurement times that went into the calculations for the values).

pmai (commented):
General comments from review and #463 cross-related issues.

osi_hostvehicledata.proto
```diff
@@ -2,31 +2,356 @@ syntax = "proto2";
 
 option optimize_for = SPEED;
 
+import "osi_version.proto";
 import "osi_common.proto";
+import "osi_object.proto";
```
ThomasSchloemicherAVL, line 8: Got following warning at the last CI unit test regarding osi_object.proto
`osi_hostvehicledata.proto: warning: Import osi_object.proto but not used.`

ThomasNaderBMW: Good point, I deleted the import!

osi_hostvehicledata.proto
ThomasSchloemicherAVL: From my point of new not. At least having the heading in the message would be beneficial for sensor systems.

osi_sensordata.proto
```diff
@@ -297,4 +300,8 @@ message SensorData
     // object hypothesis and tracking.
     //
     optional FeatureData feature_data = 26;
+
+    // Host vehicle data measured by the sensors.
+    //
+    optional HostVehicleData host_vehicle = 27;
```
ThomasSchloemicherAVL, line 306: @pmai can we add the time stamp as a DetectedEntityHeader to HostVehicleData? #463 
@jruebsam FYI

`    // General information about the \c HostVehicle .
    //
    optional DetectedEntityHeader host_vehicle_header = 27;

    // Host vehicle data measured by the sensors.
    //
    optional HostVehicleData host_vehicle = 28;`

ThomasSchloemicherAVL: @pmai or should me define/add a specific header for host vehicle?

osi_hostvehicledata.proto
Krishna626: `Optional Identified id = 3` should be referred from `osi::trafficupdate` which inturn should refer from `osi3::object.MovingObject`. Advantage: Direct linkage between `trafficupdate` and `hostvehicledata` message. In that way there is a direct linkage between `hostvehicledata` coming from a traffic participant the focused in the simulation.

osi_hostvehicledata.proto
```diff
@@ -2,31 +2,369 @@ syntax = "proto2";
...
+
+        //
+        // \brief A description for the positions of the pedals.
+        //
+        message Pedalry
+        {
+            // Position of the acceleration-pedal.
+            // Range: 0-1 (Unpressed - fully pressed)
+            //
+            optional double pedal_position_acceleration = 1;
```
Krishna626, line 148: `Range: 0-1` can be reformulated as rules. 
```
    // \rules
    // is_greater_than_or_equal_to: 0
    // is_less_than_or_equal_to: 1
    // \endrules
    //
```

osi_hostvehicledata.proto
u21n80: I also agree that we need an addtional timestamp that excludes the additional latency from the system (this is the time of the actual measurement when it was measured (real data) or generated (simulation)).

osi_hostvehicledata.proto
```diff
@@ -2,31 +2,369 @@ syntax = "proto2";
...
+        //
+        // \brief Current kinematic data based on the output of the Intertial Measurement Unit (IMU).
+        //
+        message CartesianInformation
+        {
+            // Current kinematic data based on the output of the Intertial Measurement Unit (IMU).
+            //
+            // \note Note that dimension and base_polygon need not be set.
+            //
+            optional BaseMoving cartesian_data = 1;
```
u21n80, line 316: We have GeoreferencedInformation which reads it is the output of the GPS Unit. Now my question is: If I have a localization algorithm that estimates my global position. Then were do I put this information? GeoreferencedInformation is reserved for GNSS and CartesianInformation is reserved for IMU. 

Also let's assume the algorithm provides the position in ENU according to ISO8855. This means I have a cartesian  coordinate system but I also need to provide the coordinates of the coordinate system origin so that I am able to convert between ENU and geodetic position. Even if it is allowed to write this information to either IMU or GNSS. In GeoreferencedInformation I do not have a 3d orientation (at most commented out yaw). And in CartesianInformation I am missing the geodetic coordinates of the origin.

ThomasSchloemicherAVL: @u21n80 can you add your suggestions as a pull request onto this branch?
@ThomasNaderBMW would that be ok for you to add pull request onto this branch?

osi_hostvehicledata.proto
u21n80: Actually there are sensor informations in OSI also like lidar and camera. I would suggest to rather add an additional sensor of type GNSS and add this information there. Also if we use a geodetic reference to CartesianInformation like suggested in #486 then GeoreferencedInformation would not be needed anyway, as you could set the reference to you geodetic position and the cartesian position to zero which is basically the same result as only specifying the geodetic position itself.

ThomasSchloemicherAVL: Yes, i also think a _"SensorData GNSS"_ is the most plausible way.

osi_hostvehicledata.proto
ThomasNaderBMW: I agree, that it has to made clear if it is postprocessed/combined data or raw. But where does sensorics start and end? Is a rpm measurement sth for sensorData or is there the line to hostVehicleData passed?
Sth I am interested in and for the next CCB!

osi_hostvehicledata.proto
ThomasNaderBMW: Now I made it more generic.

@u21n80 : 
1. Regarding ENU: Does it help to add this as reference?
2. And also to add the startpoint x,y,z? What is the startpoint, the engine start oder simulation start
3. Do we need x,y,z travelled?
3. Why do you need a yaw value? In geodetic description?
==> Many questions, a topic to discuss @ThomasSchloemicherAVL

u21n80: @ThomasNaderBMW 
1. adding the ENU reference helps a lot. In simulation this might not be required as simulation often times (not always) uses an arbitrary cartesian coordinate system, but without the origin you are loose the information about how to convert between the cartesian (local) coordinate frame and a geodetic position. E.g. you could not compare your data against other geodetic data, e.g. reference Gnss measurements or even overlays with a map are not possible as well without knowing the origin of your coordinate system. 
2. Having a startpoint in cartesian coordinate is not equivalent to having the ENU origin see point 1. I do not know what startpoint you are refering to. If this is in regard to a reference point: It can be chosen anyway you like, but depending on what you want to achieve both could make sense. From the viewpoint of simulation choosing the origin at the start of the simulation might make sense. If you want to compare odometry estimates (which starts when the vehicle starts), then vehicle start might make more sense. The important part is that a  reference is specified, not necessarly where.
3. xyz travelled is not very meaningful I think, as when I drive forward and then backward again, my xyz travelled distance would  be zero. We could use the distance travelled instead, but for me we define a pose of the vehicle and a pose does not have a distance associated to it. If you were to store the previous poses as well (basically the path the vehicle drove) then I would agree that the distance is a property of the path. Maybe as a compromise we could a uptime field which stores the duration how long the vehicle was active and how far it drove kind of like a diagnostic field. But I do not see it as part of the pose.
4. In my pull request #488 I already removed the geodetic information (including the yaw angle). But this pulll request is not yet approved.
Hope this feedback is useful to you.

ThomasNaderBMW:
Architecture Meeting: ToDos
- Redefinition like generic board net in a sense of a restbussimulation
- Handle the situation for doubled information (in OSI-SensorData pure from the sensor(s) vs. general board net information)

osi_hostvehicledata.proto
```diff
@@ -2,31 +2,382 @@ syntax = "proto2";
...
+    // GroundTruth::proj_string, GroundTruth::MovingObject::BaseMoving::position, GroundTruth::host_vehicle_id.
+    //
+    message VehiclePositionAndKinematics
+    {
+        // ToDo: Is a boolean necessery which signal is counting?
+ 
+        //
+        // \brief Most accurate, calculated and estimated kinematic data.
+        //
+        message CartesianInformation
```
ThomasNaderBMW, line 309: @jruebsam 
Would you use this message, if released?
Otherwise I would leave it out for the beginning.

ThomasNaderBMW: As discussed: origin + relative distance in x,y is used to generate measurements from the localization function of real vehicles.

osi_hostvehicledata.proto
```diff
@@ -2,31 +2,354 @@ syntax = "proto2";
...
+    // This message contains the most accurate information the vehicle knows about its positioning and kinematics
+    // available on the board net.
+    // Because of this the values can differ from the "true" values calculated out of
+    // GroundTruth::proj_string, GroundTruth::MovingObject::BaseMoving::position, GroundTruth::host_vehicle_id.
+    //
+    message VehiclePositionAndKinematics
+    {
+        // Geodetic origin of the ENU (east-north-up) cartesian coordinate system regarding WGS84.
+        //
+        optional GeodeticPosition enu_origin = 1;
```
ThomasNaderBMW, line 303: @u21n80 : I adapted this PR as suggested (with a smaller restructuring).
Where I have problems is to find the connection to real car signals or restbus values.
But now we have a better basement for discussions, I will create a meeting.

osi_hostvehicledata.proto
```diff
@@ -2,31 +2,354 @@ syntax = "proto2";
...
+        // The host vehicle location and kinematics at \c HostVehicleData::timestamp.
+        //
+        // \note If \c HostVehicleData::enu_origin is specified, then the \c BaseMoving parent frame used by
+        // \c HostVehicleData::location is the ENU frame according to ISO8855 with the geodetic origin of
+        // \c HostVehicleData::enu_origin. Otherwise \c BaseMoving parent frame is an arbitrary cartesian coordinate system
+        // with its axis conventions following ISO8855.
+        //
+        // \note Note that dimension and base_polygon need not be set.
+        //
+        optional BaseMoving location = 2;
```
ThomasNaderBMW, line 314: @u21n80 : I am still not convinced about BaseMoving with a missing acceleration description..

jola6897: Basically I descripted the parent frame and in [BaseMoving](https://opensimulationinterface.github.io/open-simulation-interface/structosi3_1_1BaseMoving.html#a88af69f98a0768fe9d38189ea5d4fb25) you can see that the description refers to that  parent frame.
```
The relative acceleration of the moving object w.r.t. its parent frame, noted in the parent frame. The acceleration becomes global/absolute if the parent frame is inertial (all parent frames up to ground truth).
```
This should make a clear description.

ThomasNaderBMW: What I meant is the relation to board net signals, where does the acceleration come from, where are they measured in the car? The true acceleration is already in groundtruth e.g.

stefancyliax:
@ThomasNaderBMW @kmeids 
Since this wasn't discussed today, how is the chance this will be ready for merging in the next days? Otherwise we would need to shift it to 4.0.

ThomasNaderBMW:
@stefancyliax : This is the broken precessor of https://github.com/OpenSimulationInterface/open-simulation-interface/pull/547
So this PR does not need to be merged (but the 547 one).
I will close it out of that reason.

Issue #432, opened 2020-10-22
CameraSensorViewConiguration - PB Support


kmeids:
#### Describe the feature
The feature request is related to the fact that **CameraSensorViewConfiguration** fields don't support the minimum (or most generic) set of parameters needed for configuring a Physics Based (PB)  simulation environment (for camera simulation).

#### Describe the solution you would like
1. **simulation_type**, a filed that would define the simulation type. This could be selected from a predefined list (enum) e.g. DIRECTSIMULATION, INVERSESIMULATION, etc...
2.  **number_of_rays_horizontal** (to stay consistent with RadarSensorViewConfiguration)
3. **number_of_rays_vertical** (to stay consistent with RadarSensorViewConfiguration)
4.  _additional inputs from OSI users..._

#### Describe the backwards compatibility
Should not be affected as we are adding fields and not deleting already existing ones. 

#### NOTE
The requested feature is also here to act as a place holder and as a discussion starter to further collect input and opinions in this request.

Tagging @pmai @raue @PhRosenberger 

Adding the following open issue to it https://github.com/OpenSimulationInterface/open-simulation-interface/issues/389

kmeids:
Close with https://github.com/OpenSimulationInterface/open-simulation-interface/pull/446

Issue #389, opened 2020-05-15
Missing fields for ray tracer config in radar and lidar sensor view config


ClemensLinnhoff:
For ray tracer parametrization and optimization the fields
- max_range_per_ray
- max_range_per_pixel

are missing.
If they cannot be set, the ray tracing tool does not know when to terminate a ray.

kmeids:
@PhRosenberger @ClemensLinnhoff can we close this issue now?

PhRosenberger:
> @PhRosenberger @ClemensLinnhoff can we close this issue now?

I would say yes.

PR #446, opened 2020-11-23
Updated CameraSensorViewConfiguration Message

kmeids:
**Reference to related Issues:** 
- https://github.com/OpenSimulationInterface/open-simulation-interface/issues/389
- https://github.com/OpenSimulationInterface/open-simulation-interface/issues/432

**Description**
Message extension of _CameraSensorViewConfiguration_ to better support ray-tracer configuration inside the simulation environment.

**Some questions to ask**:
What is this change?
This change will introduce a new set of messages for the ray tracer configuration when dealing with camera simulation.

What does it fix?
It enables the configuration of the simulation environment by the sensor model in case of camera simulation and it improves mesasge consistency among Camera/Lidar/RadarSensorViewConfiguration. 

It is a feature, it does not break any existing functionality.

How has it been tested?
Not tested, but was discussed in the SensorModeling WG and available as an issue (FeatureRequest) on Github.

#### Mention a member
@pmai @AndreasBossert @PhRosenberger

#### Check the checklist

### DIFF
--- a/osi_common.proto
+++ b/osi_common.proto
@@ -491,3 +491,29 @@ message StatePoint
     //
     optional Orientation3d orientation = 3;
 }
+//
+// \brief Detailed WavelengthRange message.
+//
+// Defines the start (minimum) and the end (maximum) values of the wavelength.
+// Additionally, the number of samples within this range is defined in this message.
+//
+message WavelengthData
+{
+    // The start, or the minimum wavelength value.
+    //
+    // Unit: m
+    //
+    optional double start = 1;
+
+    // The start, or the minimum wavelength value.
+    //
+    // Unit: m
+    //
+    optional double end = 2;
+
+    // Number of samples to be considered within the defined wavelength range.
+    // 
+    // \note This defines the number of wavelengths to be computed during simulation, not to be confused with samples_per_pixel.
+    //
+    optional double samples_number = 3;
+}
--- a/osi_common.proto
+++ b/osi_common.proto
@@ -512,7 +512,7 @@ message WavelengthData
     optional double end = 2;
 
     // Number of samples to be considered within the defined wavelength range.
-    // 
+    // The number of samples includes the start and the end values that are defined in this message, starting from the "start" value.
     // \note This defines the number of wavelengths to be computed during simulation, not to be confused with samples_per_pixel.
     //
     optional double samples_number = 3;
--- a/osi_common.proto
+++ b/osi_common.proto
@@ -505,7 +505,7 @@ message WavelengthData
     //
     optional double start = 1;
 
-    // The start, or the minimum wavelength value.
+    // The end, or the maximum wavelength value.
     //
     // Unit: m
     //
--- a/osi_common.proto
+++ b/osi_common.proto
@@ -517,3 +517,4 @@ message WavelengthData
     //
     optional double samples_number = 3;
 }
+
--- a/osi_sensorviewconfiguration.proto
+++ b/osi_sensorviewconfiguration.proto
@@ -686,7 +686,7 @@ message CameraSensorViewConfiguration
     //
     optional uint32 number_of_pixels_vertical = 7;
 
-    // Format for image data (includes number, kind and format of channels).
+     // Format for image data (includes number, kind and format of channels).
     //
     // In the message provided by the sensor model, this field can
     // be repeated and all values are acceptable to the model, with
@@ -701,7 +701,7 @@ message CameraSensorViewConfiguration
     // cannot provide image data in one of the requested formats.
     //
     // \rules
-    // is_greater_than_or_equal_to: 1
+    // is_greater_than_or_equal_to: 1 
     // \endrules
     //
     repeated ChannelFormat channel_format = 8;
@@ -752,25 +752,124 @@ message CameraSensorViewConfiguration
         //
         CHANNEL_FORMAT_RGB_F32_LIN = 9;
 
-        // Bayer RGGB Channels UINT8 FP Linear.
+        // Bayer BGGR Channels UINT8 FP Linear.
         //
         CHANNEL_FORMAT_BAYER_BGGR_U8_LIN = 10;
 
-        // Bayer RGGB Channels UINT16 FP Linear.
+        // Bayer BGGR Channels UINT16 FP Linear.
         //
         CHANNEL_FORMAT_BAYER_BGGR_U16_LIN = 11;
 
-        // Bayer RGGB Channels UINT32 FP Linear.
+        // Bayer BGGR Channels UINT32 FP Linear.
         //
         CHANNEL_FORMAT_BAYER_BGGR_U32_LIN = 12;
 
-        // Bayer RGGB Channels Single Precision FP Linear.
+        // Bayer BGGR Channels Single Precision FP Linear.
         //
         CHANNEL_FORMAT_BAYER_BGGR_F32_LIN = 13;
+
+        // Bayer RGGB Channels UINT8 FP Linear.
+        //
+        CHANNEL_FORMAT_BAYER_RGGB_U8_LIN = 14;
+
+        // Bayer RGGB Channels UINT16 FP Linear.
+        //
+        CHANNEL_FORMAT_BAYER_RGGB_U16_LIN = 15;
+
+        // Bayer RGGB Channels UINT32 FP Linear.
+        //
+        CHANNEL_FORMAT_BAYER_RGGB_U32_LIN = 16;
+
+        // Bayer RGGB Channels Single Precision FP Linear.
+        //
+        CHANNEL_FORMAT_BAYER_RGGB_F32_LIN = 17;
+
+        // Read Clear Clear Clear Channels UINT8 FP Linear.
+        //
+        CHANNEL_FORMAT_RCCC_U8_LIN = 18;
+
+        // Read Clear Clear Clear Channels UINT16 FP Linear.
+        //
+        CHANNEL_FORMAT_RCCC_U16_LIN = 19;
+
+        // Read Clear Clear Clear Channels UINT32 FP Linear.
+        //
+        CHANNEL_FORMAT_RCCC_U32_LIN = 20;
+
+        // Read Clear Clear Clear Channels Single Precision FP Linear.
+        //
+        CHANNEL_FORMAT_RCCC_F32_LIN = 21;
+
+        // Read Clear Clear Blue Channels UINT8 FP Linear.
+        //
+        CHANNEL_FORMAT_RCCB_U8_LIN = 22;
+
+        // Read Clear Clear Blue Channels UINT16 FP Linear.
+        //
+        CHANNEL_FORMAT_RCCB_U16_LIN = 23;
+
+        // Read Clear Clear Blue Channels UINT32 FP Linear.
+        //
+        CHANNEL_FORMAT_RCCB_U32_LIN = 24;
+
+        // Read Clear Clear Blue Channels Single Precision FP Linear.
+        //
+        CHANNEL_FORMAT_RCCB_F32_LIN = 25;
+        
 
         // TBD: Further channel permutations and padding (e.g. RGBZ,
         // BGR, BAYER_RGGB/GBRG/GRBG/...), non-BAYER filters, non-linear
         // encodings, ...
+    }
+
+    // Number of samples per pixel.
+    //
+    // \note This is a characteristic of the ray tracing engine of the
+    // environment simulation, not a direct characteristic of the sensor.
+    //
+    // \rules
+    // is_greater_than_or_equal_to: 1
+    // \endrules
+    //
+    optional uint32 samples_per_pixel = 9;
+
+    // Maximum number of interactions to take into account.
+    //
+    // \note This is a characteristic of the ray tracing engine of the
+    // environment simulation, not a direct characteristic of the sensor.
+    //
+    // \rules
+    // is_greater_than_or_equal_to: 1
+    // \endrules
+    //
+    optional uint32 max_number_of_interactions = 10;
+
+    // In use-cases where a spectral ray-tracer is used, this message determines the range of the wavelength
+    // and its desired number of samples.
+    //
+    optional WavelengthData wavelength_data = 11;
+
+    // \brief Detailed WavelengthRange message.
+    //
+    // Defines the start (minimum) and the end (maximum) values of the wavelength.
+    // Additionally, the number of samples within this range is defined in this message.
+    //
+    message WavelengthData
+    {
+        // The start, or the minimum, wavelength to be considered by the spectral ray tracer.
+        //
+        // Unit: nm
+        //
+        optional uint32 start = 1;
+
+        // The end, or the maximum, wavelength to be considered by the spectral ray tracer.
+        // Unit: nm
+        //
+        optional uint32 end = 2;
+
+        // Number of samples to be considered withing the defined wavelength range.
+        // 
+        optional uint32 samples_number = 3;
     }
 
     // TBD: Optical (and other) effects to apply to image, etc.
--- a/osi_sensorviewconfiguration.proto
+++ b/osi_sensorviewconfiguration.proto
@@ -686,7 +686,7 @@ message CameraSensorViewConfiguration
     //
     optional uint32 number_of_pixels_vertical = 7;
 
-     // Format for image data (includes number, kind and format of channels).
+    // Format for image data (includes number, kind and format of channels).
     //
     // In the message provided by the sensor model, this field can
     // be repeated and all values are acceptable to the model, with
@@ -701,7 +701,7 @@ message CameraSensorViewConfiguration
     // cannot provide image data in one of the requested formats.
     //
     // \rules
-    // is_greater_than_or_equal_to: 1 
+    // is_greater_than_or_equal_to: 1
     // \endrules
     //
     repeated ChannelFormat channel_format = 8;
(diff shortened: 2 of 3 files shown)


pmai:
First a minor note on this PR: If it is targeted to 3.3.0, then it can't renumber fields, i.e. the new fields should be numbered after the existing field 8, not renumber it (this is only possible for backward-compatibility breaking major releases).

I think what also needs to be clarified is the interaction between the rays and the pixel specifications: Can/Should they be used in parallel? If not, how is the ray tracing data returned? If yes, would not a rays-per-pixel configuration make more sense/be less in danger of leading to contradictory/useless specifications? I.e. given that we already have the pixel way of returning data, any additional data paths/configurations probably need more documentation to specify exactly how they should interact.

Unless no changes to CameraSensorView itself are needed, I think related changes to the CameraSensorView should be included in this PR, to keep discussions focused.

kmeids:
Hello @pmai,
1. I will update the number of fields as you have indicated, I think we should add your comments to the contributors guide in the documentation as well. (will not commit the changes till we resolve the other topics)

2. From my point of view, rays and pixel specifications can/should be used in parallel. I was trying to reuse predefined fields but the term _rays_per_pixel_ can be used to replace the _number_of_rays_horizontal/vertical_ fields, and it actually makes more sense!

3. In either way better documentation should still be provided since the _rays_per_pixel_ field would only be used to configure the number of rays leaving the image sensor (inverse simulation), i.e. how the ray-tracer was configured. It does not indicate the number of rays that are **received** at each pixel when the simulation is done. 

4. Finally, as we have discussed in the last _Architecture Group_ meeting, we would like to obtain the _image_data_ directly as an output of the imager (CIF5, see image below), i.e. without including any image post processing-pipeline (Gamma-corrections, white-balance, demosaicing, etc..) Since we already have a ChannelFormat message with channel formats that are also suitable for imager data, my suggestion would be to add a Boolean message e.g. "raw_image_data" to indicate whether the _image_data_  is what we are obtaining from CIF5 or CIF6. What do you think?

image

pmai:
> 4. Finally, as we have discussed in the last _Architecture Group_ meeting, we would like to obtain the _image_data_ directly as an output of the imager (CIF5, see image below), i.e. without including any image post processing-pipeline (Gamma-corrections, white-balance, demosaicing, etc..) Since we already have a ChannelFormat message with channel formats that are also suitable for imager data, my suggestion would be to add a Boolean message e.g. "raw_image_data" to indicate whether the _image_data_  is what we are obtaining from CIF5 or CIF6. What do you think?

I'll set aside for now that in order for a ray tracer to produce the actual output of an imager the whole optical pipeline in front of the imager would have to be specified via sensor view configuration extensions (The sensor view producer would have to know the filter arrangement (bayer pattern, anti-aliasing filter, IR/UV filters) and the whole lens system in front of the camera).

Furthermore the whole imager configuration itself would have to be provided (i.e. well depths, geometry, A/D conversion, ...).

I find it more realistic that an abstracted format is provided, and the transformation to the actual imager output happens in the receiving model.

In any case, even if that configuration is left out and has to be manually configured, there must still be some coordination in terms of what raw data arrangement is to be provided (i.e. mapping of intensity to linear AD output levels, bit-width, etc.).

Again setting all of that aside, I think there must be additional channel format specifications that do this (if the formats like         CHANNEL_FORMAT_BAYER_BGGR_U32_LIN are not enough, because they already imply that no de-mosaicing, white balance or gamma transformation has been done).

I don't see how an additional boolean flag would help, and what it would say.

raue:
Wrt. CIF5 only: I'd consider it as a sort of 'natural' or perhaps better 'realistic' - by looking into real world camera systems - interface, which should be covered by OSI. If I remember right, some first thoughts about having CIF5 as an interface for image based models were made in the days of Pegasus, when we talked about camera effects. There, some specifics mentioned by Pierre plus others were collected - but not strictly pursued. Therefore I'd propose starting the discussion from how CIF5 should/could look like and  what it takes in terms of the stages before (i.e. CIF1 to 4) to get there.

kmeids:
> I'll set aside for now that in order for a ray tracer to produce the actual output of an imager the whole optical pipeline in front of the imager would have to be specified via sensor view configuration extensions (The sensor view producer would have to know the filter arrangement (bayer pattern, anti-aliasing filter, IR/UV filters) and the whole lens system in front of the camera).

The sensor view producer would have to know this information in either way, due to several "IP" concerns that were risen in several previous discussions, for now we are setting corresponding sensor view configuration extensions aside.

> Again setting all of that aside, I think there must be additional channel format specifications that do this (if the formats like CHANNEL_FORMAT_BAYER_BGGR_U32_LIN are not enough, because they already imply that no de-mosaicing, white balance or gamma transformation has been done).

For Bayer color filter array possesses four possible pixel arrangements (always considering  twice as many greens as reds or blues):
- BGGR (already available)
- RGGB
- GRBG
- GBRG

For Red Clear Clear Clear (RCCC) color filter array:
- RCCC
- CRCC
- CCRC
- CCCR

For Red Clear Clear Blue (RCCB) color filter array, same as the Bayer filter array however Clear instead of Green:
- BCCR
- RCCB
- CRBC
- CBRC

@raue @AndreasBossert what do you think? Do you see any other important formats?

> I don't see how an additional boolean flag would help, and what it would say.

Current documentation:
 **optional bytes osi3::CameraSensorView::image_data = 2**
_Raw image data.
The raw image data in the memory layout and order specified by the camera sensor input configuration._
How would RGB and BAYER formats correspond to this definition? I am not sure how an RGB image is considered a raw image, maybe it is only a matter of updating the documentation and explaining what is meant by "raw".

Referring to my diagram:
- **"raw_image_data = True"**; Image data at CIF5 is the **unprocessed** digital output of the imager, which (from the currently available Channel Formats) most probably can be represented as CHANNEL_FORMAT_BAYER_x_x_x, maybe also CHANNEL_FORMAT_MONO_x_x.

- **"raw_image_data = False"**; Image data at CIF6 is the **processed** digital output of the imager, which (from the currently available Channel Formats) can be represented as CHANNEL_FORMAT_RGB_x_x,  maybe also CHANNEL_FORMAT_MONO_x_x. 

This is just a suggestion from my side/how I see things considering the current documentation, please feel free to comment!

pmai:
> For Bayer color filter array possesses four possible pixel arrangements (always considering twice as many greens as reds or blues):
> 
>     * BGGR (already available)
> 
>     * RGGB
> 
>     * GRBG
> 
>     * GBRG

Yes, the current definitions wanted to be minimal to show the way to add further formats, not complete, so adding common other patterns would be the way to go, IMHO. We might need to split channel format into multiple fields if the combinatorics get to be unwieldly, but I rather suspect that in practical terms many combinations will not matter, since there tends to be some stratification in the imagers that are actually produced. If sender and receiver agree they can put any custom formats there anyway, and specify CHANNEL_FORMAT_OTHER...

> Current documentation:
> **optional bytes osi3::CameraSensorView::image_data = 2**
> _Raw image data. The raw image data in the memory layout and order specified by the camera sensor input configuration._
> How would RGB and BAYER formats correspond to this definition? I am not sure how an RGB image is considered a raw image, maybe it is only a matter of updating the documentation and explaining what is meant by "raw".

Yes, I can see where this confusion arrises: Raw here means just the image bytes (i.e. the pixels or sub-pixels), not any image format headers like PNG, TIFF, etc. This does not mean raw as in unprocessed data (since there is actually no good definition of raw in that context, see all of the "RAW" camera formats like DNG, NEF, CRW, etc.).

> 
> Referring to my diagram:
> 
>     * **"raw_image_data = True"**; Image data at CIF5 is the **unprocessed** digital output of the imager, which (from the currently available Channel Formats) most probably can be represented as CHANNEL_FORMAT_BAYER_x_x_x, maybe also CHANNEL_FORMAT_MONO_x_x.
> 
>     * **"raw_image_data = False"**; Image data at CIF6 is the **processed** digital output of the imager, which (from the currently available Channel Formats) can be represented as CHANNEL_FORMAT_RGB_x_x,  maybe also CHANNEL_FORMAT_MONO_x_x.

I think this should be encoded into the channel format, i.e. the _x_x part would show whether gamma mapping, etc. has been done.

We might want to specify in more precise detail what kind of preprocessing (that is readily appaprent at the data level, i.e. denoising that is done inside the imager itself would not be apparent) each channel format entails.

Just my view as well... ;)

kmeids:
Discussion results based on todays SM meeting:

> Yes, the current definitions wanted to be minimal to show the way to add further formats, not complete, so adding common other patterns would be the way to go, IMHO. We might need to split channel format into multiple fields if the combinatorics get to be unwieldly, but I rather suspect that in practical terms many combinations will not matter, since there tends to be some stratification in the imagers that are actually produced. If sender and receiver agree they can put any custom formats there anyway, and specify CHANNEL_FORMAT_OTHER...

We would like to evade using CHANNEL_FORMAT_OTHER when still dealing with CFA formats that are still available on the market and heavily used, this includes the RCCC and the RCCB CFA formats. 
To evade the exponential increase in the filed numbers of the channel format message, we agreed that a restructuring should be done (field names to be defined):

1. Split it into 2 Fields:
 1.1. Field 1:
  -Channel format, e.g. CHANNEL_FORMAT_BAYER_BGGR_LIN, CHANNEL_MONO, etc…
1.2. Field 2:
  -Data type and bit depth, e.g. : U8, F32, etc…
2. Split it into 3 Fields:
 2.1. Filed 1:
  -Channel format, e.g. CHANNEL_FORMAT_BAYER_BGGR_LIN, CHANNEL_MONO, etc…
 2.2. Field 2:
  -Data type, e.g. U, F
 2.3. Filed 3:
 -Data precision, e.g. 8, 16, 32

> Yes, I can see where this confusion arrises: Raw here means just the image bytes (i.e. the pixels or sub-pixels), not any image format headers like PNG, TIFF, etc. This does not mean raw as in unprocessed data (since there is actually no good definition of raw in that context, see all of the "RAW" camera formats like DNG, NEF, CRW, etc.).

Adding your clarification to the previously available "image_data" documentation will still not clarify what does the image_data actually represents... i.e. in comparison to a real-automotive camera at which interface level is "the image bytes (i.e. the pixels or sub-pixles)" provided? Or when associated with a channel format like "CHANNEL_FORMAT_BAYER_BGGR_F32_LIN " is this representing the direct output of the imager with no post-processing whatsoever or is it undergoing some post-processing steps?

"Raw" camera formats like DNG, MEF, CRW, etc... are formats that are relevant to photography and not automotive cameras. 
During todays SM meeting, we agreed on defining "Raw Image Data" as the **Most unprocessed version of the Imager data **(refer to CIF5 for visual clarification) which would also correspond to what a ray-tracer would provide as an output.

We believe that the CameraSensorView message should include both image formats on CIF5 and on CIF6, more on that later.

> I think this should be encoded into the channel format, i.e. the _x_x part would show whether gamma mapping, etc. has been done.
> We might want to specify in more precise detail what kind of preprocessing (that is readily appaprent at the data level, i.e. denoising that is done inside the imager itself would not be apparent) each channel format entails.

We will take this topic into consideration, however it will be handled by another feature/pull-request unless specific suggestions can be made.

@raue, @AndreasBossert and @jnthi will provide us some insights regarding the current interfaces available on real-automotive cameras and based on there inputs we will try to proceed.

AndreasBossert:
Additional information and a good description of the RAW problem can be found here: [https://www.imatest.com/docs/raw/](https://www.imatest.com/docs/raw/)

pmai:
> We would like to evade using CHANNEL_FORMAT_OTHER when still dealing with CFA formats that are still available on the market and heavily used, this includes the RCCC and the RCCB CFA formats.

That's what I meant. Commonly available formats would of course receive their own channel formats, however uncommon combinations that theoretically could exist but are not commonly used can still be represented as CHANNEL_FORMAT_OTHER, which will also be true for specific new imager developments, with no loss of generalisation, since those would require specific configuration of both sides anyway.

> To evade the exponential increase in the filed numbers of the channel format message, we agreed that a restructuring should be done (field names to be defined):

This is not what I would persue until there is more practical experience with the orthogonal features that people want to vary (in the context of OSI, which is about standardized interfaces, where you can trivially switch either side of the interface to another implementation), since there are potentially many dimensions to this: Byte order, Channel order, sub-pixel order, data types (not-necessarily 8bit-aligned), scan order, tone mapping, various pre-processing steps, etc. Handling all of those separately seems like overkill (and they are not necessarily pre-apparent, nor obviously necessary for actual sensor modeling at OSI level), while switching to something that does not seems little improvement over the current situation; having 500-1000 enum values would not be problematic, so seems to be enough for the short term, until more experience in standardized exchange has been gained.

The major work will be specifying the channel formats in enough detail that they are clear in what the formats entail. This can be done using orthogonal categories, even if the enums are (for now) multiplied out.

> Adding your clarification to the previously available "image_data" documentation will still not clarify what does the image_data actually represents... i.e. in comparison to a real-automotive camera at which interface level is "the image bytes (i.e. the pixels or sub-pixles)" provided? Or when associated with a channel format like "CHANNEL_FORMAT_BAYER_BGGR_F32_LIN " is this representing the direct output of the imager with no post-processing whatsoever or is it undergoing some post-processing steps?

That's why I keep saying that talking about raw is not good to do, since it effectively means nothing (the reference to DNG, NEF, CRW was not meant to imply that they themselves are relevant, but rather that they exemplify the degree to which raw has become meaningless across the picture, since imagers themselves to quite a bit of preprocessing (ignoring ASICs) nowadays, so even at that level it is meaningless to say raw, without reference to a very specific imager chip).

The actual documentation of what is or is not entailed in the pre-processing should be part of the channel format definition of each enum value. It might even turn out, that some pre-processing steps (consider on-chip denoising) are unrelated to formats, and hence should receive their own orthogonal flags. But those should be specific and not "raw on or off", and they would likely not apply to de-mosaicing and tone mapping, since those are corellated with the formats (can't have RGB without de-mosaicing, can't have non-linear without tone mapping, ...).

> During todays SM meeting, we agreed on defining "Raw Image Data" as the **Most unprocessed version of the Imager data **(refer to CIF5 for visual clarification) which would also correspond to what a ray-tracer would provide as an output.

I think talking about raw and unprocessed still seems unhelpful here: If this is supposed to be the output of an imager chip (which really is the interface to the hardware/software processing pipeline), quite a bit of signal processing has potentially gone into it already (on-chip signal conditioning, A/D conversion, various forms of noise handling, ...), most of which is not really ray-tracing related. So why not just specify it like this, explicitly stating that no de-mosaicing, tone mapping/gamma correction, etc. has been performed.

kmeids:
Thank you all for the feedback, I agree with almost everything.
More importantly with the fact that we won't be able to directly get it right from the start, thus we will proceed with iterative updates.
I will make a new commit, to incorporate our current discussions in the .proto file and use it as a bases to add our findings in our next Sensor Modeling working group meeting.

osi_sensorviewconfiguration.proto
```diff
@@ -686,7 +686,7 @@ message CameraSensorViewConfiguration
     //
     optional uint32 number_of_pixels_vertical = 7;
 
-    // Format for image data (includes number, kind and format of channels).
+     // Format for image data (includes number, kind and format of channels).
```
pmai, line 689: This seems to add a whitespace error.

osi_sensorviewconfiguration.proto
```diff
@@ -701,7 +701,7 @@ message CameraSensorViewConfiguration
     // cannot provide image data in one of the requested formats.
     //
     // \rules
-    // is_greater_than_or_equal_to: 1
+    // is_greater_than_or_equal_to: 1
```
pmai, line 704: This seems to add a whitespace error.

osi_sensorviewconfiguration.proto
```diff
@@ -752,27 +752,126 @@ message CameraSensorViewConfiguration
...
+
+        // Bayer RGGB Channels UINT32 FP Linear.
+        //
+        CHANNEL_FORMAT_BAYER_RGGB_U32_LIN = 16;
+
+        // Bayer RGGB Channels Single Precision FP Linear.
+        //
+        CHANNEL_FORMAT_BAYER_RGGB_F32_LIN = 17;
+
+        // Read Clear Clear Clear Channels UINT8 FP Linear.
```
pmai, line 787: Shouldn't all of those be red (not read) clear...?

osi_sensorviewconfiguration.proto
```diff
@@ -752,27 +752,126 @@ message CameraSensorViewConfiguration
...
+        //
+        CHANNEL_FORMAT_RCCB_F32_LIN = 25;
+        
+
         // TBD: Further channel permutations and padding (e.g. RGBZ,
         // BGR, BAYER_RGGB/GBRG/GRBG/...), non-BAYER filters, non-linear
         // encodings, ...
     }
 
+    // Number of samples per pixel.
```
pmai, line 825: Maybe this part of the changes should be split into a new PR, to keep the PRs thematically more closely related? No strong feeling, just an idea, since reviewing this part needs a deeper look, I think, than the stuff above...

osi_sensorviewconfiguration.proto
```diff
@@ -752,27 +752,126 @@ message CameraSensorViewConfiguration
...
+    // Defines the start (minimum) and the end (maximum) values of the wavelength.
+    // Additionally, the number of samples within this range is defined in this message.
+    //
+    message WavelengthData
+    {
+        // The start, or the minimum, wavelength to be considered by the spectral ray tracer.
+        //
+        // Unit: m
+        //
+        optional uint32 start = 1;
```
pmai, line 863: Integers for physical values seem problematic, i.e. what about sub-meter wavelengths? Should probably be a double, and/or it is unclear whether the unit should not be Hz...

osi_sensorviewconfiguration.proto
```diff
@@ -752,27 +752,126 @@ message CameraSensorViewConfiguration
...
+
+        // The end, or the maximum, wavelength to be considered by the spectral ray tracer.
+        // Unit: m
+        //
+        optional uint32 end = 2;
+
+        // Number of samples to be considered within the defined wavelength range.
+        // 
+        optional uint32 samples_number = 3;
```
pmai, line 872: I think this needs more explanation that this sampling is different from the per_pixel sampling value, to not cause confusion.

osi_sensorviewconfiguration.proto
```diff
@@ -752,27 +752,126 @@ message CameraSensorViewConfiguration
...
+    // \rules
+    // is_greater_than_or_equal_to: 1
+    // \endrules
+    //
+    optional uint32 max_number_of_interactions = 10;
+
+    // In use-cases where a spectral ray-tracer is used, this message determines the range of the wavelength
+    // and its desired number of samples.
+    //
+    optional WavelengthData wavelength_data = 11;
```
pmai, line 850: Might this not also be a repeated field, to allow multiple specifications, for e.g. each filter wavelength range...

pmai (changes requested):
Some slight changes are needed, I think, but general direction seems good and clear.

kmeids:
**Output from CCB meeting - 20.01.2021**
Actions:
1. @kmeids to update the file and review the changes in the SM group on Friday.

osi_sensorviewconfiguration.proto
kmeids: Whitespace removed

osi_sensorviewconfiguration.proto
kmeids: Whitespace removed

osi_sensorviewconfiguration.proto
kmeids: Changed to red

osi_sensorviewconfiguration.proto
kmeids: Updated to double.
Since we are talking about the "wavelength" and not frequency, we will stick to SI unit "m" and not Hz.

osi_sensorviewconfiguration.proto
kmeids: Updated to repeated

osi_sensorviewconfiguration.proto
kmeids: Message updated with a note to evade confusion.

osi_sensorviewconfiguration.proto
kmeids: Currently, all issues are addressed if new issues arise then we can look into splitting this PR.

osi_common.proto
```diff
@@ -491,3 +491,30 @@ message StatePoint
...
+//
+message WavelengthData
+{
+    // The start, or the minimum wavelength value.
+    //
+    // Unit: m
+    //
+    optional double start = 1;
+
+    // The start, or the minimum wavelength value.
```
pmai, line 508:
```suggestion
    // The end, or the maximum wavelength value.
```

As discussed today.

pmai (approved):

kmeids:
**Output of CCB meeting 11.02.2021**
Actions: 
1. @kmeids  update start and end description under WavelengthData
2. @pmai could you please merge this as well.

PR #466, opened 2021-01-26
set default c11 compiler for apple machines

carstendroesser:
No open issue for this pull-request.

#### Add a description
Tell cmake which compiler to use on apple machines: Building the project out of the box on apple machines will cause Cmake to fail. Setting the correct default compiler will fix this issue. Extending CMakeLists.txt fixes the build-process and enables an 'out of the box' build without errors.

#### Mention a member
@pmai

#### Check the checklist
Crossed entries were considered as not required.
- ✅ My code and comments follow the [style guidelines](https://opensimulationinterface.github.io/osi-documentation/osi/commenting.html) and [contributors guidelines](https://opensimulationinterface.github.io/osi-documentation/osi/howtocontribute.html) of this project.
- ✅ I have performed a self-review of my own code.
- ❌ I have made corresponding changes to the [documentation](https://github.com/OpenSimulationInterface/osi-documentation).
- ✅ My changes generate no new warnings.
- ❌ I have added tests that prove my fix is effective or that my feature works.
- ✅ New and existing unit tests / travis ci pass locally with my changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,9 @@ cmake_minimum_required(VERSION 3.5)
 
 project(open_simulation_interface)
 
+# set default compiler
+set(CMAKE_CXX_STANDARD 11)
+
 # Set a default build type if none was specified
 set(default_build_type "Release")
 if(EXISTS "${CMAKE_SOURCE_DIR}/.git")


CMakeLists.txt
```diff
@@ -2,6 +2,11 @@ cmake_minimum_required(VERSION 3.5)
 
 project(open_simulation_interface)
 
+# set default apple compiler
```
pmai, line 5: I think this should very likely be done unconditionally anyway, since e.g. the OSMP examples already turn this on by themselves. If someone needs to compile OSI against non-C++11 compilers, they are likely to change this as needed.

pmai (changes requested):
I think whether C++ 11 is needed or not depends more on the protobuf implementation you will use (i.e. protobuf <3.6 does not need it, protobuf >3.6 needs it), than it does on OSI itself. 

That being said, since 3.6 has been out now for more than 2.5 years, and C++11 for nearing 10, it might be a good idea to switch to C++11 by default (people who want/need to support pre-C++11 can just override the option, and supply suitable protobuf versions). The OSMP examples already require C++ 11. 

However if we do so, we should switch across all platforms (there is nothing Apple-specific about this): Could you refactor the PR to just do this unconditionally? Thank you.

CMakeLists.txt
carstendroesser: @pmai rework request processed :-)

pmai (approved):
Looking good to me, ready for CCB discussion...

kmeids:
**Output CCB meeting 03.02.2021**
Actions:
1. @pmai  to merge.

Issue #445, opened 2020-11-23
Incomplete documentation of coordinate systems


u21n80:
#### Description
The documentation of position and orientation fields does not name the frame in which the values are given in. This issue can be observed in various aspects:
a. The actual documentation of a field on the lowest level is missing the documentation. An example for this would be [DetectedMovingObject::base::position](https://opensimulationinterface.github.io/open-simulation-interface/structosi3_1_1BaseMoving.html#a189c6f025650f630b92adb277bf59db6) where it just specifies that it is the position of the bounding box center, but does not clearify in which frame this position is given in. We could specify it is given in the parent frame and specify the parent frame in a level above (in the example it would be the documentation of base), but some frame has to be written down.
b. Sometimes frames are referenced, but never defined. Example would be [DetectedMovingObject::base](https://opensimulationinterface.github.io/open-simulation-interface/structosi3_1_1DetectedMovingObject.html). Here the parent frame is specified to be the sensor's vehicle frame. This vehicle frame is never defined.

This lack of clean documentation has several implications:
1. It is unclear how hose_vehicle_location is defined for the following reasons: 
1.1. The parent frame of [host_vehicle_location is the sensor frame](https://opensimulationinterface.github.io/open-simulation-interface/structosi3_1_1SensorData.html#a7bacbcf1f521df6038d455a80baed870). If the sensor is mounted on the vehicle this would mean that the position host_vehicle_location would not change. I guess the parent frame was intended to be a global reference frame like ENU.
1.2. The reference point (see [documentation of position of BaseMoving](https://opensimulationinterface.github.io/open-simulation-interface/structosi3_1_1BaseMoving.html) and [sketch](https://opensimulationinterface.github.io/open-simulation-interface/structosi3_1_1BaseMoving.html)) is unclear. According to the field documentation the reference point is the center of the bounding box. But some figures contradict this documentation and use the center of the rear axle instead (see [DetectedLaneBoundary sketch](https://opensimulationinterface.github.io/open-simulation-interface/structosi3_1_1DetectedLaneBoundary.html), [DetectedStationaryObject sketch](https://opensimulationinterface.github.io/open-simulation-interface/structosi3_1_1DetectedStationaryObject.html)). This contradiction is also source for issue #388.
[...]

PR #486, opened 2021-02-17
Removed limitation to IMU sensor and added enu origin to allow global…

u21n80:
Added missing enu origin (see issue  #445) and allow arbitrary signals be used to fill the  Cartesian Pose instead of only IMU.

@ThomasSchloemicherAVL could you check that the changes are in line with OSI convention. This is my first contribution for which I did not setup the CI and I also did not know that I have to add the SignOff in the commit message.


u21n80:
Changes are also part of another branch which will be signed off correctly. Therefore there is no need for this pull request any more.

PR #488, opened 2021-02-26
Added ENU coordinate system specifications to HostVehicleData

u21n80:
﻿#### Related issue
Fixes point 3 of #445 regarding the missing information about the ENU coordinate system.

#### Description
- Moved vehicle location, coordinate system origin and rmse to toplevel as there is no need for an additional message. 
- Coordinate system of location and rmse has been clearly specified, if enu_origin is not specified then the old behavior where an arbitrary cartesian coordinate system is used, is restored. 
- Removed the raw Gnss sensor information as they shall be moved to a sensor specific message like it is the case for lidar measurements and images.

Signed-off-by: u21n80 <jonas.lai@avl.com>

### DIFF
--- a/osi_hostvehicledata.proto
+++ b/osi_hostvehicledata.proto
@@ -34,9 +34,7 @@ message HostVehicleData
 
     // Geodetic origin of the ENU (east-north-up) cartesian coordinate system regarding WGS84.
     //
-    // Order: Longitude[rad], latitude[rad], altitude[m].
-    //
-    optional Vector3d enu_origin = 12;
+    optional GeodeticPosition enu_origin = 12;
 
     // The host vehicle location and kinematics at \c HostVehicleData::timestamp.
     //
@@ -294,5 +292,31 @@ message HostVehicleData
             //
             optional double slip = 6;
         }
+    }
+
+    // \brief Geodetic position.
+    //
+    message GeodeticPosition
+    {
+        // Longitude in decimal degrees regarding WGS84.
+        //
+        // Unit: Degree
+        // Range: [-180; 180]
+        //
+        optional double longitude = 1;
+
+        // Latitude in decimal degrees regarding WGS84.
+        //
+        // Unit: Degree
+        // Range: [-90; 90]
+        //
+        optional double latitude = 2;
+
+        // Height above sea level regarding EGM96.
+        //
+        // Unit: m
+        // Range: [-300; 10000]
+        //
+        optional double altitude = 3;
     }
 }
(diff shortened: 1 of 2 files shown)


ThomasSchloemicherAVL:
@u21n80 can you please add a more meaningful title

u21n80:
> @u21n80 can you please add a more meaningful title of this PR

Yes of course, seemed to be the auto-generated title and I forgot to change it. Done now.

pmai:
Is there some overarching reason why this PR would renumber the fields, thus making it non-backward compatible? Especially since this now resorts back to the original fields, this change should be completely possible to release in 3.4.0 for example...

u21n80:
> Is there some overarching reason why this PR would renumber the fields, thus making it non-backward compatible? Especially since this now resorts back to the original fields, this change should be completely possible to release in 3.4.0 for example...

Hi, no sorry I was unaware of that, I just changed the fields so that the order in the code is reflected here. I will change it back.

ThomasNaderBMW:
@u21n80 : I transferred your points into the PR #441. What I left out is the definition on the top level.
For me it makes the HostVehicleData better to be understood as a restbussimulation (all information is on the same level and not taken out of the submessages). We can discuss this in the next meeting from @ThomasSchloemicherAVL

jola6897:
@ThomasNaderBMW I made the type of the ENU origin now a message with explicit mention of longitude, latitude and altitude. The documentation is the same as your GeoreferencedInformation but I renamed it to GeodeticPosition.

Also I intend to reuse the location field (BaseMoving) to convey the information of position relative to the enu origin, but here also acceleration and velocity can be given. The coordinate frames of all values inside BaseMoving should be defined by the fact that I specify the parent frame of BaseMoving.

ThomasNaderBMW:
> @ThomasNaderBMW I made the type of the ENU origin now a message with explicit mention of longitude, latitude and altitude. The documentation is the same as your GeoreferencedInformation but I renamed it to GeodeticPosition.
> 
> Also I intend to reuse the location field (BaseMoving) to convey the information of position relative to the enu origin, but here also acceleration and velocity can be given. The coordinate frames of all values inside BaseMoving should be defined by the fact that I specify the parent frame of BaseMoving.
> 
> @ThomasNaderBMW I would also recommend to merge this pull request instead of manually copying the text to the other branch to maintain traceability.

I updated my PR regarding your suggestions with smaller restructuring. I had to do it manually because of merge conflicts here, I think a rebase would have been necessary here.

clemenshabedank:
CCB 09/05/2022: We close this because we think it was fixed by #547 . In case this is incorrect please object @u21n80 .

PR #498, opened 2021-03-15
WIP: Introduce PedestrianClassification

clemenshabedank:
Discussion in WP11.

#### Add a description
Added pedestrian classification to have head pose and upper body pose in sensorview and sensordata. So far it exists only in SensorData so that it is unclear where sensor models should get the information from. 
Should be currently seen as a discussion point in WP11 -> WIP

**Some questions to ask**:
What is this change?
What does it fix?
How has it been tested?

- [ ] My suggestion follows the [style and contributors guidelines](https://opensimulationinterface.github.io/osi-documentation/open-simulation-interface/doc/howtocontribute.html).
- [ ] I have taken care about the [documentation](https://opensimulationinterface.github.io/osi-documentation/open-simulation-interface/doc/commenting.html).
- [ ] I have done the [DCO signoff](https://opensimulationinterface.github.io/osi-documentation/open-simulation-interface/doc/howtocontribute.html#developer-certification-of-origin-dco).

### DIFF
--- a/osi_object.proto
+++ b/osi_object.proto
@@ -359,6 +359,17 @@ message MovingObject
     //
     optional MovingObjectClassification moving_object_classification = 9;    
 
+    // Specific information about the classification of the pedestrian.
+    //
+    // \note This field is mandatory if the \c #type is
+    // #TYPE_PEDESTRIAN .
+    //
+    // \rules
+    // check_if this.type is_equal_to 3 else do_check is_set
+    // \endrules
+    //
+    optional PedestrianClassification pedestrian_classification = 10;    
+
     // Definition of object types.
     //
     enum Type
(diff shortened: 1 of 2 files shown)


dbeckerAC:
Does this imply that the `BaseMoving` orientation value of the moving object describes the pedestrian's lower body pose? And if that's the case, do we need to state that somewhere?

clemenshabedank:
Good question, we should discuss it. What about defining it in the direction of prior movement? Lower body pose I find a bit problematic, because hard to define and maybe not so meaningful. I think we should also be able to assist usecases that just model the pedestrian as a moving bounding box, which might prefer the direction-of-prior-movement option. What do you think?

osi_object.proto
```diff
@@ -826,4 +837,47 @@ message MovingObject
...
     }
+
+    //
+    // \brief Information for the classification of pedestrians regarding
+    // \c MovingObject (host or other).
+    //
+    message PedestrianClassification
+    {
+        // The head pose describes the pedestrian's head orientation. In ground truth / traffic update it is relative to
+        // the global ground truth frame. In sensordata (detected pedestrians) it is relative to the host vehicle frame.
```
clemenshabedank, line 848: Comment Kmeid: Host vehicle frame correct or sensor frame?

clemenshabedank: @kmeids the description for detected pedestrians currently is:
```
Describes the head orientation w.r.t. the host vehicle orientation.
```
Therefore, I wrote "host vehicle frame" to say the same thing and be backwards compatible. Nevertheless, I think sensor frame is better (i.e. physical mounting position / mounting position, depending on sensor technology). I guess we can make this change only in a major OSI version.

stefancyliax:
OSI CCB:
- Another possible topic for 3.4
- @clemenshabedank please look at CI. Broken for now.

clemenshabedank:
Despite originally planned, I think it is better to plan this for v4.0. Clarification on a pedestrian's bounding box does not seem trivial for different use cases. Merge discussions with https://github.com/OpenSimulationInterface/open-simulation-interface/issues/526 going forward.

kmeids:
@stefancyliax. @clemenshabedank  based on https://github.com/OpenSimulationInterface/open-simulation-interface/pull/498#issuecomment-872100764 it is meant to be handled in 4.0, is this still the case?

stefancyliax:
You are correct. This is supposed to be in 4.0

clemenshabedank:
Just so that it won't be lost, here are some brainstorming slides [20210611_ASAM_OSI_Pedestrians.pptx](https://github.com/OpenSimulationInterface/open-simulation-interface/files/8714992/20210611_ASAM_OSI_Pedestrians.pptx) about the pedestrian discussions we had in 06/2021 and before. The discussions were about potential changes going beyond what is in this PR.

Issue #530, opened 2021-06-17
Antecessor and successor definition for lanes 


FabianPfeufferCAP:
#### Describe the problem
I'm currently working with OSI lanes and need some clarification on a part of the standard, that is unclear to me. 
It seems that the OSI standard doesn't clarify how antecessor_lane_id and successor_lane_id have to be set for non-driving lanes. 

#### Describe your research

From my understanding, for driving lanes, the antecessor_lane_id denotes the lane(s) at the beginning (first centerline point), the successor_lane_id the lane(s) at the end (last centerline point) of a given OSI lane. If the "centerline_is_driving_direction" attribute is false, the opposite is the case, i.e. the antecessor is connected to the OSI lane's last and the successor to its first centerline point.

I check the documentation for OSI lane pairing and OSI lanes in general, but couldn't find any information regarding this topic. However, the standard explicitly says that non-driving lanes should not have a centerline ( https://opensimulationinterface.github.io/open-simulation-interface/structosi3_1_1Lane_1_1Classification.html ) :

"... non-driving lanes do not have a center line. A vehicle must calculate this individually and depending on the situation."

#### Ask your question
So with neither the centerline, nor the centerline_is_driving_direction, how should the antecessor_lane_id and successor_lane_id be determined?
Does the lane boundary define the orientation for non-driving lanes, i.e. is the antecessor connected to the first and the successor connected to the last lane boundary point?

HendrikAmelunxen:
Please see [https://github.com/OpenSimulationInterface/open-simulation-interface/issues/411](Issue 411) and [https://github.com/OpenSimulationInterface/open-simulation-interface/pull/460](PR 460)

Defining the left/right/antecessor/succesor lanes with regard to the intended driving direction was problematic/inconsistent for different reasons. That is why this has been changed for  OSI 3.3.

clemenshabedank:
Closed. Arguments see https://github.com/OpenSimulationInterface/open-simulation-interface/pull/542.

clemenshabedank:
Closed. Arguments see https://github.com/OpenSimulationInterface/open-simulation-interface/pull/542.

clemenshabedank:
Closed. Arguments see https://github.com/OpenSimulationInterface/open-simulation-interface/pull/542.

PR #542, opened 2021-06-25
Improve lane pairing documentation

FabianPfeufferCAP:

Antecessor and successor definition for lanes #530 

#### Add a description

* Added an extended version of the image OSI_LaneBoundaries_And_CenterLine.svg
  that includes, connected by a junction.
  - The new image is called
    OSI_LaneBoundaries_And_CenterLine_Intersection.svg and will replace
    the old one in the Lane/Classification/Centerline section.
  - The old image will still be used in the LaneBoundary/BoundaryPoint
    section.

* Added a more precise description to Lane/Classification/lane_pairing
  on how antecessor and successor have to be set.

**Some questions to ask**:

What is this change?

- Improvement for OSI documentation

What does it fix?

- Documentation for OSI lane pairing

### DIFF
--- a/doc/images/OSI_LaneBoundaries_And_CenterLine_Intersection.svg
+++ b/doc/images/OSI_LaneBoundaries_And_CenterLine_Intersection.svg
@@ -1,0 +1,2212 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="283.11191mm"
+   height="166.65927mm"
+   viewBox="0 0 283.11191 166.65927"
+   version="1.1"
+   id="svg1002"
+   inkscape:version="0.92.3 (2405546, 2018-03-11)"
+   sodipodi:docname="OSI_LaneBoundaries_And_CenterLine_Intersection.svg">
+  <defs
+     id="defs996" />
+  <sodipodi:namedview
+     id="base"
+     pagecolor="#002000"
+     bordercolor="#000000"
+     borderopacity="0.90196078"
+     inkscape:pageopacity="0"
+     inkscape:pageshadow="2"
+     inkscape:zoom="0.70000001"
+     inkscape:cx="347.94309"
+     inkscape:cy="244.37872"
+     inkscape:document-units="mm"
+     inkscape:current-layer="layer1"
+     showgrid="true"
+     inkscape:snap-bbox="true"
+     inkscape:bbox-paths="true"
+     inkscape:bbox-nodes="true"
+     inkscape:snap-bbox-edge-midpoints="true"
+     inkscape:snap-bbox-midpoints="true"
+     showguides="false"
+     inkscape:snap-grids="false"
+     inkscape:guide-bbox="true"
+     inkscape:lockguides="false"
+     inkscape:snap-midpoints="true"
+     inkscape:snap-smooth-nodes="true"
+     inkscape:snap-intersection-paths="true"
+     inkscape:object-paths="true"
+     inkscape:snap-object-midpoints="true"
+     inkscape:snap-center="true"
+     inkscape:window-width="1920"
+     inkscape:window-height="1052"
+     inkscape:window-x="0"
+     inkscape:window-y="0"
+     inkscape:window-maximized="1"
+     units="mm"
+     height="225mm"
+     fit-margin-top="0"
+     fit-margin-left="0"
+     fit-margin-right="0"
+     fit-margin-bottom="0">
+    <inkscape:grid
+       type="xygrid"
+       id="grid1547"
+       originx="-10.066611"
+       originy="-20.516049" />
+    <sodipodi:guide
+       position="45.698397,215.36896"
+       orientation="1,0"
+       id="guide4761"
+       inkscape:locked="false" />
+    <sodipodi:guide
+       position="51.522568,215.36896"
+       orientation="1,0"
+       id="guide4763"
+       inkscape:locked="false" />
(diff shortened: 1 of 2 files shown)


HendrikAmelunxen:
This topic has been discussed at the harmonisation working group meeting. Some points are not clear, e.g. does it make sense to have such a strict rule regarding the overlapping of start/end-points of centerlines and laneboundaries? How to handle road-road-connections? etc. 
The attached ppt shows the current status of discussion to be further discussed.
[Lane_Pairings_Discussion.pptx](https://github.com/OpenSimulationInterface/open-simulation-interface/files/6734661/Lane_Pairings_Discussion.pptx)

FabianPfeufferCAP:
Thanks for the update, I looked at the ppt and it seems fine for roads, but I see an issue with non-driving roads here.

Let's say a Pedestrian is walking from one sidewalk onto another one (both connected to each other).
The pedestrian walks along the first sidewalk in e.g. ascending lane boundary order. When reaching the second sidewalk, the pedestrian has to determine the new order of the lane boundary order. To do this, the pedestrian has to check the antecessor/successor of the second sidewalk and see, if one of the antecessors/successors matches the first sidewalk. This is a bit inconvenient, but should be fine.
However, in the roundabout case shown in the ppt, this approach is no longer possible. The first sidwalk will appear as antecessor and successor of the second one, so the result will be arbitrary.

How should the definition order for non-driving roads be determined in this case?

FabianPfeufferCAP:
@HendrikAmelunxen 
I have an additional question regarding intersections and sidewalks.
The picture below shows the intersection used in the OSI-Lane documentation ( https://opensimulationinterface.github.io/open-simulation-interface/structosi3_1_1Lane.html ).

image

There are tree sidewalks (marked in red) that lead across this intersection. Are these sidewalks part of the intersection itself, i.e. do they have the TYPE_INTERSECTION type and SUBTYPE_SIDEWALK subtype?

If this is indeed the case, I have another question with regards to the successor/antecessor pairing, proposed in the ppt. The ppt states that the successor/antecessor pairing of the intersection should be done using the "centerline_is_driving_direction" attribute of the connected roads. This is fine for driving lanes, but how should the successor/antecessor pairing be done for sidewalks?

HendrikAmelunxen:
@FabianPfeufferCAP 
Regarding closed connection of two sidewalks: Yes, you are right. This situation is undefined for non-driving lanes. In this specific situation a third, intermediate lane would solve the situation.

Regarding the sidewalks at intersections: Since sidewalks have a left and right boundary I would not assign them to an intersection.

In general, the current roadmodel does not fit very well to the usecase of navigating along non-driving lanes. Therefore this is also a usecase which is addressed in the [road model 2.0 project](https://github.com/OpenSimulationInterface/open-simulation-interface/projects/11). Feel free to contribute there as well.

ThomasNaderBMW:
@HendrikAmelunxen and @clemenshabedank , we should not forget that topic here. Does a discussion in the Road Model Group make sense?

clemenshabedank:
@ThomasNaderBMW yes we can discuss it in the Road Model Group. Should I invite Fabian for that?

clemenshabedank:
Might be handled through https://github.com/OpenSimulationInterface/open-simulation-interface/pull/599

clemenshabedank:
Let's try to make a decision what to do with this quite old PR considering the upcoming v3.5 release! Here is my opinion: 

- syntactically, concatenating 2 identifiers to one identifier as suggested (identifier is of type uint64, the symbols with letters are just for illustrative purpose I think) does not look like a solution to me, as the identifier would not be splittable unambiguously from the receiving side. Besides this syntax issue I think it is more than just a documentation change leading to much effort on the receiving side.
- since we currently still have the centerline_is_driving_direction flag, the antecessor and successor with respect to the **real** driving direction can be unambiguously determined by a combination of this flag and the storage order of centerline points, as far as I can tell. So in your l2 -> l3 example, l2 would be the antecessor (points going away from the intersection and driving direction false) and l3 would be the successor (points going away from the intersection and driving direction true).
- the in the conversation section discussed issue with not connected non-driving lanes is not solved with this PR and would need to be a separate one. The already merged PR599 seems like a proper solution for those use cases. In my opinion the highlighted lanes in the picture in https://github.com/OpenSimulationInterface/open-simulation-interface/pull/542#issuecomment-884751920 shall be modelled as non-driving lanes not being part of the intersection. 

For those reasons I would not like to merge it as-is and rather close it. In case logical connections for non-driving lanes are needed and not already covered by the logical lane approach  I suggest to open a new PR. 

Is my argumentation correct and would this be fine for you @FabianPfeufferCAP ?

clemenshabedank:
Let's try to make a decision what to do with this quite old PR considering the upcoming v3.5 release! Here is my opinion: 

- syntactically, concatenating 2 identifiers to one identifier as suggested (identifier is of type uint64, the symbols with letters are just for illustrative purpose I think) does not look like a solution to me, as the identifier would not be splittable unambiguously from the receiving side. Besides this syntax issue I think it is more than just a documentation change leading to much effort on the receiving side.
- since we currently still have the centerline_is_driving_direction flag, the antecessor and successor with respect to the **real** driving direction can be unambiguously determined by a combination of this flag and the storage order of centerline points, as far as I can tell. So in your l2 -> l3 example, l2 would be the antecessor (points going away from the intersection and driving direction false) and l3 would be the successor (points going away from the intersection and driving direction true).
- the in the conversation section discussed issue with not connected non-driving lanes is not solved with this PR and would need to be a separate one. The already merged PR599 seems like a proper solution for those use cases. In my opinion the highlighted lanes in the picture in https://github.com/OpenSimulationInterface/open-simulation-interface/pull/542#issuecomment-884751920 shall be modelled as non-driving lanes not being part of the intersection. 

For those reasons I would not like to merge it as-is and rather close it. In case logical connections for non-driving lanes are needed and not already covered by the logical lane approach  I suggest to open a new PR. 

Is my argumentation correct and would this be fine for you @FabianPfeufferCAP ?

FabianPfeufferCAP:
@clemenshabedank 

From my point of view, we can close this issue as the logical lanes should solve this problem for my use case.
However, I have a some doubts about the combination of "centerline_is_driving_direction" + point order to determine if a connected lane is a successor or an antecessor.

For a human looking at it, it's pretty straightforward to determine if points go away from an intersection, but turning this statement into code can be quite tricky.
Let's say we use the distance of the centerline points from the intersection as our metric to determine "going away". Here are a couple of remarks/questions:

1. Where is the reference point on the intersection to determine the distance?
2. Let's assume we found a suitable reference point. How many centerline points have to be checked until "going away" is true? It's pretty easy to create an example where half the centerline points lead away and the other half towards the intersection. Do you only need to check the first two?
3. Will the standard define how  "going away" should be determined?
4. Using distance checks to determine the intersection's connection orientation seems quite expensive (not necessarily an issue if done during initialization time)

PR #546, opened 2021-06-30
Added Timestamp for each Moving Object.

ThomasNaderBMW:
Signed-off-by: Nader Thomas <thomas.nader@bmw.de>

#### Add a description
The simulation timestamp the calculated "base" parameters (especially "position") of the moving object apply to.
Background: The timestamps can differ between e.g. host vehicle, rest of the vehicle traffic and pedestrians (or other types), because they may be calculated in different modules. And even inside those groups the timestamps can differ if one part is calculated in one cycle and the rest in the following cycle out of performance reasons.
Usage: E.g. for extrapolation of the whole traffic in graphic engines.

**Some questions to ask**:
What is this change? What does it fix?
The groundTruth timestamp is too unspecific and not working for that use case.
No.
How has it been tested?
Yes and working. Tested by Christopher Kiwus.

- [ ] I have taken care about the [documentation](https://opensimulationinterface.github.io/osi-documentation/open-simulation-interface/doc/commenting.html).

### DIFF
--- a/osi_object.proto
+++ b/osi_object.proto
@@ -422,6 +422,28 @@ message MovingObject
     //
     repeated ExternalReference source_reference = 10;
 
+    // The simulation timestamp the calculated "base" parameters (especially
+    // "position") of the moving object apply to.
+    // Background: The timestamps can differ between e.g. host vehicle, rest of
+    // the vehicle traffic and pedestrians (or other types), because they may be
+    // calculated in different modules. And even inside those groups the
+    // timestamps can differ if one part is calculated in one cycle and the rest
+    // in the following cycle out of performance reasons.
+    // Usage: E.g. for extrapolation of the whole traffic in graphic engines.
+    //
+    // Regarding the "simulation timestamp":
+    // The zero time point is arbitrary but must be identical for all messages.
+    // Recommendation: Zero time point for start point of the simulation.
+    //
+    // \note Zero time point does not need to coincide with the UNIX epoch.
+    //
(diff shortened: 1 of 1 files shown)


stefancyliax:
CCB 13.07.2021:
Should be talk about from an architectural standpoint. (@pmai)

stefancyliax:
CCB Meeting 04.08.2021:
- topic will be in clarification at BMW. Decision will be done on CCB 18.08.2021.

ThomasNaderBMW:
BMW internal discussion is done. We came to the conclusion that:

- it is not state of the art that all ground truth data is calculated exactly at the point of time the ground truth timestamp suggests. The modules calculation the ego vehicle, pedestrians, bicycles and rest of traffic mostly have different cycle times and can be asynchronous
- an extrapolation doesn't fix this as it is just an estimation how the vehicles _could_ move and not the reality
 ==> So the ground truth timestamp is already now not showing the real data. For most applications it may still be accurate enough, but for the graphics engine use cases it is not. And that is why we need an additional timestamp of the calculation time.

Please give feedback @kmeids , @pmai , @stefancyliax 
Having this in 3.4 would make things easier for us!

stefancyliax:
OSI CCB:
- Change would be major impact in simulation quality. It can not be merged at this point. Topic is candidate for removal from 3.4.
- Change would be breaking change that would only be possible in major release.
- Overall timing and synchronization topics in OSI should be topic for 4.0 release.

ThomasNaderBMW:
Won't be done atm.

Issue #512, opened 2021-04-22
Centralize and unify color codings


HendrikAmelunxen:
#### Describe the feature
Currently the color enumerations are distributed ("osi_roadmarking.proto", "osi_object.proto", "osi_lane.proto", "osi_featuredata.proto") and not aligned; e.g. COLOR_VIOLET = 5 (osi_object.proto) vs. COLOR_VIOLET = 8 (osi_lane.proto).

#### Describe the solution you would like
It would make sense to have a central definition of the color codings, e.g. in osi_common.proto.

#### Describe the backwards compatibility
Since the different codings are not aligned, the centralization and alignment would break backward compatibility; colors will be interpreted different.

MarcelOliver:
At the moment, color is defined in four different files (osi_lane, osi_object, osi_featuredata and osi_roadmarking) as shown here:

osi_lane.proto:

    COLOR_UNKNOWN = 0;
    COLOR_OTHER = 1;
    COLOR_NONE = 2;
    COLOR_WHITE = 3;
    COLOR_YELLOW = 4;
    COLOR_RED = 5;
    COLOR_BLUE = 6;
    COLOR_GREEN = 7;
    COLOR_VIOLET = 8;
    COLOR_ORANGE = 9;

osi_object.proto:

    COLOR_UNKNOWN = 0;
    COLOR_OTHER = 1;
    COLOR_YELLOW = 2;
    COLOR_GREEN = 3;
    COLOR_BLUE = 4;
    COLOR_VIOLET = 5;
    COLOR_RED = 6;
    COLOR_ORANGE = 7;
    COLOR_BLACK = 8;
    COLOR_GREY = 9;
    COLOR_WHITE = 10;

osi_featuredata.proto:

    COLOR_UNKNOWN = 0;
    COLOR_OTHER = 1;
    COLOR_BLACK = 2;
    COLOR_GREY = 3;
    COLOR_WHITE = 4;
    COLOR_YELLOW = 5;
    COLOR_ORANGE = 6;
    COLOR_RED = 7;
    COLOR_VIOLET = 8;
    COLOR_BLUE = 9;
    COLOR_GREEN = 10;
    COLOR_REFLECTIVE = 11;

osi_roadmarking.proto:

    COLOR_UNKNOWN = 0;
    COLOR_OTHER = 1;
    COLOR_WHITE = 2;
    COLOR_YELLOW = 3;
    COLOR_BLUE = 5;
[...]

PR #554, opened 2021-07-27
Introducing new color models

thomassedlmayer:
Signed-off-by: Thomas Sedlmayer <tsedlmayer@pmsfit.de>

#### Description
This PR emerged from the Harmonisation group and the discussions related to https://github.com/OpenSimulationInterface/open-simulation-interface/issues/512. It should serve as a basis for further discussion. The color models (Grey, RGB, RGBIR, HSV, LUV) were adopted from the ISO-23150. 

It is not completely clear yet where and how to apply the new message and how to manage the transition to complement or replace existing color enums. As a result from discussions, color enums that represent **semantic meaning** (e.g. traffic light colors, lane marking colors, lane boundary colors) should probably be kept as is or could be unified as separate enums (e.g. "SemanticColors"). Semantic colors are colors that specifically specify an objects meaning in traffic and do not necessarily represent its visual color. E.g. a lane boundary that is semantically classified as orange does not have to be visually orange (e.g. washed out to white). Hence, we should clearly differentiate semantic colors from visual colors that represent appearance only.

Color enums for stationary objects and camera detections could be complemented (short-term) or replaced (long-term) with the new ColorDescription. Also, moving objects do not have a color field yet and could also contain the new ColorDescription.
The name of the message (ColorDescription) is still under discussion.

### DIFF
--- a/osi_common.proto
+++ b/osi_common.proto
@@ -728,7 +728,8 @@ message ColorHSV
 {
     // Hue
     //
-    // Range: [0,1]
+    // Unit: deg
+    // Range: [0,360[
     //
     optional double hue = 1;
 
--- a/osi_common.proto
+++ b/osi_common.proto
@@ -551,6 +551,7 @@ message StatePoint
     //
     optional Orientation3d orientation = 3;
 }
+
 //
 // \brief Detailed WavelengthRange message.
 //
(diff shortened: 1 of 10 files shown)


HendrikAmelunxen:
I think h (in hsv) has a value from 0 to 360; and v (in hsv) might also has a value from -1 to 1.
Since cmyk is just used for print media I am not sure if we should support it here.

stefancyliax:
OSI CCB:
- please do documentation review @pasched

pasched (approved):

Issue #583, opened 2021-09-29
Inconsistency and confusion over large vehicle classification


caspar-ai:
#### Describe the feature
There are a few related concerns:

- there are inconsistent vehicle categories between OpenScenario 1.x and OSI
- some of the vehicle categories are poorly defined and do not match with standard naming
- comments could do with clarification to aid consistency

#### Additional context and links
OSI defines a `TYPE_HEAVY_TRUCK` but no `TYPE_TRUCK`, see [here](https://opensimulationinterface.github.io/osi-documentation/open-simulation-interface/structosi3_1_1MovingObject_1_1VehicleClassification.html#exhale-struct-structosi3-1-1movingobject-1-1vehicleclassification).  Whereas OpenScenario defines a type `truck` but no `heavytruck`, see [here](https://www.asam.net/static_downloads/modelDocumentation/content/VehicleCategory.html).

Also worth noting is that:

- OpenLABEL draft only recognises a single truck "A large, heavy motor vehicle used for carrying goods and materials", draft version of that spec can be found [here](https://www.asam.net/active-projects/public-review/asam-openlabel-v100/download-asam-openlabel-v100/), in the "Behaviour tags" document
    - note this document is still under review
    - it already has a comment indicating it should be consistent with OpenScenario and OSI :)
- UK regulations (COBA) distinguish between "OGV1" and "OGV2" vehicles which could be mapped to "truck" and "heavy truck", see [here](https://democracy.york.gov.uk/mgConvert2PDF.aspx?ID=38557)
    - although under UK definitions a vehicle can change from being OGV1 to OGV2 depending on whether it currently has a trailer, such changing of a classification during a simulation may prove undesirable
- US regulations (FHWA) have many classes which are very specific in terms of the number of axles in both tractor and trailer, see [here](https://www.fhwa.dot.gov/publications/research/infrastructure/pavements/ltpp/13091/002.cfm)
    - such precise categorisation may also be undesirable in an ADS perception context
- UNECE only recognises a single category of truck ("designed and constructed for the carriage of goods"), see [here](https://unece.org/classification-and-definition-vehicles)

[...]

Issue #500, opened 2021-03-23
Best-Practice Guide


DerBaertige:
#### Describe the feature
The standard should and cannot define everything, which can occur during a real world implementation. In many projects, in addition to the actual standard, the user is given additional information that is intended to simplify use. During the discussion of the Pull-Request #465 the idea of "Best-Practice Guide" growth, which describes different good/tested solutions, which cannot/should not be added to the standard directly.

#### Describe the solution you would like
This issue should be the start a collection of the usage of OSI, which facilitate the usage of different parts and the interaction with different standards, not restricted to ASAM standards.

#### Describe the backwards compatibility
There should be no issues regarding to backwards compatibility, as it is an additional extension to the documentation.

The first idea was, that the guide should be part of the [OSI-Documentation](https://github.com/OpenSimulationInterface/osi-documentation), where the different parts of OSI and its tools are combined.

##### Mapping of OpenDrive Roads and OSI Lanes
Based on both standards, a transfer of the road network to each other has to implemented. As OpenDrive and the OSI usees different paradigms to represent the network, a direct assignment is not possible. For simplicity and traceability a direct mapping of the `t_road` &#8594; `t_road_lanes_laneSection` &#8594; `t_road_lanes_laneSection` segments to the `message Lane` should be performed.

HendrikAmelunxen:
**Laneboundary Height and Width**
In [](https://github.com/OpenSimulationInterface/open-simulation-interface/pull/436) examples are given, how laneboundaries with height and width should be described. This could be useful in a best practice guide.

clemenshabedank:
Explain usage of DASH_GAP if https://github.com/OpenSimulationInterface/open-simulation-interface/pull/573 gets merged.

PR #585, opened 2021-09-30
Add TYPE_TRUCK to vehicle classifications

caspar-ai:
To align with OpenSCENARIO and OpenLABEL.  Also deprecate the use of TYPE_HEAVY_TRUCK and TYPE_SEMITRAILER as ambiguous.

For more details see #583.


caspar-ai:
Not sure who is best placed to review this.

@adrianschultz as you recently added an issue about vehicle classification I'm starting with you :)  please feel free to reassign, or add people, if you think they are better placed or would be interested.

Also, I realise I've raised a few harmonisation issues recently, if it would be helpful for me to come along to a harmonisation working group meeting do drop me an invite.

Thanks!

caspar-ai:
Adding the `ReadyForCCBReview` as a means for getting some eyes on this and how to progress it forwards.

stefancyliax:
OSI CCB 10.11.2021:
- After discussion about classification of trucks and trailers, we would propose another solution. 
- Proposal:
**SemiTrailer** -> improve comment, type describes only trailer without truck
**SemiTractor** -> new type to clarify pulling part of semi-truck
**AgricaturalVehicle** -> new type for farm vehicles
**HeavyTruck** -> should be kept, don't add new type TRUCK

Please also add info about mapping to OpenSCENARIO. 
Please also evaluate which types would benefit from a example picture. (e.g. **Deliveryvan**, **Semitrailer**, **Semitractor**, **HeavyTruck**, ...)

caspar-ai:
I've done as suggested and added semi-tractor, rather than removing semi-trailer.

I've also tried to write down the mapping for each type to its OSC 1.x and 2.x counterparts.  In particular, it is worth noting that OSC 2.x (and also OpenLABEL I believe) while both still being defined, they seem to be going the way of a few broad types rather than more narrower types.

It sounds like consensus is that there is value in OSI have a richer set of vehicle types than OpenScenario / OpenLabel.  I would strongly encourage that reasoning to be captured within the OSI definitions as a record for future users and a more permanent record of these discussion.  I'm not sure what the reasoning is, so I've left a placeholder for now.

@HendrikAmelunxen is that something you can provide based on the last CCB discussion?

osi_object.proto
```diff
@@ -745,66 +748,133 @@ message MovingObject
...
+            //
+            // Definition: The cab for an articulated vehicle.
+            //
+            // This has no equivalent in OpenScenario 1.x.
+            //
+            // This is one of the OSI types which map from "truck" in OpenScenario 2.x.
+            //
+            TYPE_SEMITRACTOR = 16;
+
+            // This vehicle is a trailer unit for an articulated vehicle.
```
HendrikAmelunxen, line 809: Suggestion:
This vehicle is a semi-trailer that can be pulled by a semi-tractor.

Definition: A semi-trailer is a trailer without a front axle. The front part of the semi-trailer is mounted on the semi-tractor which leads to a significant overlap of the bounding-boxes of semi-tractor and semi-trailer.

osi_object.proto
```diff
@@ -745,66 +748,133 @@ message MovingObject
...
+            // This corresponds to "van" in OpenScenario 1.x.
+            //
+            // This is one of the OSI types which map from "truck" in OpenScenario 2.x.
+            //
             TYPE_DELIVERY_VAN = 6;
 
-            // Vehicle is a heavy truck.
+            // Vehicle is a (heavy) truck.
+            //
+            // Definition: A rigid-body truck that cannot pull a semi-trailer.
```
HendrikAmelunxen, line 791: Suggestion:
"Definition: A rigid-body truck."
I would leave out the addition regarding the semi-trailer since this is also valid for each other type, except the semi-tractor.

osi_object.proto
```diff
@@ -745,66 +748,133 @@ message MovingObject
...
+            //
+            // This is one of the OSI types which map from "bus" in OpenScenario 2.x.
+            //
             TYPE_BUS = 12;
 
             // Vehicle is a tram.
             //
+            // This corresponds to "tram" in OpenScenario 1.x.
+            //
+            // This is one of the OSI types which map from "bus" in OpenScenario 2.x.
```
clemenshabedank, line 860: I am not sure OpenSCENARIO 2.0 intends to describe a tram with bus. Did you talk with the Domain Model people there? While things are still under development there, they currently have an `intended_infrastructure` enum with the value `tram `while (you are right here) no value in `vehicle_category` for tram. Unless clarified I think I would delete this line here.

osi_object.proto
```diff
@@ -745,66 +748,133 @@ message MovingObject
...
+            //
             TYPE_TRAM = 13;
 
             // Vehicle is a train.
             //
+            // This corresponds to "train" in OpenScenario 1.x.
+            //
+            // This is one of the OSI types which map from "bus" in OpenScenario 2.x.
```
clemenshabedank, line 868: I am not sure OpenSCENARIO 2.0 intends to describe a tram with bus. Did you talk with the Domain Model people there? While things are still under development there, they currently have an `intended_infrastructure` enum with the value `rail`while (you are right here) no value in `vehicle_category` for train. Unless clarified I think I would delete this line here.

osi_object.proto
```diff
@@ -745,66 +748,133 @@ message MovingObject
...
+            //
+            // This is one of the OSI types which map from "VRU_vehicle" in OpenScenario 2.x.
+            //
             TYPE_MOTORBIKE = 10;
 
             // Vehicle is a bicycle (without motor and specific lights).
             //
+            // This corresponds to "bicycle" in OpenScenario 1.x.
+            //
+            // This is one of the OSI types which map from "bus" in OpenScenario 2.x.
```
clemenshabedank, line 844: Probably copy paste typo. I think it should map to VRU_vehicle from OpenSCENARIO 2.0
```suggestion
            // This is one of the OSI types which map from "VRU_vehicle" in OpenScenario 2.x.
```

osi_object.proto
caspar-ai: I'd have thought most trailers don't have a front axle?  A quick google search of "car trailer" suggests most have only rear or mid axles.

Perhaps the distinction is instead weight or size rather than axles or joining mechanism / overlap?

osi_object.proto
caspar-ai: Yep, makes sense, done.

osi_object.proto
caspar-ai: Yes, I must confess this was a guess :)

The definition in OSC 2.0 is to cover UN categories M2 and M3.  They are:

- 2.2.2. "Category M2": Vehicles used for the carriage of passengers, comprising more than eight seats in addition to the driver's seat, and having a maximum mass not exceeding 5 tonnes.
- 2.2.3. "Category M3": Vehicles used for the carriage of passengers, comprising more than eight seats in addition to the driver's seat, and having a maximum mass exceeding 5 tonnes.

So, trams and trains match both of those, but I guess it depends what you define "vehicle" to be.

It would be great to chat to the OSC team to work out their intention.

osi_object.proto
caspar-ai: See comment on trams.

osi_object.proto
caspar-ai: Good spot, fixed!

osi_object.proto
clemenshabedank: They are currently very busy, but I can post it as a question in the review process which is currently ongoing over there... if that' fine for you.

caspar-ai: Yes, perfect, thanks.

I don't think there is a rush here, I mainly put this in to highlight that we should be aiming to provide this kind of detail wherever we can.

osi_object.proto
HendrikAmelunxen: There are different trailer types, e.g., trailers with one axle in the middle or a tandem trailer. These are mainly towed by vehicles or agricultural vehicles.
Trailers towed by trucks mostly have either a front and a rear axle (often referred to as “full-trailer”) or just a rear-axle (“semi-trailer”). The semi-trailer is towed by a semi-tractor, which is not able to carry additional loads, but its only purpose is to tow the semi-trailer. “full-trailer” can be towed by trucks or agricultural vehicles. One specialty is a dolly, which is an adapter to couple semi-trailer to trucks or agricultural vehicle or generate road trains consisting of semi-trailers. This dolly variant transforms semi-trailers into full-trailers and should be handled like full-trailers.

I think semi-trailers should be handled with a separate type since they are widely used and have special characteristics due to their coupling mechanism. In that regard, it also makes to have semi-tractors as separate types to describe the pulling part for semi-trailers.

It might be reasonable to introduce the full-trailer as an additional trailer variant. By this we would have:
full-trailer: for big trailers with front and rear axle
Semi-trailer: for semi-tractor + semi-trailer couplings
Trailer: describing the general vehicle trailers

Would that be an approach?

caspar-ai: As discussed in the harmonisation meeting.  This is a complex area and trying to define precise boundaries between types is more than we want to take on as part of this PR which is primarily about clarifying semi-trailer / tractor definitions.

As such, I have remove any new attempts at defining boundaries, and I have written a note in the top level enum comment indicating that the imprecise boundaries are deliberate and left to the implementor to decide on in their specific use case.

Does that work for you?

osi_object.proto
caspar-ai: Thanks for raising!

I have removed any reference to 2.x from both tram and train while this discussion is ongoing.

stmswald:
Nice work, how about adding additional visual descriptions?
This could help to decide what type of car/truck/etc.

like:
image
ref: DOI: 10.1109/JSEN.2021.3115589

osi_object.proto
```diff
@@ -748,132 +761,125 @@ message MovingObject
             //
             // Definition: Hatchback car with maximum length 4 m.
             //
-            // This is one of the OSI types which map from "car" in both OpenScenario 1.x and 2.x.
+            // This is one of the OSI types which map from "car" in both
```
stmswald, line 764: How about setting those descriptions into a `\note` statement. This could help to show that OSI does not only "supports" openScenario?

thempen:
As we have discussed it in the last Harmonization group meeting, two points:

- The OpenSCENARIO 1.x and 2.x references should be specific versions, like OpenSCENARIO 1.1 and 2.0. We do not know how it will change in the future. 
- We have now multiple points in the whole documentation with references to OSC and ODR. I am not exactly sure how many and to which standards we have mappings. I have concerns in future maintenance with these references, because of the changes in the referencing documents. As we have suggested in #500, we could write Best-Practices guides for, e.g. the usage with OpenSCENARIO or OpenDRIVE to bundle the information about connections. Perhaps it could be some WP for 4.0.

caspar-ai:
> Nice work, how about adding additional visual descriptions? This could help to decide what type of car/truck/etc.
> 
> like: image ref: DOI: 10.1109/JSEN.2021.3115589

We can do these, but given we already know there are fuzzy boundaries and are saying defining the boundaries is left as an exercise for the user, I don't know if it is correct to try and then add guides which do try and precise the boundaries.

caspar-ai:
> As we have discussed it in the last Harmonization group meeting, two points:
> 
> * The OpenSCENARIO 1.x and 2.x references should be specific versions, like OpenSCENARIO 1.1 and 2.0. We do not know how it will change in the future.
> * We have now multiple points in the whole documentation with references to OSC and ODR. I am not exactly sure how many and to which standards we have mappings. I have concerns in future maintenance with these references, because of the changes in the referencing documents. As we have suggested in [Best-Practice Guide #500](https://github.com/OpenSimulationInterface/open-simulation-interface/issues/500), we could write Best-Practices guides for, e.g. the usage with OpenSCENARIO or OpenDRIVE to bundle the information about connections. Perhaps it could be some WP for 4.0.

Removed all references to OSC.

osi_object.proto
caspar-ai: Suggestion from the harmonisation group was to remove completely so I've done that.

stmswald:
> > Nice work, how about adding additional visual descriptions? This could help to decide what type of car/truck/etc.
> > like: image ref: DOI: 10.1109/JSEN.2021.3115589
> 
> We can do these, but given we already know there are fuzzy boundaries and are saying defining the boundaries is left as an exercise for the user, I don't know if it is correct to try and then add guides which do try and precise the boundaries.

DeliveryVan
Heavy Truck
SemiTrailer
SemiTruck

OSI Harmonization Group 12/7/2021 Ideas for Outline shapes

caspar-ai:
@stmswald those outlines look good to me.

HendrikAmelunxen:
@stmswald I think that helps to distinguish between the different trucks and trailer types. Maybe an additional shape for the trailer makes sense, like
image

caspar-ai:
@stmswald / @HendrikAmelunxen what's the next step here?  I'm assuming we need to turn those outlines into `.svg` files, add them to this PR and reference them from the comments so that they appear in the documentation.

Is one of you looking at doing that?

caspar-ai:
After discussing with @stmswald via email we think it would be good to get review from the CCB.

There are two things which we are unsure about with the documentation images that it would be good to get some feedback on from the CCB.

1. Are the images too big?  They are .svg files so can be scaled as needed, but I don't know if doxygen allows scaling hints, so do the uploaded images need to be set with a smaller default size?  Or are they OK as-is?
2. Do the images need to conform to a consistent style?  For example, many are done with black and white outline drawings.  Are there specific people, or tools, that are used to ensure a consistent style?

HendrikAmelunxen:
@caspar-ai Currently the images are too big, and I think it would make sense to have them in black and white in order to be more aligned with the other images. I think, a styleguide for images does not exist though.

image

I think it would make sense to scale the images down and make them in black and white. After that it would make sense to contact CCB.

caspar-ai:
Fixed up the images to be both smaller, and black outline drawings.  Given no particular style I've gone for the slightly "hand drawn" style using https://excalidraw.com/ hopefully they are OK.

Adding back in the CCB label based on the comment above, but let me know if you think they need tweaking again first.

kmeids:
**Output CCB 02.03.2022**:
1. @pmai to merge this.

pmai (approved):

PR #599, opened 2021-11-19
Logical Lanes in OSI

tbleher:
Adds "logical lanes" to OSI. These logical lanes correspond much more closely to lanes as defined in map formats like OpenDRIVE or OpenLaneModel. Thus they can be used to represent maps more closely in a standard manner.

Logical Lanes make it much easier to implement agent models using OSI as their input. Especially since the logical lanes on intersections can be represented now, fixing a long-known problem when trying to implement agent models using OSI.

The change is backward compatible, so can be added to OSI 3.5.


osi_common.proto
```diff
@@ -338,6 +338,45 @@ message Spherical3d
     optional double elevation = 3;
 }
 
+//
+// \brief Assignment of an object to a logical lane
+//
+// An object is assigned to a logical lane if it overlaps the logical lane.
+// Assignment happens even if the reference point is outside the lane, and only
```
rockteresa, line 345: Is there a definition how large the overlapping area must be for a lane assignment?
Is it possible to read  the size of the overlapping area?

osi_common.proto
```diff
@@ -338,6 +338,45 @@ message Spherical3d
...
+    //
+    // s_position might be outside [s_start,s_end] of the lane (and even
+    // outside [startS,endS] of the reference line) if the reference point is
+    // outside the lane, but the object overlaps the lane or a TrafficSign or
+    // TrafficLight is assigned to a lane.
+    //
+    optional double s_position = 2;
+
+    // T position of the object reference point on the lane, in the ST
+    // coordinate system of the lane.
```
rockteresa, line 368: s/t of the lane or the reference line?

osi_logicallane.proto
```diff
@@ -0,0 +1,471 @@
...
+    // Notes on design decisions:
+    // - The S coordinate is included directly, both for OpenDRIVE
+    //   compatibility, and to speed up calculations.
+    // - The rules on S coordinates (e.g. the calculation in 2D space) are
+    //   there to ensure OpenDRIVE compatibility.
+    // - The rules on T coordinates are there to ensure OpenDRIVE compatibility
+    //   for lanes without superelevation, and to make it easier to convert
+    //   between OSI and OpenDRIVE in case superelevation is present.
+    //
+    // TODO Describe needed precision.
```
rockteresa, line 123: Is a minimum of points per length/curverture required in order t prevent errors regarding interpolation and t coordinate assignment?

osi_logicallane.proto
```diff
@@ -0,0 +1,471 @@
...
+//     more data needed. This can mostly be mitigated by only transmitting the
+//     lane boundaries during initialization (e.g. via the OSMP GroundTruthInit
+//     message).
+// - The LogicalLaneBoundary contains all data directly which an agent model is
+//   likely to need. It does not include information normally only used by
+//   sensor models (e.g. the exact length of the color markings on the road).
+//   This information can be gotten from the physical lane referenced in the
+//   LogicalLane, if needed.
+//
+// TODO To Discuss: It would be entirely possible to add ST coordinates also to
```
rockteresa, line 162: Open to discuss if we adjust the  the initializing process, so that the user is able to decide which informaton are transmitted in online use?

osi_logicallane.proto
```diff
@@ -0,0 +1,471 @@
...
+    // The reference line is used as a coordinate system on this lane.
+    //
+    // The reference line should roughly have the same shape as the lane, so
+    // that S coordinates continually increase/decrease along the lane. It is
+    // not required that the reference line has the same direction as the
+    // driving direction of the lane.
+    //
+    // Neighbouring lanes are strongly encouraged to use the same
+    // reference_line, so that vehicles that are next to each other on
+    // neighbouring lanes have comparable S positions.
```
rockteresa, line 344: neighbour relation vs overlapping relation: how to define neighbour relations when lanes may overlap? When is a lane the right/left neighbour?

osi_logicallane.proto
```diff
@@ -0,0 +1,471 @@
...
+    // - The S distance between two points may be larger than the 2D euclidean
+    //   distance, but should be not much larger. It is allowed to be larger if
+    //   the underlying reference line (e.g. in an OpenDRIVE map) is a curve,
+    //   and thus the sampled road has a smaller length than the original
+    //   curve.
+    //
+    // Together, these rules allow directly putting OpenDRIVE S coordinates
+    // into an OSI ReferenceLine.
+    //
+    // Between to ReferenceLinePoints, both the world coordinate and the S
```
clemenshabedank, line 59:
```suggestion
    // Between two ReferenceLinePoints, both the world coordinate and the S
```

osi_logicallane.proto
```diff
@@ -0,0 +1,471 @@
...
+    // All LogicalLaneBoundary elements referenced here must refer to the same
+    // ReferenceLine as this lane.
+    // The boundaries together must cover the whole length of the lane (the
+    // range [start_s,end_s]) without gap or overlap.
+    //
+    // TODO Is it allowed for the boundaries to be longer than the lane?
+    // Theoretically, this could be done. Then the user would have to limit the
+    // boundary itself to [start_s,end_s].
+    //
+    repeated Identifier left_boundary_id = 13;
```
tbleher, line 408: TODO Add condition that boundaries are sorted in ascending S direction.

osi_common.proto
tbleher: The ST coordinate system is defined by the reference line, so these are identical. I can reword to refer to the reference line.

osi_logicallane.proto
```diff
@@ -0,0 +1,502 @@
...
+    // - Later points in the list must have strictly larger S coordinates than
+    //   earlier points.
+    // - For consecutive points, the S difference between them  must be at
+    //   least as large as the 2D euclidean distance between the points (2D
+    //   distance == euclidean distance between the points taking only X and Y
+    //   into account).
+    // - The S distance between two points may be larger than the 2D euclidean
+    //   distance, but should be not much larger. It is allowed to be larger if
+    //   the underlying reference line (e.g. in an OpenDRIVE map) is a curve,
+    //   and thus the sampled road has a smaller length than the original
```
clemenshabedank, line 53:
```suggestion
    //   and thus the sampled reference line has a smaller length than the original
```

osi_logicallane.proto
```diff
@@ -0,0 +1,502 @@
...
+    // coordinate is added. T is the signed 2D distance (i.e. hypot(A.X-B.X,
+    // A.Y-B.Y), if A and B are the two points) between the point to describe
+    // and the nearest point on the polyline (this point might either be on a
+    // line segment or at an edge between two line segments). The distance is
+    // positive if the point is left of the polyline (in definition direction),
+    // negative if it is right of it. Note that the "nearest point on the
+    // polyline" is calculated in 3D, in order to choose the correct point for
+    // 3D curves (think reference lines for roads in parking decks). If there
+    // are several "nearest points", the one with the smallest S coordinate on
+    // the polyline is chosen.
```
clemenshabedank, line 96: TODO: At this point I could imagine to incorporate Daniel's images on that topic.

osi_logicallane.proto
```diff
@@ -0,0 +1,502 @@
...
+    // \image html line_approximation_error.svg "Approximation error"
+    // Approximation error green line.
+    //
+    // The Z error (difference in Z height between boundary_line and the "real"
+    // line) must also not exceed 5cm.
+    //
+    // Note: if two lanes have different Z heights (e.g. a driving lane is
+    // beside a sidewalk, where the sidewalk is 10cm higher than the road),
+    // then these lanes cannot share a boundary, since their boundaries have
+    // different Z heights.
```
clemenshabedank, line 207: That's a good comment, which we should also add to the physical boundaries IMO.

osi_logicallane.proto
```diff
@@ -0,0 +1,502 @@
...
+    //
+    // If the boundary approximates a curve (e.g. a cubic function in
+    // OpenDRIVE), the points must be chosen in a way that the lateral distance
+    // to the ideal line does not exceed 5cm. As shown in the following image:
+    //
+    // \image html line_approximation_error.svg "Approximation error"
+    // Approximation error green line.
+    //
+    // The Z error (difference in Z height between boundary_line and the "real"
+    // line) must also not exceed 5cm.
```
clemenshabedank, line 202: I think in the last meeting we said that we should think about if this is a reasonable value for z. The value we decide probably also makes sense for physical lane boundaries.

osi_logicallane.proto
```diff
@@ -0,0 +1,502 @@
...
+    repeated LogicalBoundaryPoint boundary_line = 2;
+
+    // The reference line for this boundary.
+    //
+    // The reference line is used as a coordinate system for this boundary.
+    // All points of this LogicalLaneBoundary must have S coordinates in the
+    // range [sStart,sEnd].
+    //
+    // The reference line should roughly have the same shape as the boundary, so
+    // that S coordinates continually increase along the lane.
```
clemenshabedank, line 218: I think there is no mention yet, for how many lanes a reference line should be used and where it should be located. If this is on purpose left up to implementation, probably this needs a mention, otherwise maybe some rules like in OpenDRIVE. The singular use in `should roughly have the same shape as the boundary` sounds a bit as if each boundary should have its own reference line which I think is not what you intend.

osi_logicallane.proto
```diff
@@ -0,0 +1,502 @@
...
+        // End s position of the relation
+        //
+        // Requirement: end_s > start_s
+        //
+        optional double end_s = 3;
+
+        // Start s position of the relation on the other lane.
+        //
+        // This is the same place as start_s, but measured along the reference
+        // line of the other lane.
```
clemenshabedank, line 485: In line 362 you recommend for neighboring lanes to use the same reference line. So when adopting this recommendation this field here would have the same value as start_s?

osi_logicallane.proto
```diff
@@ -0,0 +1,502 @@
...
+        //
+        optional double start_s_other = 4;
+
+        // End s position of the relation on the other lane.
+        //
+        // This is the same place as end_s, but measured along the reference
+        // line of the other lane.
```
clemenshabedank, line 492: In line 362 you recommend for neighboring lanes to use the same reference line. So when adopting this recommendation this field here would have the same value as end_s?

osi_logicallane.proto
```diff
@@ -0,0 +1,502 @@
...
+    //
+    // TODO Is it allowed for the boundaries to be longer than the lane?
+    // Theoretically, this could be done. Then the user would have to limit the
+    // boundary itself to [start_s,end_s].
+    //
+    repeated Identifier left_boundary_id = 13;
+
+    // Lanes that directly are connected to this lane at the beginning
+    //
+    repeated LaneConnection predecessor_lane = 14;
```
clemenshabedank, line 443: Just a suggestion to shorten code and reuse things: Couldn't you just take `repeated Lane.Classification.LanePairing` with a comment that logical lanes should be used? And skip everything till line 462? I am not sure if this suggestion would be better: I think it would definitely be nicer if the fields in the LanePairing message itself were repeated (and then using it with `optional`).

osi_logicallane.proto
```diff
@@ -0,0 +1,502 @@
...
+//
+// \brief A logical lane in the road network.
+//
+// A logical lane is part of a road. Compared to a physical lane (OSI type
+// \c Lane), its existence doesn't hinge on the existence of road markings. So
+// e.g. a narrow urban road with two driving directions but no road markings
+// in-between would be presented as two LogicalLanes, but only one Lane. So one
+// Lane can consist of multiple LogicalLanes. E.g. on intersections, each
+// driving path is one LogicalLane, but the whole area is one \c Lane of type
+// \c #TYPE_INTERSECTION.
```
clemenshabedank, line 269: I think here we might need an even more precise definition of what a logical lane really is / what its boundaries are when not limited by physical lane boundaries (e.g. how wide are the overlapping logical lanes in narrow urban roads?). Or write that this is free for implementations to decide. 

While I currently don't have a general approach, I think in intersections it would be quite easy to define (width of ingoing and outgoing lanes) and also for the case you sketch in line 330ff. (width of the preceding and succeeding lanes.

osi_logicallane.proto
```diff
@@ -0,0 +1,502 @@
...
+    // The reference line is used as a coordinate system on this lane.
+    //
+    // The reference line should roughly have the same shape as the lane, so
+    // that S coordinates continually increase/decrease along the lane. It is
+    // not required that the reference line has the same direction as the
+    // driving direction of the lane.
+    //
+    // Neighbouring lanes are strongly encouraged to use the same
+    // reference_line, so that vehicles that are next to each other on
+    // neighbouring lanes have comparable S positions.
```
ndunningbmw, line 364: How can neighbouring lanes have the same reference line, when the reference line is defined in ground truth frame-space? Does this refer to having the same S length?

osi_logicallane.proto
```diff
@@ -0,0 +1,502 @@
...
+    //  l1                       l2
+    // ----------    l3        ---------
+    //  l4                       l5
+    // ---------------------------------
+    //
+    // In this case, we have five physical lanes: l1, l2, l4 and l5 where a
+    // lane marking is present. And l3 is one lane covering the whole road
+    // (because no road marking is present).
+    //
+    // Now there are two basic possibilities to define the logical lanes:
```
adrianschultz, line 340: We (Esther and I) would prefer option 1 with the two logical lanes for this example.

osi_logicallane.proto
```diff
@@ -0,0 +1,502 @@
...
+    // Entries must be ordered: first by start_s, then by end_s.
+    //
+    repeated LaneRelation left_adjacent_lane = 10;
+
+    // Lanes that partially or completely overlap this lane. Only overlaps
+    // laterally larger than 5cm are considered overlaps for the purpose of
+    // this relation.
+    // Entries must be ordered: first by start_s, then by end_s.
+    //
+    repeated LaneRelation overlapping_lane = 11;
```
adrianschultz, line 405: Does this only apply for junctions or, are there other places where this can happen?

osi_logicallane.proto
```diff
@@ -0,0 +1,502 @@
...
+    // All LogicalLaneBoundary elements referenced here must refer to the same
+    // ReferenceLine as this lane.
+    // The boundaries together must cover the whole length of the lane (the
+    // range [start_s,end_s]) without gap or overlap. The boundaries must be
+    // stored in ascending order, starting with the smallest S position.
+    // Consecutive boundaries must share a point: the last point of the
+    // previous boundary must be identical to the first point of the next
+    // boundary.
+    //
+    // TODO Is it allowed for the boundaries to be longer than the lane?
```
adrianschultz, line 418: We would allow to have boundaries longer than lanes. Esther will make an example.

osi_logicallane.proto
tbleher: `LanePairing` cannot be used, because it doesn't identify where the lanes are connected.

Example 1 (arrow describes the reference line direction):

```
 ---------------  ---------------
   -> lane 1            -> lane 2
 --------------- ----------------
```

Example 2:
```
 ---------------  ---------------
   -> lane 1            <- lane 2
 --------------- ----------------
```

In both examples, lane 2 is a successor of lane 1, since it is connected to lane 1 at the end of lane 1 (in reference line direction).

But in example 1, `at_begin_of_other_lane` is true, but in example 2, it is false.
This distinction is important for agent models, and cannot be expressed using `LanePairing` currently.

osi_logicallane.proto
tbleher: Agreed. I think it depends on the use-case. If doing visualization for humans (e.g. in a driving simulator), 5cm is already quite large, since e.g. a pedestrian will either appear to float, or his shoes will not be completely visible, if the error is 5cm. So for this use-case, the error should be more in the 1-2cm range.

For agent models deciding where and how they can drive, 5cm error also sounds a bit large: imagine two lanes are connected and in reality they have a height difference of 5cm. Humans would drive really slow if the road height changed by 5cm (imagine road works where the road suddenly is 5cm lower). If this is hidden from OSI, then agents will drive far too fast across such an edge. So here, too, 5cm is too large, but probably even 2cm error is too large for this use-case.

So something like 1-2cm for Z error?

osi_logicallane.proto
tbleher: Hmm, the original purpose of this pull request was to make as much map information as possible available to OSI. So I would rather not define this too strictly, but say something like "lanes from the underlying map should be represented as accurately as possible". I think it is OK to split or merge them like this:

```
 --------------          --------  ------
 --> lane1       ==>  lane 2  lane 3
 --------------          --------  ------
```

Splitting them like this doesn't change any semantics. But the widths etc. should not be changed.

Rationale: if we define separate rules, then it WILL diverge from OpenDRIVE, which means that again OpenDRIVE maps cannot be represented in OSI correctly, and implementations will have more effort to convert between different formats.

osi_logicallane.proto
tbleher: It depends on the map format. Currently in OpenDRIVE this can only happen in junctions (at least if correctly following the specification). This might change depending on how bike lanes will be defined in OpenDRIVE 1.8 (see https://code.asam.net/simulation/standard/opendrive/-/issues/18).

If using e.g. OpenLaneModel as the underlying map, then restricted lanes (like the above-mentioned bike lines, but also bus lanes) can already overlap other lanes, even outside of junctions.

osi_logicallane.proto
tbleher: I'll have to clarify this one. I'll add some additional text and pictures later.

Basically one can use the same reference line if the lanes are parallel for their whole length.
But if you think of an intersection with a left-turn and a right-turn lane, e.g like this:
```
        /     /
-------/     /
left turn   /
-----------X
right turn  \
---------\   \
          \   \
```
These are neighbours for a while (which should be reflected in their lane relations), but cannot reasonably share a reference line. Since the reference line might even be in different directions for the area where they are neighbours (can be done in OpenDRIVE), their s values might be hugely different. That's why separate values per lane are needed there.

osi_logicallane.proto
clemenshabedank: 1-2cm sounds like a good approximation to me. So e.g. speed bumps could be modelled I assume. @arnedueselder what do you think? Do you think that would be a reasonable value for the surface lines as well?

osi_logicallane.proto
clemenshabedank: After yesterday's discussion I have a better feeling deviating from `Lane.Classification.LanePairing` because that definition might be dropped anyhow at some point due to ambiguities as you point out. Also I actually like your approach better, because predecessor and successor lanes a separated fields which make predecessor/successor relationships easier for some roads like this one:
```
---------------\                                       /----------------------
   ->   l1       \------------ -----------------------/          ->    l2
 ----------                    ->        l3                -----------------
   ->   l4      /----------------- -------- ----------\           ->    l5
  ------------ /                                        \-----------------------
```

clemenshabedank: @FabianPfeufferCAP for your awareness. This new suggestion might make the current LanePairing a candidate for deprecation, especially if the centerline gets dropped. Would this suggestion here also be a solution to your issues raised in https://github.com/OpenSimulationInterface/open-simulation-interface/pull/542?

osi_logicallane.proto
```diff
@@ -0,0 +1,502 @@
...
+    // The reference line should roughly have the same shape as the boundary, so
+    // that S coordinates continually increase along the lane.
+    //
+    optional Identifier reference_line_id = 3;
+
+    // The classification of the lane boundary.
+    //
+    optional LaneBoundary.Classification classification = 4;
+
+    // Optional external reference to the lane boundary source.
```
EstherHekele, line 226: This LaneBoundary classification is reused from the physical LaneBoundary. i.e noLine, Dashes etc.
As this is a logical lane we could maybe replace this with logical information passing to the next lane is allowed.
NoLine and dashed would in both case be passing. 
But we must state in with direction the passing/or noPassing applies.
While I was thinking on how and where to this comment, I realised, that we must discuss Solid-broken roadmarks in our discussion of physical and logical LaneBoundary. I put something on the miro board we can use as a basis for our discussion.

osi_groundtruth.proto
```diff
@@ -171,4 +172,16 @@ message GroundTruth
...
+    //
+    repeated ReferenceLine reference_line = 17;
+
+    // Logical lane boundaries used by LogicalLane
+    //
+    repeated LogicalLaneBoundary logical_lane_boundary = 18;
+
+    // Logical lanes used e.g. by traffic agents
+    //
+    repeated LogicalLane logical_lane = 19;
```
EstherHekele, line 186: Not sure if this is the right place. 
The logical lanes and physical lanes should be the same. I.e curb is a logical and Physical lane

osi_logicallane.proto
arnedueselder: We have to keep in mind that logical lane boundaries might be placed even if there is no "real" LaneBoundary (as maybe on wider bi-directional streets), but i agree that maybe the LaneBoundary classification doesnt have to be used here and be replaced with your idea of passing.

osi_logicallane.proto
```diff
@@ -0,0 +1,634 @@
...
+    //
+    // This describes how vehicles may move across the LogicalLaneBoundary. The
+    // PassingRule is determined solely based on the road, not on any signs
+    // (i.e. it may be overridden by signs).
+    //
+    enum PassingRule {
+        // Passing rule cannot be determined, e.g. because it depends on the
+        // agent type.
+        //
+        CANNOT_BE_DETERMINED = 0;
```
clemenshabedank, line 321:
```suggestion
            // Passing rule is unknown (must not be used in ground truth).
            //
            PASSINGRULE_UNKNOWN = 0;

            // Other (unspecified but known) passing rule.
            //
            PASSINGRULE_OTHER = 1;
```
as discussed in today's meeting.

osi_logicallane.proto
```diff
@@ -0,0 +1,634 @@
...
+
+        // Only passing from right to left (in definition direction, so in
+        // increasing T direction) is allowed.
+        //
+        RIGHT_TO_LEFT = 3;
+
+        // Passing is allowed in both directions (left to right and right to
+        // left).
+        //
+        BOTH_ALLOWD = 4;
```
clemenshabedank, line 341:
```suggestion
        BOTH_ALLOWED = 4;
```
typo

osi_logicallane.proto
```diff
@@ -0,0 +1,634 @@
...
+    //   position (if there are multiple physical lane boundaries), a new
+    //   LogicalLaneBoundary must be created.
+    // - The referenced LaneBoundary objects may be longer than the
+    //   LogicalLaneBoundary which references them, but must never be shorter.
+    //
+    // Example:
+    //      Lane 1
+    // --------a------------------ - - - -c- - - -
+    // - - - -b- - - -
+    //      Lane 2
```
clemenshabedank, line 259: As discussed, a bit more explanation e.g. where goes the reference line in this example...

osi_logicallane.proto
```diff
@@ -0,0 +1,634 @@
...
+    enum PassingRule {
+        // Passing rule cannot be determined, e.g. because it depends on the
+        // agent type.
+        //
+        CANNOT_BE_DETERMINED = 0;
+
+        // No passing is allowed (neither from left to right nor from right to
+        // left).
+        //
+        NONE_ALLOWED = 1;
```
clemenshabedank, line 326: change naming convention and enumeration values accordingly for the other enum field names...

osi_logicallane.proto
tbleher: Somehow was not notified of this suggestion; added something similar to a followup commit.

osi_logicallane.proto
```diff
@@ -0,0 +1,516 @@
...
+    //
+    // \rules
+    // is_globally_unique
+    // \endrules
+    //
+    optional Identifier id = 1;
+
+    // The type of the logical lane.
+    //
+    optional Lane.Classification.Subtype type = 2;
```
clemenshabedank, line 269: Decision in Road Model Meeting: Should be separated between logical lane classification and physical lane classification because those types might diverge.

osi_common.proto
tbleher: Added a note that assignment has to be added as soon as object overlaps more than 5cm.
Size of the overlapping area has to be calculated by the application, but st coordinates help there.

osi_logicallane.proto
tbleher: Sounds reasonable. I won't do it for this pull request (I don't want to make it larger or more contentious than necessary). Do you want to create a pull request for this?

clemenshabedank: Yes makes sense, I'll do and close this comment

osi_logicallane.proto
tbleher: Clarified the text. Probably needs an image to explain.

osi_logicallane.proto
tbleher: The reference line just defines a non-euclidean coordinate system for use by the lane. If two lanes share a reference line, then it is much easier to compare e.g. vehicle positions.

osi_logicallane.proto
tbleher: Changed as per your preference :)

osi_logicallane.proto
tbleher: Makes sense. Adapted accordingly.

osi_logicallane.proto
tbleher: Done in a new version of the pull request.

osi_logicallane.proto
tbleher: Done.

osi_logicallane.proto
tbleher: Addressed in the latest commit.

osi_logicallane.proto
tbleher: This should already work: _osi3::SensorViewConfiguration::omit_static_information_ can be set by the agent model, and should result in it not getting any logical lanes at runtime (but still getting the lane assignments of the objects).

osi_logicallane.proto
tbleher: Added an image showing both overlapping and neighbour lanes.

osi_logicallane.proto
tbleher: I added a similar image, roughly based on the ones from @dbeckerAC . I think the text from Daniel could be added as an additional separate text, but I won't do this as part of this change (because of lack of time). I think this can be done later on-top.

osi_logicallane.proto
tbleher: Added some explanation.

osi_groundtruth.proto
tbleher: Logical lanes can now reference physical lanes, so the relation between the two types is now represented in OSI.

tbleher:
A small note: the remaining test errors (from the Proto2 and Proto3 builds) are from `python -m unittest discover tests` requiring that members of a message not use \brief in their documentation. However, I find the generated documentation much more readable if \brief is used. Rationale: the attribute list in Doxygen then already shows  a short summary what each attribute is for.
So I kindly request that this restriction is lifted, and \brief be allowed for members of a message.

osi_logicallane.proto
tbleher: Allowed Z error set to 2cm.

pmai:
@OpenSimulationInterface/asam-project-members CCB decision 2022-02-02: This PR is the first outcome of the road model working group to better support non-sensor model uses of OSI in terms of information for the road model. Please review the PR in the near term for fundamental fitness for your use cases and provide feedback, so that CCB evaluation can be based on a broad review.

The PR is currently slated for inclusion in the next minor release, since it is fully backward compatible.

stefancyliax:
> This PR was already discussed extensively in the Road Model Working Group, where several implementers of both agent models and simulation frameworks are present. As you can see from the history of this PR, there were lots of comments and reviews, and the PR was adapted accordingly. Before forwarding this to the CCB, there was unanimous consent in the working group that this should be merged in its current form.
> 
> If there are specific points that needs adapting, I'm happy to discuss them.

The CCB yesterday didn't feel equipped to merge this PR, due to missing feedback from implementers. Maybe it was just lack of information.
There is now a timeline and tasks defined. I don't see any more hurdles to merging on 30.03.

stefancyliax:
Got a positive feedback from OpenDRIVE project.

osi_referenceline.proto
```diff
@@ -0,0 +1,190 @@
...
+    //   points that are "outside" the reference line.
+    //
+    // The sampling of the polyline must be chosen such that the error
+    // when converting coordinates is "small enough". The exact needed
+    // precision is defined for each user, where the reference line is
+    // referenced.
+    //
+    // ## Various notes
+    //
+    // Notes on OpenDRIVE compatibility:
```
ndunningbmw, line 149: The only thing I am missing here is any advise regarding how the Reference Line should be handled when the source is now OpenDRIVE. Is the intention or recommendation that similar rules to OpenDRIVE are followed? e.g. on a bi-directional road, the reference line is between the lanes of opposing direction and on uni-directional roads, the reference line is nearest to the central division between the opposing direction.

If this is the case, I would find it beneficial to state that here, as the Reference Line is an integral part of the Logical Lane structure.

HendrikAmelunxen:
In general this looks good, but I am stumbling over the LogicalLane-TYPE_CURB. 
The curb will be a specific logical lane; this fits to OpenDRIVE. But the curb is also a specific physical laneboundary. How to describe the curb shape and height? E.g. lowered curbs, curbs with lateral inclination, etc. Physical laneboundaries, that are able to describe the shape of curbs to a certain extend, cannot be referenced by logical lanes. 
Is it supposed that the height and the shape of the curbs are handled by the surface lines later on? 
Will sections with changed curbstones, e.g. section with lowered stones, introduce a new lane?
Maybe this has to be described a little bit more.

tbleher:
> In general this looks good, but I am stumbling over the LogicalLane-TYPE_CURB. The curb will be a specific logical lane; this fits to OpenDRIVE. But the curb is also a specific physical laneboundary. How to describe the curb shape and height? E.g. lowered curbs, curbs with lateral inclination, etc. Physical laneboundaries, that are able to describe the shape of curbs to a certain extend, cannot be referenced by logical lanes. Is it supposed that the height and the shape of the curbs are handled by the surface lines later on? Will sections with changed curbstones, e.g. section with lowered stones, introduce a new lane? Maybe this has to be described a little bit more.

A few points:

- A LogicalLane will have associated lane boundaries, and these can reference the physical lane boundaries.
- The logical lane boundaries have a height. So even without associated physical lane boundaries, the height at the left and the right is described, which should cover most cases (e.g. lowered curbs can be described, linear lateral inclination can be described). For more fancy cases, surface lines presumably can be used.
- A changed curb height (like a lowered curb in front of a building) should not normally lead to a new lane. The height can be described using the Z position of the boundaries.

Does that answer your question, @HendrikAmelunxen ?
I think I can add some additional description regarding height changes.

Regarding curb stones in particular: I think the mapping between OpenDRIVE and LogicalLanes should be fairly straightforward. Mapping from OpenDRIVE to physical lanes and lane boundaries seems more difficult. I think it would be good to have a document that describes these compatibility issues. But I think this goes beyond the scope of this patch.

tbleher:
@HendrikAmelunxen I have added https://github.com/OpenSimulationInterface/open-simulation-interface/pull/599/commits/34484f41b60848148e3e19a80e06840cbe1500e6 to explain height changes a little more.
Does that look sensible to you?

I thought about describing the lane height further, but as far as I know, lane surface information (or height in general) is not really described in OSI. I think describing this better belongs to PR #619.

osi_referenceline.proto
tbleher: @ndunningbmw Good question. I added a note at https://github.com/OpenSimulationInterface/open-simulation-interface/pull/599/commits/c3c0244f392938d4cbee073eeecb539d5261de6d . Does that answer your question?

ndunningbmw: Looks good to me!

tbleher: Thanks for the fast response :) Closing this then.

clemenshabedank:
As you @HendrikAmelunxen and @tbleher state, hight information is planned to come primarily from surface lines (PR619). The x,y,z fields in the lane boundaries and the logical lane boundaries will be a bit of a duplication regarding hight information but they currently carry certain differing requirements (e.g. 5cm rule for curvature). Going forward it could be discussed to reuse the surface lines to describe lane boundaries (partly), i.e. deprecate LaneBoundary.boundary_line and instead point to a surface line.  

For the hight profile of curbs there are currently different approaches on the table (incl. PR608 and PR436) while they all seem to have shortcomings. For coming up with a decision we said we needed input from CCB, maybe even ISO23150. 

A bit of a problem I still see (and I think @HendrikAmelunxen you might have the same reading your comment above). So thinking the LogicalLane with TYPE_CURB through, we would have no value for LogicalLane.physical_lane_reference (I think), because there doesn't exist a physical Lane of type curb. Then I guess we would have a LogicalLane.right_boundary_id and a LogicalLane.left_boundary_id. But what values should the respective LogicalLaneBoundary.physical_boundary_id and LogicalLaneBoundary.passing_rule have? So probably LogicalLaneBoundary.physical_boundary_id should be empty and LogicalLaneBoundary.passing_rule for the lower edge PASSING_RULE_NONE_ALLOWED. Maybe this PR should be a bit more specific regarding that issue, if it is meant like this. @tbleher what do you think?

clemenshabedank:
Here a some options on curbs (physical) we already discussed in the Road Model group: [Curb_ModellingOptions.pptx](https://github.com/OpenSimulationInterface/open-simulation-interface/files/8368399/Curb_ModellingOptions.pptx)

clemenshabedank:
@tbleher thanks for doing this survey on such short notice! It's interesting to see that we can't really harmonize with both OpenDRIVE and NDS at the same time in that regard. To complement this the ISO23150 (where we also have harmonization requirements - in my feeling the strongest) treats curbs also as boundaries AFAIK. Therefore as well as for backward-compatibility reasons I would prefer to leave curbs as (physical) lane **boundaries**. 

Since we probably can't make the decision of potentially changing the (phyiscal) curb to a lane (which I would not prefer) prior to merging this PR I think we should go with the current definition of a curb as a (physical) lane **boundary**.

Nevertheless since agent models have stronger requirements to harmonize with OpenDRIVE, and LogicalLanes shall be consumed especially by agent models rather than sensor models I think we can and should keep it as it is in this PR (although it is ugly that we treat curbs differently). 

So I would suggest to add comments for: 
- LogicalLane.physical_lane_reference should stay empty for the curb case -> we seem to agree on this. 
- LogicalLaneBoundary.physical_boundary_id: I would find referencing only one (left or right) random. Both is probably better than empty because then there is access to fields of the physical curb (e.g. hight). 
- The passing rule currently reads more from a legal perspective. If traversing is possible or safe for vehicles I guess depends on other things like vehicle type and speed, so I would say we better leave it out of scope and up for the agent model to decide. We currently have the hight field for physical curbs, so this information could be used by agent models if needed. So a comment that for LogicalLaneBoundary.passing_rule of LogicalLanes of type curb should always be PASSING_RULE_OTHER might be the best solution. 

Do you want me to make an inline suggestion for that @tbleher ?

clemenshabedank:
And of course @HendrikAmelunxen would that help to clarify?

tbleher:
@clemenshabedank Thank you for all the sensible suggestions! Yes, an inline suggestion would be appreciated :)

osi_logicallane.proto
```diff
@@ -0,0 +1,805 @@
...
+        //
+        optional double t_position = 3;
+    }
+
+    //
+    // Passing rule of the LogicalLaneBoundary.
+    //
+    // This describes how vehicles may move across the LogicalLaneBoundary. The
+    // PassingRule is determined solely based on the road, not on any signs
+    // (i.e. it may be overridden by signs).
```
clemenshabedank, line 200:
```suggestion
    // This describes how vehicles are legally allowed to move across the 
    // LogicalLaneBoundary. The PassingRule is determined solely based 
    // on the semantics of physical road markings, not on any signs
    // (i.e. it may be overridden by signs).
```

tbleher:
> These suggestions try to fix the issues in [#599 (comment)](https://github.com/OpenSimulationInterface/open-simulation-interface/pull/599#issuecomment-1081642449)

Thank you @clemenshabedank for the suggestions! All applied.

clemenshabedank (approved):

stefancyliax:
OSI CCB 30.03.2022:
- All feedback from BMW was implemented. There was no other feedback that would prevent a merge. 
- @tbleher is fixing the DCO
- @stefancyliax to merge

stefancyliax:
Merging this as decided in the OSI CCB on 30.03.2022

PR #600, opened 2021-11-21
Draft for discussion: Surface lines to model road shapes and heights

arnedueselder:
#### Surface lines concept as proposed in the OSI Road Network Model Group
First draft on how to implement the surface line concept in OSI. As discussed, the approach follows the concept of Vector3D line segments that form a lane and are used to retrieve height information at different parts of that lane.
Speed bumps etc. could be modelled with more points, while almost flat surfaces would require none or almost none surface lines at all.

Height information of lane boundaries should be also included. The concept should only be used for a very rough description of the height profile and not used for a very detailed description of the road surface.
It would replace the currently implemented centerline for now (but not necessarily). The intend is to separate physical properties from logical information. It would still be possible though to place a surface line in the center of the lane to keep that information.

Concept will be expanded to be used in junctions, but is still open for discussion within the concept group.


arnedueselder:
Discussions from Meeting at 22.11.2021:

- If more than one surface line exists they should maybe contain an equal number of points along the lane to make indexing positions easier. 
-   That would probably mean also that either lane boundaries should be sampled the same or that a lane boundary also contains its own surface line with the appropriate number of total points.
-   Approximation definition for the current centerline should be expanded and applied also for height information to minimize errors between to points. 
- Surface lines possibly are better placed within logical lanes instead of lanes #599 

Enums for new variables should be added at the end to be backwards compatible.

thomassedlmayer:
Maybe we would also like to describe the transition from a road surface to an elevated sidewalk with this approach.

In order to describe the curb elevation using the current surface concept we would have to define another right- or left-adjacent surface line (or lane boundary) which is located above the bottom lane boundary. Is this a good way to define it?

Also, there is a height parameter for each `BoundaryPoint` of a lane boundary, which already defines some vague height of curbs (height ≠ z-coordinate). There is no definition of its "direction vector" though (e.g. perpendicular to some surface that does not exist yet?). We should probably think about some precise specification there. Maybe it could be used for the surface concept?
image

osi_lane.proto
```diff
@@ -192,69 +192,91 @@ message Lane
...
         //
-        // \note Intersections and non-driving lanes do not have a center line.
-        // A vehicle must calculate this individually and depending on the
-        // situation.
+        // \rules
+        // is_globally_unique
+        // \endrules
         //
-        repeated Vector3d centerline = 3;
+        optional Identifier surfaceline_id = 4;
```
clemenshabedank, line 240: Because potentially multiple surface lines per lane are envisioned, it wouldn't work like this I think. I think it would make sense to construct a new `message surfaceline{}` with an optional field `Identifier surfaceline_id` and a repeated field `Vector3d surfaceline`. This can message can be used here repeatedly, then. It could also (or instead) be included in the LogicalLanes as you write in the conversation section, so for easier usage it might make sense to put in `osi_common.proto`. What do you think?

osi_lane.proto
```diff
@@ -192,69 +192,91 @@ message Lane
...
+        // \attention If more than one surface line exists, they are not allowed to cross each other 
+        // or the lane boundaries, otherwise you could have conflicting height information at these points.
         //
-        // \note The center line is the line that a typical vehicle follows more
-        // or less (depending on the situation, a little more to the left or
-        // right of the center line). The intended direction of travel on the
-        // lane is given by the direction defined by the sequence of points
-        // forming the \c #centerline w.r.t. \c
-        // #centerline_is_driving_direction.
+        repeated Vector3d surfaceline = 3;
```
clemenshabedank, line 229: see comment below (line 240)

osi_lane.proto
```diff
@@ -274,7 +296,7 @@ message Lane
         // check_if this.type is_different_to 4 else do_check is_set
         // \endrules
         //
-        repeated Identifier left_adjacent_lane_id = 5;
+        repeated Identifier left_adjacent_lane_id = 8;
```
clemenshabedank, line 299: For backward-compatibility reasons please avoid renumbering. Instead, new fields should get not yet used numbers, while the textual order inside the file doesn't matter. (Applies also to changes below, of course)

osi_lane.proto
```diff
@@ -192,69 +192,91 @@ message Lane
...
+        
+        // List of IDs of all surfaceline segments that are directly adjacent to the
+        // surfaceline on the left side (w.r.t. ascending order of surfaceline points
+        // and lane boundary points). Note that lengths of surfaceline segments are 
+        // not synchronized and therefore can contain multiple adjacent segments.
+        //
+        // \note If none or just one surfaceline is present, this may be empty.
+        //
+        // \note Up for discussion: If surface line is adjacent to a left lane boundary,
+        // this could contain the left_lane_boundary_id
```
clemenshabedank, line 264: I would say it should be left empty in that case, because the information would just be duplicated. The field `repeated Identifier left_lane_boundary_id = 9;` should be sufficient IMO.

osi_lane.proto
```diff
@@ -192,69 +192,91 @@ message Lane
...
+        repeated Identifier left_adjacent_surfaceline_id = 6;
+
+        // List of IDs of all surfaceline segments that are directly adjacent to the
+        // surfaceline on the right side (w.r.t. ascending order of surfaceline points
+        // and lane boundary points). Note that lengths of surfaceline segments are 
+        // not synchronized and therefore can contain multiple adjacent segments.
+        //
+        // \note If none or just one surfaceline is present, this may be empty.
+        //
+        // \note Up for discussion: If surface line is adjacent to a right lane boundary,
```
clemenshabedank, line 275: Analogue to comment in line 264.

osi_lane.proto
```diff
@@ -192,69 +192,91 @@ message Lane
...
+        // Example: \c #surfaceline_is_centerline = \c true for surfaceline sl2
+        // and \c #surfaceline_is_centerline = \c false for surfacelane sl1 in
         // image \ref HighwayExit .
         //
-        // \note The \c #centerline_is_driving_direction is defined for \c #type
+        // \note The \c #centerline is currently only defined for \c #type
         // = \c #TYPE_DRIVING .
         //
-        optional bool centerline_is_driving_direction = 4;
+        optional bool surfaceline_is_centerline = 5;
```
clemenshabedank, line 254: see comment line 240

osi_surfaceline.proto
```diff
@@ -0,0 +1,169 @@
+syntax = "proto2";
+
+option optimize_for = SPEED;
+
+import "osi_common.proto";
+import "osi_lane.proto";
+
+package osi3;
+
+// A lane's surface line (as a list of segments).
```
thomassedlmayer, line 10: Failing build test: \brief section required for messages with longer description.

osi_surfaceline.proto
```diff
@@ -0,0 +1,169 @@
...
+    //
+    // Defined and used for driving lanes.
+    // \c true means that the surface line is also the centerline of the lane.
+    //
+    // Example: \c #surfaceline_is_centerline = \c true for surfaceline sl2
+    // and \c #surfaceline_is_centerline = \c false for surfacelane sl1 in
+    // image \ref HighwayExit .
+    //
+    // \note The \c #centerline is defined only for \c #type =
+    // \c #TYPE_DRIVING and if exactly one or no \c #lane_pairing pair
```
thomassedlmayer, line 85: Failing build test: Linking fields from other files/messages need some reference to the concerning message. Probably Lane::Classification::type, Lane::Classification::TYPE_DRIVING and Lane::Classification::lane_pairing?

osi_surfaceline.proto
arnedueselder: It seems that \brief is mandatory for any message

osi_surfaceline.proto
arnedueselder: I added the appropriate references

clemenshabedank:
Closing in favor of https://github.com/OpenSimulationInterface/open-simulation-interface/pull/619.

PR #593, opened 2021-10-20
Environmental conditions, added Sun, Wind, CloudState

stmswald:
With this PR we try to fulfill the needs of harmonization of the Issue #544, #577, #578 as well as Issue #255 and the PR #582 
The discussion has been over a year now and we try to merge with small steps.

So far we only added the `Sun`, `CloudState` as well as the `Wind` as messages.
This should help merging this PR in Version V3.5.0. more easily.

There is another PR #601 in order to get the Enum types and continues values for e.g. `Fog`, `Precipitation`, `Ambient Illumination` to V4.0.0. #577 #578 #255 

According to #544 the following data structures were created within this PR:
- Sun:
  - azimuth [rad]
  - elevation [rad]
  - intensity [lx]
- Wind:
   - direction [rad]
   - speed [m/s]
- CloudStates (__enum__):
   - Level 1 - 8 to tell how many clouds cover the blue sky

most of these requirements are form ASAM openSCENARIO

PR #582, opened 2021-09-21
Feature/issue 544, 577, 578 Weather Conditions

m-pauly:
﻿#### Reference to a related issue in the repository
Add a reference to a related issue in the repository.

#### Add a description
Add a description of the changes proposed in the pull request.

**Some questions to ask**:
What is this change?
What does it fix?
Is this a bug fix or a feature? Does it break any existing functionality or force me to update to a new version?
How has it been tested?

#### Take this checklist as orientation for yourself, if this PR is ready for the Change Control Board:
- [ ] My suggestion follows the [style and contributors guidelines](https://opensimulationinterface.github.io/osi-documentation/open-simulation-interface/doc/howtocontribute.html).
- [ ] I have taken care about the [documentation](https://opensimulationinterface.github.io/osi-documentation/open-simulation-interface/doc/commenting.html).
- [ ] I have done the [DCO signoff](https://opensimulationinterface.github.io/osi-documentation/open-simulation-interface/doc/howtocontribute.html#developer-certification-of-origin-dco).
- [ ] My changes generate no errors when passing CI tests. 
- [ ] I have successfully implemented and tested my fix/feature locally.
- [ ] Appropriate reviewer(s) are assigned.

If you can’t check all of them, please explain why.
If all boxes are checked or commented and you have achieved at least one positive review, you can assign the label ReadyForCCBReview!

Issue #578, opened 2021-09-13
Introducing new Fog Model


m-pauly:
Signed-off-by: Markus Waldmann markus.waldmann@stud.hs-kempten.de & Markus Pauly markus.pauly@stud.hs-kempten.de

**Description**

This PR emerged from the Harmonisation group and the discussions related to #544. It should serve as a basis for further discussion. The weather-models were adopted from OSC.

It is not completely clear yet where and how to apply the new message and how to manage the transition to complement or replace existing Fog enums. 

Fog enums could be complemented (short-term) or replaced (long-term) with the new FogDescription. 
The name of the message (FogExtended) is still under discussion.

thempen:
@kmeids Before bringing the PR to ReadyToMerge, we would need some feedback from the SensorModeling group.

thomassedlmayer:
Close because draft is not available anymore.

Issue #577, opened 2021-09-13
Introduction of a additive description for old and new weather conditions


m-pauly:
Signed-off-by: Markus Waldmann markus.waldmann@stud.hs-kempten.de & Markus Pauly markus.pauly@stud.hs-kempten.de

**Description** 

Introduction of the use of **continuous values** instead of enums for several weather-values for example:
- _Clouds:_ OSC & OSI provide 5 states. However, the cloud cover is officially divided into eighth values between 0 and 1: like in the Glossar of the German weather department DWD (dwd.de - https://www.dwd.de/DE/service/lexikon/Functions/glossar.html?lv2=100932&lv3=101016). This values should be adapted to each other.
- _Ambient Illumination_: Describes the Illumination in the environment in LUX, consists of 11 values. Should be described by continuous **double values**
- _Precipitation:_ Contains 9 possible values Continuous values, should be described by **double values** - instead of categories according to the context of ISO 2315
- _Fog Density:_ 10 enum values describe the visibility/density in "m" (meter), should be described by **double values**

thempen:
@kmeids Before bringing the PR to ReadyToMerge, we would need some feedback from the SensorModeling group.

kmeids:
@thempen I only see an open Issue, is there a PR?

FYI:
In the SM group we already had several discussions on this topic and alignments with OSC as well.
We went through several suggestions and initially agreed on one that  @jruebsam is trying to draft the first message structure.
We need to align during our next SM or Harmonization meeting (depending on your and @jruebsam's availability).

adrianschultz:
Within the new OpenSCENARIO project this is for discussion:

```
Definition of cloud states using the fractional cloud cover in oktas.

    International commision on illumination (CIE):
    "Quotient of the sum of the solid angles subtended by clouds and the solid angle 2π sr of the whole sky.
    Note 1 to entry: Observation of total cloud amount is usually recorded in tenths or oktas, where one okta indicates an extent of cloud cover equal to one eighth of the sky.
    Note 2 to entry: The total cloud amount has unit one.     Note 3 to entry: This entry was numbered 845-09-92 in IEC 60050-845:1987.
    Note 4 to entry: This entry was numbered 17-1321 in CIE S 017:2011."


    Oktas   German description      English description                 mapping to deprecated enum
    ---------------------------------------------------------------------------
    0/8     wolkenlos               Cloudless, clear sky, no clouds     free
[...]

PR #601, opened 2021-11-23
Environmental conditions DRAFT for V4.0.0.

stmswald:
With this draft the `osi_environment.proto` gets a major overhaul for OSI V4.0.0.

The basic idea is to move all attributes of the environment to an own message to create new messages easier for all attributes of the environmental conditions.

For now this is done **exemplarily** for the ambient illumination message.
Within this message there is the old Enum type, but it is easier now to create a numerical value range.

This draft is created and handled by the harmonization group.

TODO's:
- [ ] moving all attributes to there own messages
- [x] assignment of numerical value ranges #255 #577 
- [x] assignment of numerical value ranges **AND** Enum types #578 
- [ ] local environment descriptions #311 
- [ ] precipitation type (rain, snow, hail or graupel) #415 
- [x] fog bounding box #544 

Will be for filled by PR #593 and merged into this PR after release of V3.5.0.
- [x] sun description #544 
- [x] cloud state #544 
- [x] wind direction and speed #544 

Related:
- Discussion in #582
- Discussion in #593
- Discussion in #418

### DIFF
--- a/osi_environment.proto
+++ b/osi_environment.proto
@@ -208,97 +208,6 @@ message EnvironmentalConditions
         FOG_DENSE = 9;
     }
 
-    // Definition of discretized ambient illumination states:
-    // Ambient light is any light in the vehicle's environment that is not
-    // emitted by the vehicle itself. It can include sun/moon light, light from
-    // other cars, street lights etc.
-    //
-    // lx ("lux") is the SI unit of luminance or illumination of an area of exact
-    // one square meter, which is equal to one lumen per square meter 1 lx =
-    // 1 lm/m^2 [1].
-    // lm ("lumen") is the SI derived unit of luminous flux and a measure of
-    // the total quantity of visible light emitted by a source. The lumen is
-    // defined in relation to cd ("candela") as 1 lm = 1 cd sr, where sr
-    // denotes steradian, the unit of solid angle used in 3D geometry analogous
-    // to the radian [1].
-    //
-    // Categorization is done based on natural day/night time illuminance levels
-    // [2] and standards for required lighting levels on roads [2, 3, 4, 5].
-    //
-    // \par References:
-    // [1] DIN Deutsches Institut fuer Normung e. V. (1982). <em>DIN 5031-3 Strahlungsphysik im optischen Bereich und Lichttechnik - Groessen, Formelzeichen und Einheiten der Lichttechnik</em>. (DIN 5031-3:1982-03). Berlin, Germany. \n
-    // [2] National Optical Astronomy Observatory. (2015, December 02). <em>Recommended Light Levels</em>. Retrieved January 25, 2020, from https://www.noao.edu/education/QLTkit/ACTIVITY_Documents/Safety/LightLevels_outdoor+indoor.pdf \n
-    // [3] Wang, Y. & Zou, Y., (2016, March). <em>Study on Illumination for State Highways</em>. Washington State Department of Transportation. Retrieved January 25, 2020, from http://www.wsdot.wa.gov/research/reports/fullreports/847.1.pdf \n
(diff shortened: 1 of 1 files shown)


osi_environment.proto
```diff
@@ -324,4 +233,102 @@ message EnvironmentalConditions
...
+
+    //
+    // \brief Definition of the ambient illumination
+    //
+    // \note Ambient light is any light in the vehicle's environment that is not
+    // emitted by the vehicle itself. It can include sun/moon light, light from
+    // other cars, street lights etc.
+    //
+    message AmbientIllumination
+    {
```
caspar-ai, line 245: You're missing the actual field value which uses the enum from this message.

For example `AmbientIlluminationLevel level = 1;`.

osi_environment.proto
```diff
@@ -324,4 +233,102 @@ message EnvironmentalConditions
...
+        // [3] Wang, Y. & Zou, Y., (2016, March). <em>Study on Illumination for State Highways</em>. Washington State Department of Transportation. Retrieved January 25, 2020, from http://www.wsdot.wa.gov/research/reports/fullreports/847.1.pdf \n
+        // [4] Laperriere, A. (2011, May). <em>LED street lighting in the municipality of Saint-Gedeon-de-Beauce within the framework of advanced lighting technologies</em>. Retrieved January 25, 2020, from http://sslnet.ca/wp-content/uploads/2011/10/LTE-RT-2011-0076-Anglais.pdf \n
+        // [5] Crabb, G. I., Beaumont, R. & Webster, D. (2008, October 17). <em>Review of the class and quality of street lighting</em>. Transport Research Laboratory. Retrieved January 25, 2020, from http://courtneystrong.com/wp-content/uploads/2017/07/css-sl1-class-and-quality-of-street-lighting.pdf
+        //
+        enum AmbientIlluminationLevel
+        {
+            // Ambient illumination is unknown (must not be used in ground truth).
+            //
+            AMBIENT_ILLUMINATION_UNKNOWN = 0;
```
thomassedlmayer, line 271: The enum types should be prefixed with the full enum name and include another underscore right before the actual type. So if the enum name is `AmbientIlluminationLevel` it should be `AMBIENT_ILLUMINATION_LEVEL_1`, `AMBIENT_ILLUMINATION_LEVEL_OTHER `etc.

see https://opensimulationinterface.github.io/osi-documentation/#_interface_conventions

osi_environment.proto
stmswald: You are absolutely right.  I haven't added a field yet, because how to name the ENUM field and how to name the field for the numerical value range?

`AmbientIlluminationLevel level = 1`
`double intensity = 2`

osi_environment.proto
stmswald: Right, but then the naming goes like this
`AMBIENT_ILLUMINATION_LEVEL_UNKNOWN = 0`
...
`AMBIENT_ILLUMINATION_LEVEL_LEVEL1`
`AMBIENT_ILLUMINATION_LEVEL_LEVEL2`
sounds wired, how to solve this?

thomassedlmayer: `AMBIENT_ILLUMINATION_LEVEL_1` would not sound too bad, I think. Though, I don't know if we actually need the "AmbientIllumination" term in the types if the enum is located within the `AmbientIllumination` message anyway. Maybe we could name the enum `Level` and have the types `LEVEL_UNKOWN`, `LEVEL_OTHER`, `LEVEL_1`...`LEVEL_9`? It has been done similarly with `Type`, `Material`, etc. in other messages like `StationaryObject`.

yash-shah-asam:
We will close this PR due to no active 4.0 dev group.

PR #615, opened 2022-02-01
Traffic Command Update

clemenshabedank:
This PR comes from the SET Level project.

#### Add a description
Offers a basic backchannel for traffic participant models regarding successful/unsuccessful execution of TrafficCommands. Interpretation is then up to the scenario engine. 

**Some questions to ask**:
What is this change?
What does it fix?
How has it been tested?

### DIFF
--- a/osi_trafficcommand.proto
+++ b/osi_trafficcommand.proto
@@ -65,8 +65,9 @@ message TrafficCommand
 // model that certain actions must or shall be terminated, there are 
 // explicit actions nested inside this message (AbortActionsAction, 
 // EndActionsAction), which hold a reference to the respective actions.
-// It is highly recommended for the scenario engine to handle the state 
-// machine of traffic actions and to use AbortActionsActions / 
+//
+// It is highly recommended that the scenario engine handles the state 
+// machine of traffic actions and uses AbortActionsActions / 
 // EndActionsAction to transfer information to the traffic participant. 
 // Furthermore, the scenario engine shall listen to DismissedActions of 
 // \c TrafficCommandUpdate to reflect this in the evaluation of the 
--- a/osi_trafficcommand.proto
+++ b/osi_trafficcommand.proto
@@ -59,12 +59,18 @@ message TrafficCommand
 // regarding that are not part of this message, yet are seen as a task of the 
 // scenario description, for example, OpenSCENARIO. 
 //
-// \note All traffic actions are sent only once just before they are about 
+// \attention All traffic actions are sent only once just before they are about 
 // to start. This is also true, if their execution is expected to 
 // take simulation time. To inform the traffic participant 
 // model that certain actions must or shall be terminated, there are 
 // explicit actions nested inside this message (AbortActionsAction, 
 // EndActionsAction), which hold a reference to the respective actions.
+// It is highly recommended for the scenario engine to handle the state 
+// machine of traffic actions and to use AbortActionsActions / 
+// EndActionsAction to transfer information to the traffic participant. 
+// Furthermore, the scenario engine shall listen to DismissedActions of 
+// \c TrafficCommandUpdate to reflect this in the evaluation of the 
+// scenario run. 
 //
 message TrafficAction
 {
--- a/osi_trafficcommand.proto
+++ b/osi_trafficcommand.proto
@@ -59,19 +59,14 @@ message TrafficCommand
 // regarding that are not part of this message, yet are seen as a task of the 
 // scenario description, for example, OpenSCENARIO. 
 //
-// \attention All traffic actions are sent only once just before they are about 
+// \note All traffic actions are sent only once just before they are about 
 // to start. This is also true, if their execution is expected to 
 // take simulation time. To inform the traffic participant 
 // model that certain actions must or shall be terminated, there are 
 // explicit actions nested inside this message (AbortActionsAction, 
 // EndActionsAction), which hold a reference to the respective actions.
-//
-// It is highly recommended that the scenario engine handles the state 
-// machine of traffic actions and uses AbortActionsActions / 
-// EndActionsAction to transfer information to the traffic participant. 
-// Furthermore, the scenario engine shall listen to DismissedActions of 
-// \c TrafficCommandUpdate to reflect this in the evaluation of the 
-// scenario run. 
+// Futhermore, there exists a \c TrafficCommandUpdate message for the 
+// traffic participant to report back on potentially dismissed actions. 
 //
 message TrafficAction
 {
--- a/osi_trafficcommandupdate.proto
+++ b/osi_trafficcommandupdate.proto
@@ -17,11 +17,9 @@ package osi3;
 // This message allows a traffic participant to send feedback if an action 
 // cannot happen as requested by the \c TrafficCommand. Currently, it is out of
 // scope to standardize the exact reason for non-executability or failed execution 
-// because the reason can have multiple explantions. The point in time 
-// for this message to be sent is only restricted to be after (or at the same time) 
-// the \c TrafficCommand with the corresponding traffic action(s) has been sent. The 
-// responsibility for deciding about successful or unsuccessful scenario execution
-// lies fully on the side of the scenario engine.  
+// because the reason can have multiple explanations. The responsibility for deciding 
+// about successful or unsuccessful scenario execution lies fully on the side of the 
+// scenario engine.  
 //
 // \note This interface is currently just a placeholder and could be
 // changed in experimental ways to support semantics of upcoming OpenSCENARIO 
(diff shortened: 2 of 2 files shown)


osi_trafficcommand.proto
```diff
@@ -59,12 +59,18 @@ message TrafficCommand
...
 // scenario description, for example, OpenSCENARIO. 
 //
-// \note All traffic actions are sent only once just before they are about 
+// \attention All traffic actions are sent only once just before they are about 
 // to start. This is also true, if their execution is expected to 
 // take simulation time. To inform the traffic participant 
 // model that certain actions must or shall be terminated, there are 
 // explicit actions nested inside this message (AbortActionsAction, 
 // EndActionsAction), which hold a reference to the respective actions.
+// It is highly recommended for the scenario engine to handle the state
```
caspar-ai, line 68:
```suggestion
//
// It is highly recommended that the scenario engine handles the state 
```

osi_trafficcommand.proto
```diff
@@ -59,12 +59,18 @@ message TrafficCommand
...
 //
-// \note All traffic actions are sent only once just before they are about 
+// \attention All traffic actions are sent only once just before they are about 
 // to start. This is also true, if their execution is expected to 
 // take simulation time. To inform the traffic participant 
 // model that certain actions must or shall be terminated, there are 
 // explicit actions nested inside this message (AbortActionsAction, 
 // EndActionsAction), which hold a reference to the respective actions.
+// It is highly recommended for the scenario engine to handle the state 
+// machine of traffic actions and to use AbortActionsActions /
```
caspar-ai, line 69:
```suggestion
// machine of traffic actions and uses AbortActionsActions / 
```

osi_trafficcommandupdate.proto
```diff
@@ -0,0 +1,74 @@
...
+
+import "osi_version.proto";
+import "osi_common.proto";
+
+package osi3;
+
+//
+// \brief This message enables the traffic participant model to send updates 
+// to the scenario engine about the execution of its received \c TrafficCommand input. 
+// While traffic actions are usually executed successfully by the traffic participant
```
caspar-ai, line 13:
```suggestion
// While traffic actions are usually executed successfully by the traffic participant,
```

osi_trafficcommandupdate.proto
```diff
@@ -0,0 +1,74 @@
...
+import "osi_common.proto";
+
+package osi3;
+
+//
+// \brief This message enables the traffic participant model to send updates 
+// to the scenario engine about the execution of its received \c TrafficCommand input. 
+// While traffic actions are usually executed successfully by the traffic participant 
+// there may be actions which the traffic participant is not able to execute
+// either for capability or situation-specific reasons. This message gives
```
caspar-ai, line 15: I would suggest starting a new paragraph with the sentence that starts "This message..." to break up the text a little.

I would also suggest this wording for the start of the first sentence:

"This message allows a traffic participant to send feedback if..."

osi_trafficcommandupdate.proto
```diff
@@ -0,0 +1,74 @@
...
+// the traffic participant the basic possiblity to send feedback if an action 
+// cannot happen as requested by the \c TrafficCommand. Currently, it is out of
+// scope to standardize the exact reason for non-executability or failed execution 
+// because the reason can have multiple explantions. The point in time 
+// for this message to be sent is only restricted to be after (or at the same time) 
+// the \c TrafficCommand with the corresponding traffic action(s) has been sent. The 
+// responsibility for deciding about successful or unsuccessful scenario execution
+// lies fully on the side of the scenario engine.  
+//
+// \note This interface is currently just a placeholder and could be
```
caspar-ai, line 25: How should a user of OSI interpret this?

Is it meant to suggest that this message is more likely than others to be changed in a breaking way in a future major release?

osi_trafficcommandupdate.proto
```diff
@@ -0,0 +1,74 @@
...
+    optional Timestamp timestamp = 2;
+
+    // The ID of this traffic participant which must coincide with a prior sent ID, cf.
+    // \c TrafficCommand::traffic_participant_id.
+    //
+    optional Identifier traffic_participant_id = 3;
+
+    // Actions which a traffic participant dismisses.
+    //
+    // \note If more than one action id is supplied it means that
```
caspar-ai, line 49: I'm not sure if you need this `\note`, partly because I think the explanation is the obvious one that can be drawn from the message definition, and partly because it is referencing a field inside the child structure which doesn't feel "right".

If you think this is worth highlighting, then I would suggest you change this to read "If more than one dismissed action is supplied it means that..."

osi_trafficcommandupdate.proto
```diff
@@ -0,0 +1,74 @@
...
+    // multiple actions are regarded as dismissed.
+    //
+    repeated DismissedAction dismissed_action = 4;
+    
+    // 
+    // \brief Action which a traffic participant dismisses.
+    //
+    message DismissedAction 
+    {
+        // Dismissed traffic action id from the perspective of a traffic participant, if any.
```
caspar-ai, line 59: What does the "if any" on the end imply?

Also, I would have expected "Dismissed traffic action id from the perspective of _this_ traffic participant" - as in this a traffic participant should only be dismissing actions it (and given `TrafficCommand` is sent to a single participant) and only it, was sent.

osi_trafficcommandupdate.proto
```diff
@@ -0,0 +1,74 @@
...
+        // \note A dismissed traffic action id identifies a \c TrafficAction which cannot be executed or 
+        // completed by the traffic participant. This field must have the same value as 
+        // \c TrafficAction::ActionHeader::action_id of a prior sent \c TrafficCommand and must correspond
+        // to the same traffic participant.
+        //
+        optional Identifier dismissed_action_id = 1;
+
+        // Information about the reason of failure.
+        // 
+        // \note This is just a custom, informal string without a standardized meaning.
```
caspar-ai, line 70:
```suggestion
        // \note This is a custom, informal string without a standardized meaning. 
```

osi_trafficcommandupdate.proto
```diff
@@ -0,0 +1,74 @@
...
+    message DismissedAction 
+    {
+        // Dismissed traffic action id from the perspective of a traffic participant, if any.
+        //
+        // \note A dismissed traffic action id identifies a \c TrafficAction which cannot be executed or 
+        // completed by the traffic participant. This field must have the same value as 
+        // \c TrafficAction::ActionHeader::action_id of a prior sent \c TrafficCommand and must correspond
+        // to the same traffic participant.
+        //
+        optional Identifier dismissed_action_id = 1;
```
caspar-ai, line 66:
```suggestion
        optional Identifier action_id = 1;
```

Minor, don't feel strongly.  Only suggesting because the field is already inside a structure with the word `Dismissed` so it would be `DismissedAction.dismissed_action_id` which introduces some redundancy.

(there is even an argument to just make this field `id`, but that I'm happy with `action_id` as that mirrors what is done for a `TrafficCommand`)

caspar-ai (commented):
Thanks for the PR, looks good to me!

I think the fields and structure look good.  I've added a few comments questions and suggestions but happy with the proposal.

osi_trafficcommandupdate.proto
clemenshabedank: Done.

osi_trafficcommandupdate.proto
clemenshabedank: To your second question: I think, yes. A main motivation for TrafficCommand and TrafficCommandUpdate was to be well compatible to OpenSCENARIO semantically which is changing a lot, currently (OpenSCENARIO 1.x and 2.x, but also each stream itself). I think this is why we have a similar comment in https://github.com/OpenSimulationInterface/open-simulation-interface/blob/master/osi_trafficcommand.proto lines 15-18. This should probably be discussed more broadly in a meeting with OpenSCENARIO liaisons and fixed in a major OSI version.

osi_trafficcommandupdate.proto
clemenshabedank: I think highlighting is good. Changed it according to your suggestion.

osi_trafficcommandupdate.proto
clemenshabedank: If you don't feel strongly about it, I would leave it as is, because SETLevel implementations would potentially need to adapt.

osi_trafficcommandupdate.proto
caspar-ai: OK, thanks.  Assuming this isn't intending to imply that the interface may change in a breaking way as part of a minor release I'm happy to leave as-is.

(if we do want that flexibility, then I think it is worth calling that out very explicitly)

stefancyliax:
OSI CCB 16.03.2022:
- topic has to wait until @clemenshabedank can take part in the CCB. Can you @clemenshabedank take part on the 30.03.?
- Addition to osi_trafficcommand.proto has to be evaluated and be discussed. What is the normative intent of this addition?

clemenshabedank:
I will attend in today's meeting, so we can discuss the open points and if there need to be changes.

clemenshabedank:
I will attend in today's meeting, so we can discuss the open points and if there need to be changes.

kmeids:
**CCB  output 13.04.2022:**
1. @clemenshabedank Please update the description for _message DismissedAction_ to better indicate the intended usage of this message.

clemenshabedank:
Hi @ThomasNaderBMW please have a look if [this commit ](https://github.com/OpenSimulationInterface/open-simulation-interface/pull/615/commits/38e8a48c54ea4b09a18342cd33bc6bef1e9f9585) suits your expectations about the semantics of dismissed actions.

kmeids:
**CCB output 09.05.2022:**
1. @clemenshabedank ready to merge.

pmai (approved):

PR #617, opened 2022-02-11
Dependebility: Bumb protobuf version

jdsika:
@pmai there is a security issue with protobuffer that created an automatic version bumb PR in [osi-validation](https://github.com/OpenSimulationInterface/osi-validation/pull/41). I am wondering three things here:
1) Why not bumb to the newest version [3.19.4](https://github.com/protocolbuffers/protobuf/releases)?
2) Where are the places we have defined the protobuffer version in this repository?
3) Do we need to create a new patch version for OSI?

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Install `pip3` and missing python packages:
 ```bash
 $ sudo apt-get install python3-pip python3-setuptools
 ```
-Install `protobuf` 3.0.0:
+Install `protobuf`:
 ```bash
 $ sudo apt-get install libprotobuf-dev protobuf-compiler
 ```


pmai:
Contrary to osi-validation, which used a fixed version of protobuf (for unknown reasons, not recommended anyway), we do not specify a version of protobuf, or only specify a minimum version due to required features.

Furthermore OSI is not intended to be used with untrusted sources for many other reasons anyway; for that reason I do not think we should change anything here, except maybe dropping the version in the README entirely.

We are not in a position to force the use of a version of protobuf for unrelated, and often unnecessary reasons.

For osi-validation I have changed the requirement to be a minimum version, and moved to 3.15.0 since one could argue that the validator should be more robust.

jdsika:
So, two options:

1) remove the version and just say "protobuffer"
2) recommend the same minimum version as in osi-validation

pmai:
I'd only go for option 1; for osi-validation there was already a version tie, so the proper course was to update, given that osi-validation is a validator. For OSI itself I would not talk about versions beyond what we currently have: There is little attack surface in most settings, and there are prescribed versions due to the overall build processes (e.g. need for same version across subsystems), so that all users already have to deal with that themselves.

We'll see how the CCB feels about this, but that would be my opinion.

jdsika:
Then we have the discussion and facts for a decision in the CCB. I have labeled the PR accordingly

kmeids:
**Output CCB 16.02.22:**
1. @pmai please merge, taking into consideration not to specify any specific protobuf version in the README file.

pmai (approved):

stefancyliax (commented):
Approve

Issue #645, opened 2022-06-28
Calculation of s-t coordinates using t-axis definition


weissdavid:
#### Describe the feature
The current definition of s-t coordinates relative to a reference line has two issues on the corners (multiple points with same coordinates on the outside, discontinuity on the inside). I would like to suggest a modification of the calculation that solves this problems and gives s-t coordinates that are closer to OpenDRIVE.

#### Describe the solution you would like
For each ReferenceLinePoint a new field defining the t-axis (i.e. all points with the same s coordinate) is introduced. Points outside the reference line or then projected onto the reference line not orthogonally (as in the current definition) but by drawing the straight line of the point to the intersection of the t-axes of the appropriate section (see sketch). s and t coordinate are then calculated as in the current definition (s is the s coordinate of the projection point, t is the euclidean distance to the projection point).
![ReferenceLineSuggestion](https://user-images.githubusercontent.com/108330371/176136786-8b20b297-00fe-4986-8cf4-d0aab525ca7d.svg)

#### Describe the backwards compatibility
If the t-axis is not defined, one could instead use the angle bisector for the same calculation.

#### Additional context
For a possible implementation of this calculation see 
[Proposal calculation s t.docx](https://github.com/OpenSimulationInterface/open-simulation-interface/files/8999323/Proposal.calculation.s.t.docx)

ReinhardBiegelIntech:
Looks like this is already addressed in v3.6.0 via #651?

PR #651, opened 2022-07-08
Adapt ReferenceLine ST definition to avoid transformation ambiguity and discontinuities

ReinhardBiegelIntech:
This addresses #645 

Current definition of s,t coordinates has some (well documented) downsides, which we would like to mitigate with this proposed change. With the updated s,t definition, x,y <-> s,t associations are unique inside the lane geometries and there are no gaps in the s coordinate anymore.

This approach is used and tested by [openPASS ](https://gitlab.eclipse.org/eclipse/simopenpass/simopenpass).

Checks will follow (but the links to `howtocontribute.html` seem to be dead... where to get this information?)

- [ ] My suggestion follows the [style and contributors guidelines](https://opensimulationinterface.github.io/osi-documentation/open-simulation-interface/doc/howtocontribute.html).

### DIFF
--- a/osi_referenceline.proto
+++ b/osi_referenceline.proto
@@ -91,8 +91,11 @@ message ReferenceLine
     // and a projected point (P_proj) on the polyline. The T axis (projecting
     // axis) is the line going through P and the intersection point (I). I is
     // defined as the intersection of both T axes of two consecutive
-    // ReferenceLinePoints (see example and image below for illustration). The T
-    // coordinate of the point in question is then defined as
+    // ReferenceLinePoints (see example and image below for illustration). If
+    // both T axes of the neighboring ReferenceLinePoint are parallel (so no
+    // intersection point exists), the resulting T axis direction is equal to
+    // the T axis of these ReferenceLinePoints.
+    // The T coordinate of the point in question is then defined as
     // <code>hypot(P.X-P_proj.X,P.Y-P_proj.Y)</code>. The projected point P_proj
     // might either be on a line segment or at an edge between two line segments.
     // The distance is positive if the point is left of the polyline (in
@@ -129,13 +132,14 @@ message ReferenceLine
     // and two points (P1 and P2) not part of the reference line.
     //
     // Calculation of ST for P1:
-    // - Calculate the instersection point I of the T axes of R0 and R1.
+    // - Calculate the intersection point I of the T axes of R0 and R1.
     // - As P1 lies in the sector defined by these T axes it is considered part
     //   of the reference line section between R0 and R1.
     // - The point P1 is projected onto the line segment [R0, R1] via the
     //   straight line through I (by calculating the intersection of the line
     //   segment and the projection axis), resulting in point P1_proj.
-    //   If the T axes are parallel, a simple orthogonal projection is used.
+    //   If the T axes are parallel, projection is applied in the direction of
+    //   these axes.
     // - The S coordinate of P1 is the S coordinate of P1_proj
     // - The T coordinate of P1 is the signed Euclidean distance to P1_proj.
     //
--- a/osi_referenceline.proto
+++ b/osi_referenceline.proto
@@ -37,7 +37,7 @@ message ReferenceLine
 
     // ReferenceLine types
     //
-    // ReferenceLinePoints might be interpreted diffrently depending on the type
+    // ReferenceLinePoints might be interpreted differently depending on the type
     // of the ReferenceLine.
     //
     // See also: "Adding T coordinates"
@@ -47,19 +47,25 @@ message ReferenceLine
         // ReferenceLine is a polyline, where the coordinates of points are
         // calculated by projection onto the nearest point on the line.
         //
+        // \attention DEPRECATED: Due to the shortcomings documenten below, this
+        //            type will be removed in 4.0.0.
+        //
         TYPE_POLYLINE = 0;
 
         // ReferenceLine is a polyline, where the coordinates of points are
-        // calculated using the t axis definition.
+        // calculated using the T axis definition.
+        //
+        // \note If this type is used, ReferenceLinePoint::t_axis_yaw must be set.
         //
         TYPE_POLYLINE_WITH_T_AXIS = 1;
     }
 
     // Points comprising the polyline.
     //
-    // At least 2 points must be given.
+    // At least two points must be given.
     // The polyline is defined as the lines between consecutive points.
-    // Each point has an S coordinate.
+    // Each point has an S coordinate. Other attributes might be set, depending
+    // on the type of the polyline (see Type).
     //
     // ## Rules on the S position
     //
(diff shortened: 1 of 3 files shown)


ArunDas1:
@tbleher @ThomasNaderBMW @jdsika Here is our proposal regarding the discussed issue with computation of s-t-coordinates.

osi_referenceline.proto
```diff
@@ -87,31 +87,33 @@ message ReferenceLine
...
-    // Notes:
-    // - The "nearest point on the polyline" is determined in 3D (even if the
-    //   resulting T value is only the 2D distance), in order to choose the
-    //   correct point for 3D curves (think reference lines for roads in parking
-    //   decks).
-    // - If there are several "nearest points", the one with the smallest S
-    //   coordinate on the polyline is chosen.
+    // is added. T is the signed 2D distance between the point to describe (P)
+    // and a projected point (P_proj) on the polyline. The T axis (projecting
+    // axis) is the line going through P and the intersection point (I). I is
```
tbleher, line 92: What happens if there is no intersection point?

osi_referenceline.proto
```diff
@@ -121,23 +123,23 @@ message ReferenceLine
...
-    //   (drawn in grey).
-    // - Finally, \c P5 shows how the reference line is extended infinitely for
-    //   points that are "outside" the reference line.
+    // \image html OSI_ReferenceLine1.svg "S, T calculation"
+    //
+    // This shows a reference line (consisting of three points R0, R1 and R2)
+    // and two points (P1 and P2) not part of the reference line.
+    //
+    // Calculation of ST for P1:
+    // - Calculate the instersection point I of the T axes of R0 and R1.
```
tbleher, line 132: s/instersection/intersection/

osi_referenceline.proto
ReinhardBiegelIntech: Short internal discussion showed, that orthogonal projection isn't 100% correct. To wrap it up:
- There is no intersection point I, if two ReferenceLinePoint neighbors have the same t_axis_yaw
- This t_axis_yaw has to be used as projection direction

We'll update the description accordingly.

osi_referenceline.proto
```diff
@@ -202,6 +204,14 @@ message ReferenceLine
         // S position on the reference line
         //
         optional double s_position = 2;
+
+        // Yaw angle of the T axis in the world coordinate system
+        //
+        // When converting from formats like OpenDRIVE, the yaw angle is equal to
```
tbleher, line 210: I guess it doesn't matter whether the normal goes to the left or to the right - correct?

That also means that "there is no I because two consecutive t_axis_yaw are the same" can also happen if the two angles differ by 180 degree (or PI, since this is stored in radians). Correct?

osi_referenceline.proto
```diff
@@ -121,23 +123,23 @@ message ReferenceLine
...
-    // - Finally, \c P5 shows how the reference line is extended infinitely for
-    //   points that are "outside" the reference line.
+    // \image html OSI_ReferenceLine1.svg "S, T calculation"
+    //
+    // This shows a reference line (consisting of three points R0, R1 and R2)
+    // and two points (P1 and P2) not part of the reference line.
+    //
+    // Calculation of ST for P1:
+    // - Calculate the instersection point I of the T axes of R0 and R1.
+    // - As P1 lies in the sector defined by these T axes it is considered part
```
tbleher, line 133: This change removes the \c formatting markers. I found these helpful for readability when reading the rendered document (as will be displayed on the website). I suggest you check how it looks in Doxygen, and probably re-add the \c formatting markers.

osi_referenceline.proto
weissdavid: Correct

osi_referenceline.proto
ReinhardBiegelIntech: Agree, these should be added again.

tbleher:
Change looks good to me.

ReinhardBiegelIntech:
As discussed in the ASAM OSI Other Models Workgroup meeting on Monday, we updated the PR to clarify the constraints on the T axis.

lemmer-fzi:
After reading the documentation and having a look at the provided diagrams i have a question:
Consider a reference line that only has two points (straight line). How would you calculate the (s,t) coordinates for a point that lies on a line perpendicular to the last point of the segment. As far as i can, using the proposed definition of the t axis their is no way to define a t-axis that crosses the point and still maps it to a valid s-coordinate (a value of s smaller than the maximal value specified by the last point on the reference line).
In large, this question also applies to the last segment of a specified reference line. Could you maybe elaborate on this?

weissdavid:
> After reading the documentation and having a look at the provided diagrams i have a question: Consider a reference line that only has two points (straight line). How would you calculate the (s,t) coordinates for a point that lies on a line perpendicular to the last point of the segment. As far as i can, using the proposed definition of the t axis their is no way to define a t-axis that crosses the point and still maps it to a valid s-coordinate (a value of s smaller than the maximal value specified by the last point on the reference line). In large, this question also applies to the last segment of a specified reference line. Could you maybe elaborate on this?

You can always define the t-axis to be the line perpendincular to the segment. In this case your point, which is perpendicular to the last point lies exactly on this t-axis and therefore has the same s coordinate (the t-axis consists of all points with equal s coordinate). Furthermore all s coordinates are valid (even negative values).

ReinhardBiegelIntech:
@ThomasNaderBMW
We updated the PR as discussed last week (addition of a reference line type enum).

osi_referenceline.proto
```diff
@@ -31,6 +31,30 @@ message ReferenceLine
     //
     optional Identifier id = 1;
 
+    // The type of the reference line.
+    //
+    optional Type type = 3;
+
+    // ReferenceLine types
+    //
+    // ReferenceLinePoints might be interpreted diffrently depending on the type
```
tbleher, line 40: s/diffrently/differently/

osi_referenceline.proto
```diff
@@ -87,31 +111,57 @@ message ReferenceLine
...
-    //   decks).
-    // - If there are several "nearest points", the one with the smallest S
-    //   coordinate on the polyline is chosen.
+    // is added. T is the signed 2D distance between the point to describe (P)
+    // and a projected point (P_proj) on the polyline. There are two ways of
+    // defining this point, depending on the ReferenceLine type.
+    //
+    // ## Nearest point (TYPE_POLYLINE)
+    //
+    // The projection point is the nearest point on the polyline (this point might
```
tbleher, line 120: This seems far too short (e.g. it doesn't mention which side is positive). I guess the goal here is to keep backward compatibility. Then either the old documentation should be kept, or common parts between the two modes should be in a common place of the documentation. E.g. now only TYPE_POLYLINE_WITH_T_AXIS mentions which side is positive, but this applies to both modes.

osi_referenceline.proto
```diff
@@ -87,31 +111,57 @@ message ReferenceLine
...
+    // The T coordinate of the point in question is then defined as
+    // <code>hypot(P.X-P_proj.X,P.Y-P_proj.Y)</code>. The projected point P_proj
+    // might either be on a line segment or at an edge between two line segments.
+    // The distance is positive if the point is left of the polyline (in
+    // definition direction), negative if it is right of it. The S position of
+    // such a point outside the reference line is the same as the S value of the
+    // projected point on the polyline, resulting in all points on a single T
+    // axis having the same S coordinate.
+    //
+    // \note The ST coordinate system shall be defined in 2D only. When
```
tbleher, line 140: Note that this explicitly is different for the old mode (and thus should be mentioned). There, the nearest point is determined in 3D, even if the distance is 2D. I think this makes much more sense, e.g. if you have a spiraling upward road in a multi-storey parking lot, you really do want to get the nearest point in 3D. Otherwise the S position will make no sense. So I think the question is whether you really want to define this only in 2D. See the previous definition on line 100.

osi_referenceline.proto
```diff
@@ -121,23 +171,24 @@ message ReferenceLine
...
-    // - \c P2 has multiple points "nearest points" on the polyline.
-    //   As can be seen here, two  ST coordinates map to \c P2 (red and grey
-    //   dotted line).  Following the rules above, the one with the smallest S
-    //   value is chosen (the red dotted line).
-    // - \c P3 has a unique "nearest point" on the polyline. However, multiple
-    //   points map to the same ST coordinate as that of \c P3, e.g. \c P4
-    //   (drawn in grey).
-    // - Finally, \c P5 shows how the reference line is extended infinitely for
-    //   points that are "outside" the reference line.
+    // \image html OSI_ReferenceLine1.svg "S, T calculation"
```
tbleher, line 174: This now references the new image. But if the old mode is to be kept, then the old image also needs to stay, I think (plus the documentation that accompanied the image). Otherwise it is unclear how to implement the old mode, and how to interpret it.

tbleher (commented):
I didn't have time to participate in the discussion, so the following review is purely based on reading the patchset. This seems fine for keeping backward compatibility (but also means that all the old documentation will have to be kept), but also means that for agent models, it now becomes MUCH harder to support this, since they will have to implement TWO modes for ST calculations instead of one (simulation environments need only support one, but agent models need to support two). So I don't really like this idea.

Personally, I would be OK with breaking backward compatibility here (considering this is a very new feature, and probably noone has implemented it fully), but I don't know if this will gain consensus.

If the two modes are kept, then please at least mark one of the modes as deprecated, and advise agent models to only support the newer modes. If some agent models only support the first mode, and some only support the second mode, then no interoperability is achieved!

weissdavid:
[OSI_s-t_calculation.pdf](https://github.com/OpenSimulationInterface/open-simulation-interface/files/9950660/OSI_s-t_calculation.pdf)
As discussed in the CCB meeting, we prepared an exemplary calculation.

osi_referenceline.proto
ReinhardBiegelIntech: re-added old documentation

osi_referenceline.proto
ReinhardBiegelIntech: removed 2D restriction again

thempen:
From WG Harmonization: Please add a hint at the beginning, that the ST coordinate definition should not be misunderstood as the ST definition in OpenDRIVE and OpenSCENARIO 1.x.

The current proto definition for the enum fields "other" and "unknown" are missing, so it does not follow the coding convention guidelines: https://opensimulationinterface.github.io/osi-documentation/#_contributing_changes 
To be determined by CCB if this is an exceptional situation and can be accepted.

pmai:
OSI CCB 2022-11-21: Can be merged as is with necessary changes to resolve technical conflicts. Merge and cleanup done by @pmai.

pmai (approved):

PR #669, opened 2022-10-11
Fix path to asciidoc-resources

philipwindecker:
Quick Fix for Asciidoctor build


philipwindecker:
This hotfix is related to the build pipeline of https://github.com/OpenSimulationInterface/osi-documentation and has no content impact what-so-ever.
Will merge this without review.

PR #685, opened 2022-11-07
Add new message for BoundingBox sub-sections

ndunningbmw:

#### Add a description
This PR proposes the addition of bounding box sub-sections to BaseMoving and BaseStationary objects, to help sub-divide and classify different parts of the overall object structure.

**Some questions to ask**:
What is this change?
What does it fix?
How has it been tested?

- [ ] I have done the [DCO signoff](https://opensimulationinterface.github.io/osi-documentation/open-simulation-interface/doc/howtocontribute.html#developer-certification-of-origin-dco).


osi_common.proto
```diff
@@ -387,6 +387,68 @@ message LogicalLaneAssignment
...
+    {
+        // General stuff
+        TYPE_UNKNOWN = 0;
+        TYPE_OTHER = 1;
+
+        // Car stuff
+        TYPE_CHASIS = 2;
+        TYPE_DOOR = 3;
+        TYPE_SIDEMIRROR = 4;
+        TYPE_CARGO = 5;
```
ndunningbmw, line 440: Should `TYPE_WHEEL` also be added?

osi_common.proto
```diff
@@ -569,6 +631,12 @@ message BaseMoving
     // The polygon is defined counter-clockwise.
     //
     repeated Vector2d base_polygon = 7;
+    
+    // Sub-divisions of the overall bounding box of the BaseMoving object.
+    //
+    // \note The bounding box sections can include side mirrors for vehicles.
```
ndunningbmw, line 637:
```suggestion
    // \note The bounding box sections can include side mirrors, doors, etc. for vehicles.
```

ThomasNaderBMW:
Hi @tbleher , hi Jakob Peintner,

would you also like to give a review on this PR?

osi_common.proto
thomassedlmayer: I think @ThomasNaderBMW has made some more notes on this but before we forget the outcome of the meeting (21th Nov: WG Other Models):
We should add TYPE_WHEEL but state that more wheel information can be added in the MovingObject/VehicleAttributes/WheelData struct if desired (even though this means kind of redundant information).

We also wanted to have clarified how sub bounding boxes are contained in the main bounding box. As per current OSI definition, side mirrors are not included in the main bounding box but it would probably make more sense to include all sub bounding boxes in the main bounding box as the name suggests.

osi_common.proto
```diff
@@ -387,6 +387,68 @@ message LogicalLaneAssignment
...
+        TYPE_SIDEMIRROR = 4;
+        TYPE_CARGO = 5;
+
+        // People/animal stuff
+        // TODO: Those interested in pedestrian modeling need to add their input here
+        TYPE_TORSO = 6;
+        TYPE_LIMB = 7;
+        TYPE_HEAD = 8;
+
+        // TODO: Do we want to handle data for BaseStationary here as well? What problems would having this solve?
```
thomassedlmayer, line 448: Outcome WG Other Models (21th Nov): We want to handle the most significant cases for stationary objects here as well (e.g. tree trunk, tree crown, ...).

osi_common.proto
```diff
@@ -569,6 +687,19 @@ message BaseMoving
...
+    // Sub-divisions of the overall bounding box of the \c BaseMoving object.
+    //
+    // The bounding box sections can include side mirrors, cargo, etc. for 
+    // vehicles, as well as body-part sections for pedestrians.
+    //
+    // \note When one or more \c BoundingBox s are associated to a 
+    // \c BaseMoving , the expectation is that all sections are contained
+    // within the bounds of the \c #dimension . Currently there several accepted
+    // exceptions to this rule, specifically the sub-sections for cargo, vehicle
+    // side mirrors, and doors not in the their closed position.
```
ndunningbmw, line 700: Proposal: Move the text for the "exceptional cases" to the enum field where they would be used from.

osi_common.proto
```diff
@@ -387,6 +387,113 @@ message LogicalLaneAssignment
...
+        TYPE_SIDE_MIRROR = 4;
+        
+        // Additional cargo attached to the a vehicle which is temporarily
+        // attached.
+        // 
+        TYPE_CARGO = 5;
+        
+        // The wheel of a vehicle.
+        // 
+        TYPE_WHEEL = 6;
```
ndunningbmw, line 465: Proposal: Mention the data encoded in MovingObject/VehicleAttributes/WheelData and in which case this one should be used, and how the 2 fields are different

tbleher:
> Hi @tbleher , hi Jakob Peintner,
> 
> would you also like to give a review on this PR?

I think the initative is a good one :)
I can only really talk about bounding boxes of static objects (which is what I'm currently looking at for the OSIPedestrian model): it's very good for e.g. pedestrian models if tree trunk and tree arm can be separated (and similarly light pole and light arm). However, I have two main concerns with the proposal right now:
- How would this be filled in from OpenDRIVE? A static object in OpenDRIVE can have an arbitrary number of bounding boxes, so one would expect that a tree has two bounding boxes. But there is no labeling of the bounding boxes. So how would one know which one is which? (Of course, I can add a heuristic to the code, but most implementations will probably just make it TYPE_OTHER, since OpenDRIVE doesn't have these labels).
- I see no language in the proposal that the object is fully contained in the sub-bounding boxes. So AFAICS there is no assurance that if I have a tree with sub-bounding boxes, that the tree is only in that sub-area, and the pedestrian can walk freely in areas that are covered by the parent dimension, but is not filled by the specific bounding boxes. Without this guarantee, the sub-bounding boxes are useless for the purpose of pedestrian models.

So, speaking as someone creating agent models, it would be much more important to have the ability to segment an object such that one has the guarantee that some areas of the dimension are in fact free. For such use-cases, the exact type does not matter at all, and could be left out. Is there any user who cares about this (the type distinction) for static objects like trees and traffic lights? I can see how this is important for wheels vs mirrors and such, but not so much for crown vs trunk - I just want to have a better 3D approximation of the tree so the pedestrian can accurately walk around it.

PR #684, opened 2022-11-04
Fix image_data comment on pixel order

thomassedlmayer:
#### Description
We noticed that the pixel order of the image_data field in CameraSensorView was not specified by CameraSensorViewConfiguration as the description of image_data says:
> The raw image data in the memory layout and order specified by the camera sensor input configuration.

Since both radar and lidar reflections are currently defined to be ordered left-to-right, top-to-bottom (see following quote), it could be the case that this was accidently omitted for image_data. ([RadarSensorView](https://opensimulationinterface.github.io/open-simulation-interface/structosi3_1_1RadarSensorView.html), [LidarSensorView](https://opensimulationinterface.github.io/open-simulation-interface/structosi3_1_1LidarSensorView.html))
> This field includes one entry for each ray, in left-to-right, top-to-bottom order (think of scan lines in a TV). 

Therefore, the default pixel order should probably align with the definitions of radar and lider reflections. If further pixel orders are desired a enum could be added to CameraSensorViewConfiguration but I think this small change would already clarify enough to be able to reliably exchange image data with OSI.

This change can be seen as a bug fix. The description of image_data that a pixel order can be specified in CameraSensorViewConfiguration is definitely misleading.

#### Checklist
- [x] My suggestion follows the [style and contributors guidelines](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/specification/contributing/start_contributing.html).
- [x] I have taken care about the [message documentation](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/specification/contributing/commenting_messages.html) and the [fields and enums documentation](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/specification/contributing/commenting_fields_enums.html).
- [x] I have done the [DCO signoff](https://opensimulationinterface.github.io/osi-documentation/open-simulation-interface/doc/howtocontribute.html#developer-certification-of-origin-dco).
- [ ] My changes generate no errors when passing CI tests. 
- [x] I have successfully implemented and tested my fix/feature locally.
- [x] Appropriate reviewer(s) are assigned.

﻿#### References
#681

Issue #620, opened 2022-03-16
Add orientation indicator to CameraSensorViewConfiguration (flip/rotate)


RIFJo:
#### Describe the feature
The raw image data provided by some sensors needs to be treated as mirrored and/or rotated before further processing or display on a screen. Real-world cameras use the EXIF Orientation tag to record this information (https://sirv.com/help/articles/rotate-photos-to-be-upright/#exif-orientation-values). This hint cannot currently be represented in OSI, so generators of `CameraSensorView `messages are forced to reorganize the pixel values in `CameraSensorView::image_data` to match the sensor coordinate system to prevent confusion about how to interpret the image data for receivers. 
Instead of forcing this byte-reordering step, The `CameraSensorViewConfiguration `should have a way to preserve this orientation hint in the CameraSensorViewConfiguration, indicating how the data in `CameraSensorView::imageData` should be interpreted.

#### Describe the solution you would like
Add an integer field "orientation" (required, default 1, range 1..8) to CameraSensorViewConfiguration to indicate how the image data should be interpreted (or how a sensor/sensor simulator should fill the image_data field). The integer values should correspond to the orientation values of the exif data (i.e. 1 = "image is in correct orientation", 4 = "image has been flipped back-to-front and is upside down", etc.)

#### Describe alternatives you have considered
- use separate boolean field for "mirrored" and a rotation enum (ROTATE_0, ROTATE_90, ROTATE_180, ROTATE_270)
- Using negative values in field_of_view_vertical and _horizontal to indicate flipped data (=>rotation not possible)
- Adding new channel formats to indicate flipped / rotated data

#### Describe the backwards compatibility
By choosing the default "orientation = 1", backwards compatibility is maintained. 

#### Additional context
EXIF specification https://web.archive.org/web/20180921145139if_/http://www.cipa.jp:80/std/documents/e/DC-010-2017_E.pdf

Examples for flipped image data:
- OpenGL image coordinates (0,0 is bottom left) vs "standard" image coordinates (0, 0 is top left)
- Cameras used as mirror replacements
- model-specific technical details of the camera/sensor chip

RIFJo:
@kmeids i have successfully added a pull-request that passes DCO signoff. The first one had the wrong author name in it.

pmai:
CCB 2022-05-02: The use case for this in regards to OSI SensorView (rather than native camera output formats, which are not OSI-based) is not very clear yet, and should be elaborated.

[...]

PR #698, opened 2022-12-15
(feat SensorModelling) add pixel_order field

RIFJo:
Signed-off-by: RIFJo <jochmann@rt.rif-ev.de>

refers to #620

#### Add a description
Adds a "pixel_order" field and enum to allow mirrored versions of the pixel layouts.  This is the result of the discussions with the participants in the sensor modelling workgroup meetings.

This pull request also includes the changes from Thomas Sedlmayer in https://github.com/OpenSimulationInterface/open-simulation-interface/pull/684

### DIFF
--- a/osi_sensorview.proto
+++ b/osi_sensorview.proto
@@ -313,8 +313,10 @@ message CameraSensorView
 
     // Raw image data.
     //
-    // The raw image data in the memory layout and order specified by the
-    // camera sensor input configuration.
+    // The raw image data in the memory layout specified by the camera sensor
+    // input configuration. Default pixel order is left-to-right, top-to-bottom
+    ///(think of scan lines in a TV), but modifications can be 
+    // requested/indicated in CameraSensorViewConfiguration.pixel_order.
     //
     optional bytes image_data = 2;
 }
--- a/osi_sensorview.proto
+++ b/osi_sensorview.proto
@@ -315,7 +315,7 @@ message CameraSensorView
     //
     // The raw image data in the memory layout specified by the camera sensor
     // input configuration. Default pixel order is left-to-right, top-to-bottom
-    ///(think of scan lines in a TV), but modifications can be 
+    // (think of scan lines in a TV), but modifications can be 
     // requested/indicated in CameraSensorViewConfiguration.pixel_order.
     //
     optional bytes image_data = 2;
--- a/osi_sensorview.proto
+++ b/osi_sensorview.proto
@@ -314,9 +314,7 @@ message CameraSensorView
     // Raw image data.
     //
     // The raw image data in the memory layout specified by the camera sensor
-    // input configuration. Default pixel order is left-to-right, top-to-bottom
-    // (think of scan lines in a TV), but modifications can be 
-    // requested/indicated in CameraSensorViewConfiguration.pixel_order.
+    // input configuration. The pixel order is specified in CameraSensorViewConfiguration.pixel_order with the default value PIXEL_ORDER_LEFT_RIGHT_TOP_BOTTOM.
     //
     optional bytes image_data = 2;
 }
--- a/osi_sensorview.proto
+++ b/osi_sensorview.proto
@@ -313,8 +313,10 @@ message CameraSensorView
 
     // Raw image data.
     //
-    // The raw image data in the memory layout specified by the camera sensor
-    // input configuration. The pixel order is specified in CameraSensorViewConfiguration.pixel_order with the default value PIXEL_ORDER_LEFT_RIGHT_TOP_BOTTOM.
+    // The raw image data in the memory layout specified by the camera    
+    // sensor input configuration. The pixel order is specified in 
+    // CameraSensorViewConfiguration.pixel_order with the 
+    // default value PIXEL_ORDER_LEFT_RIGHT_TOP_BOTTOM.
     //
     optional bytes image_data = 2;
 }
--- a/osi_sensorview.proto
+++ b/osi_sensorview.proto
@@ -316,7 +316,7 @@ message CameraSensorView
     // The raw image data in the memory layout specified by the camera    
     // sensor input configuration. The pixel order is specified in 
     // CameraSensorViewConfiguration.pixel_order with the 
-    // default value PIXEL_ORDER_LEFT_RIGHT_TOP_BOTTOM.
+    // default value PIXEL_ORDER_DEFAULT (i.e. left-to-right, top-to-bottom).
     //
     optional bytes image_data = 2;
 }
--- a/osi_sensorviewconfiguration.proto
+++ b/osi_sensorviewconfiguration.proto
@@ -733,7 +733,7 @@ message CameraSensorViewConfiguration
     //
     // By default, the pixel data is is left-to-right, top-to-bottom. 
     // 
-    // \note for rotations of the pixel data, use the camera coordinate system
+    // \note for rotations of the pixel data, use the camera coordinate system.
     //
     optional PixelOrder pixel_order = 12;
 
@@ -743,8 +743,9 @@ message CameraSensorViewConfiguration
     // in CameraSensorView. 
     //
     // \note this enum does not contain an entry to do mirroring upside down
-    // and left-to-right at the same time, because this is equivalent to a 180°
-    // rotation, which should be indicated in the sensor coordinate system 
+    // and left-to-right at the same time, because this is equivalent to a
+    // 180-degree rotation, which should be indicated in the sensor coordinate
+    // system.
     //
     enum PixelOrder
     {
@@ -752,7 +753,7 @@ message CameraSensorViewConfiguration
         //
         PIXEL_ORDER_UNKNOWN = 0;
 
-        // Unspecified but unknown pixel order.
+        // Known pixel order that is not supported by OSI.
         // Consider proposing an additional format if using
         // \c #PIXEL_ORDER_OTHER.
         //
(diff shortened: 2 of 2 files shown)


osi_sensorviewconfiguration.proto
```diff
@@ -729,6 +729,52 @@ message CameraSensorViewConfiguration
...
+    //
+    optional PixelOrder pixel_order = 12;
+
+    // Pixel layout 
+    //
+    // Pixel layout documents the order of pixels in the \c image_data
+    // in CameraSensorView. 
+    //
+    // \note this enum does not contain an entry to do mirroring upside down
+    // and left-to-right at the same time, because this is equivalent to a 180°
```
thomassedlmayer, line 746: The build tests show that the degree sign leads to some problems by not being an ASCII character.

osi_sensorviewconfiguration.proto
```diff
@@ -729,6 +729,52 @@ message CameraSensorViewConfiguration
...
+    // rotation, which should be indicated in the sensor coordinate system 
+    //
+    enum PixelOrder
+    {
+        // Pixel order is unknown (must not be used).
+        //
+        PIXEL_ORDER_UNKNOWN = 0;
+
+        // Unspecified but unknown pixel order.
```
thomassedlmayer, line 755: Usually if "other" is used, it is referred to a known (not unknown) thing, e.g. a known pixel order that is not specified in OSI.

osi_sensorviewconfiguration.proto
RIFJo: ok, i have fixed this in the follow-up commit [a1fc975](https://github.com/OpenSimulationInterface/open-simulation-interface/pull/698/commits/a1fc9757224f356a13f3dba377a9857c0af76e37)

osi_sensorviewconfiguration.proto
RIFJo: ok, i have fixed this in the follow-up commit [a1fc975](https://github.com/OpenSimulationInterface/open-simulation-interface/pull/698/commits/a1fc9757224f356a13f3dba377a9857c0af76e37)

RIFJo:
plus one more commit to clean up the triple-slash.

osi_sensorviewconfiguration.proto
```diff
@@ -729,6 +729,53 @@ message CameraSensorViewConfiguration
...
+        // Known pixel order that is not supported by OSI.
+        // Consider proposing an additional format if using
+        // \c #PIXEL_ORDER_OTHER.
+        //
+        PIXEL_ORDER_OTHER = 1;
+        
+        // Pixel data is not mirrored (Default).
+        // Pixels are ordered left-to-right, top-to-bottom
+        //
+        PIXEL_ORDER_DEFAULT = 2;
```
LukasElster, line 765: Outcome Sensor Modeling Group meeting 25.01.2023: Place PIXEL_ORDER_DEFAULT on position 0 and remove PIXEL_ORDER_UNKNOWN

jdsika: The unknown is a mandatory field in OSI due to the fact that it is an enum.

jdsika: we would have to change the rules for enum definitions then?

thomassedlmayer: We (Pierre, Lukas and I) were aware of that guideline in OSI but discussed that it would not make sense to ever use unknown in this context as this is just a "technical" enum describing the pixel order that applies to the image data. It is not some kind of classification for which the guideline makes sense. I think there already were some other places in OSI where it was decided that the unknown/other guideline does not have to be enforced, especially when describing technical stuff.

jdsika:
We have to check if we can edit the [interface_conventions](https://opensimulationinterface.github.io/osi-documentation/#top-interface_conventions)

jdsika:
* Rename default to the actual pixel order 
* Reference default enum value correctly in the comments

ClemensLinnhoff:
> * Rename default to the actual pixel order
> * Reference default enum value correctly in the comments

I changed it accordingly. I also changed the (for me) confusing naming of "mirrored" to the actual pixel order.

RIFJo:
> I also changed the (for me) confusing naming of "mirrored" to the actual pixel order.
> @RIFJo does this work for you?
I am unhappy about this change of wording, but i can live with it. Maybe keep the enum entries as it is now, but mention the mirroring direction in the documentation comment ("the pixels are ordered right to left, top to bottom. This represents a mirroring about the vertical axis (left-to-right) compared to the default order") ?

Anyway, thats just minor nitpicking. I'm okay if this goes into 3.6 as it is now.

thomassedlmayer:
Actually, we did not want every possible pixel order to be present in this enum but only the ones that manipulate the image in a way that can not be done by repositioning the sensor with the sensor mounting position (e.g. turn sensor by 180 deg upside down). So we only provide the fields for mirroring and otherwise the default should be used. I think this information should still somehow be provided in the description otherwise people will probably think that some pixel orders are just missing. Another option may be to not call the field "pixel order" but mirroring state or similar and also state that this field influences the image data's pixel order? It should be clear what we do/do not want to achieve with this field.

jdsika:
> Actually, we did not want every possible pixel order to be present in this enum but only the ones that manipulate the image in a way that can not be done by repositioning the sensor with the sensor mounting position (e.g. turn sensor by 180 deg upside down). So we only provide the fields for mirroring and otherwise the default should be used. I think this information should still somehow be provided in the description otherwise people will probably think that some pixel orders are just missing. Another option may be to not call the field "pixel order" but mirroring state or similar and also state that this field influences the image data's pixel order? It should be clear what we do/do not want to achieve with this field.

Feel free to add to the comments. The default is now clearly defined and IMO it is way easier to understand now. Setting an enum to `standard` or to `left to right` is telling the user in a better way what the standard is.

ClemensLinnhoff:
> Actually, we did not want every possible pixel order to be present in this enum but only the ones that manipulate the image in a way that can not be done by repositioning the sensor with the sensor mounting position (e.g. turn sensor by 180 deg upside down). So we only provide the fields for mirroring and otherwise the default should be used. I think this information should still somehow be provided in the description otherwise people will probably think that some pixel orders are just missing. Another option may be to not call the field "pixel order" but mirroring state or similar and also state that this field influences the image data's pixel order? It should be clear what we do/do not want to achieve with this field.

This is still part of the documentation:
> // \note this enum does not contain an entry for right-to-left and bottom-to-top, because this is equivalent to a
    // 180-degree rotation of the default, which should be indicated in the sensor coordinate
    // system.
    //

pmai:
I'm unclear on the sudden changes to this PR; AFAICT the definitions in this PR were the result of various long running discussions in the sensor modelling WG. The now added changes do not seem to improve upon this, but rather seem to diverge, without fixing the enum issue (the default enum value should be 0).

So if this really is supposed to be rediscussed, it needs rediscussion in the sensor modeling group prior to being ready for CCB. 

On the enum issue: The interface guidelines are just internal project guidelines and do not themselves have normative force; for ground-truth-like data this approach still makes sense, however for technical protocol-level configuration data, it does not, hence deviation is warranted, and has already happened, see the `ReferenceLine` `Type` enum, as an example.

Here the 0 value should be used for the old default behavior, since this keeps 3.5 and 3.6 interoperability consistent. This aids backward/forward-compatibility,

osi_sensorviewconfiguration.proto
```diff
@@ -729,6 +729,51 @@ message CameraSensorViewConfiguration
     //
     repeated WavelengthData wavelength_data = 11;
 
+    // Indicates if and how the the pixel data is ordered
+    //
+    // The default value is PIXEL_ORDER_LEFT_RIGHT_TOP_BOTTOM.
```
thomassedlmayer, line 734: @RIFJo I think this field's comment needs some adaptation as well. At least it is referring to a enum value that is no longer there.

RIFJo: good catch, fixed in f09484

osi_sensorviewconfiguration.proto
```diff
@@ -729,6 +729,53 @@ message CameraSensorViewConfiguration
...
+        // Consider proposing an additional format if using
+        // \c #PIXEL_ORDER_OTHER.
+        //
+        PIXEL_ORDER_OTHER = 1;
+                
+        // Pixels are ordered right-to-left, top-to-bottom.
+        // Pixel data is mirrored along the vertical axis.
+        // The image is mirrored left-to-right.
+        //
+        PIXEL_ORDER_RIGHT_LEFT_TOP_BOTTOM = 3;
```
thomassedlmayer, line 769: Number 2 was skipped. (Sorry that I saw that only now after you already made another commit @RIFJo.)
It should probably be:
PIXEL_ORDER_RIGHT_LEFT_TOP_BOTTOM = 2
PIXEL_ORDER_LEFT_RIGHT_BOTTOM_TOP = 3

osi_sensorview.proto
```diff
@@ -313,8 +313,10 @@ message CameraSensorView
 
     // Raw image data.
     //
-    // The raw image data in the memory layout and order specified by the
-    // camera sensor input configuration.
+    // The raw image data in the memory layout specified by the camera    
+    // sensor input configuration. The pixel order is specified in 
+    // CameraSensorViewConfiguration.pixel_order with the 
+    // default value PIXEL_ORDER_LEFT_RIGHT_TOP_BOTTOM.
```
thomassedlmayer, line 319: @RIFJo I just found another wrong enum in this comment. As soon as this is fixed we should probably add the label ReadyForCCB to this PR, right?

RIFJo: I have fixed the codumentation. I think i am unable to add the ReadyForCCB label, someone else needs to do that.

pmai:
CCB on 2023-03-27: Merge as-is.

pmai (approved):

PR #701, opened 2023-01-18
Logical lanes: add street name

tbleher:
The street name is helpful when displaying the map (e.g. for navigation purposes). The street name is added to the logical lanes, since the street name will typically come from the underlying map, and most map formats are more similar to logical lanes than to physical lanes.

### DIFF
--- a/osi_logicallane.proto
+++ b/osi_logicallane.proto
@@ -582,6 +582,10 @@ message LogicalLane
     //
     repeated LaneConnection successor_lane = 15;
 
+    // Name of the street this lane belongs to.
+    //
+    optional string street_name = 16;
+
     //
     // Definition of available lane types.
     //


ThomasNaderBMW (approved):
Passt auch aus meiner Sicht zu den logical lanes.
Es ist keine Eigenschaft die physisch vorliegt oder von Sensoren detektiert werden könnte.

thomassedlmayer:
Harmonization WG 2023/03/01: You would have to create a new lane if the streetname of a road changes. Could maybe also/instead be added to the RouteSegment proposed in #705?

ThomasNaderBMW:
WG Other Models:
Route Segment is just a piece of the map and doesnt describe the whole map.
For us it is ok.

pmai:
CCB 2023-03-13: Merge as-is.

Issue #462, opened 2021-01-11
No RCS can be assigned to detected stationary objects


hschoen:
Whereas an RCS can be assigned to detected moving objects this is not possible for stationary objects (and traffic lights/signs).

For detected moving objects the related messages are:

`optional RadarSpecificObjectData | radar_specifics = 100`

`optional double osi3::RadarSpecificObjectData::rcs = 1`

pmai:
I agree that the sensor specific fields (not only for radar) should also be added to DetectedStationaryObject. This would be an issue for the SensorModeling group, and could be done backward compatibly for a 3.4, for example.

kmeids:
@hschoen, I agree as well!
Is it possible for you to join one of our Sensor Modeling WG meetings?
They take place on a Bi-weakly basis on Friday from 9:00 to 10:00 CET.

hschoen:
Hello,

yes, I can join one of your Sensor Modeling WG meetings. Could you forward me a meeting invitation? I could join this week, but not on 5 Feb.

Mit freundlichen Grüßen / Best regards

Dr. Helmut Schoen

HF-Sensoren, EMV (CR/AEL1)
Robert Bosch GmbH | Postfach 10 60 50 | 70049 Stuttgart | GERMANY | www.bosch.com
Tel. +49 711 811-10880 | Mobil +49 173 9187997 | Telefax +49 711 811 | Threema<threema://add/?id=K286UVMU> / Threema Work<threemawork://add/?id=K286UVMU>: K286UVMU | Helmut.Schoen@de.bosch.com<mailto:Helmut.Schoen@de.bosch.com>

Sitz: Stuttgart, Registergericht: Amtsgericht Stuttgart, HRB 14000;
Aufsichtsratsvorsitzender: Franz Fehrenbach; Geschäftsführung: Dr. Volkmar Denner,
Prof. Dr. Stefan Asenkerschbaumer, Filiz Albrecht, Dr. Michael Bolle, Dr. Christian Fischer,
Dr. Stefan Hartung, Dr. Markus Heyn, Harald Kröger, Rolf Najork, Uwe Raschke
​
Von: Kmeid <notifications@github.com>
Gesendet: Mittwoch, 27. Januar 2021 13:28
An: OpenSimulationInterface/open-simulation-interface <open-simulation-interface@noreply.github.com>
[...]

PR #702, opened 2023-01-25
Add sensor-specific data to detected stationary objects

pmai:
This addresses #462 partially, probably data should also be added to detected traffic signs and lights. Maybe add to DetectedItemHeader instead? Needs to be harmonized with ISO 23150 as well.

### DIFF
--- a/osi_detectedobject.proto
+++ b/osi_detectedobject.proto
@@ -118,6 +118,34 @@ message DetectedStationaryObject
     //
     optional ColorDescription color_description = 5;
 
+    // Additional data that is specific to radar sensors.
+    //
+    // \note Field needs not to be set if simulated sensor is not a radar
+    // sensor.
+    //
+    optional RadarSpecificObjectData radar_specifics = 100;
+
+    // Additional data that is specific to lidar sensors.
+    //
+    // \note Field needs not to be set if simulated sensor is not a lidar
+    // sensor.
+    //
+    optional LidarSpecificObjectData lidar_specifics = 101;
+
+    // Additional data that is specific to camera sensors.
+    //
+    // \note Field needs not to be set if simulated sensor is not a camera
+    // sensor.
+    //
+    optional CameraSpecificObjectData camera_specifics = 102;
+
+    // Additional data that is specific to ultrasonic sensors.
+    //
+    // \note Field needs not to be set if simulated sensor is not an ultrasonic
+    // sensor.
+    //
+    optional UltrasonicSpecificObjectData ultrasonic_specifics = 103;
+
     //
     // \brief A candidate for a detected stationary object as estimated
     // by the sensor.


LukasElster (approved):

jdsika:
Where is the definition of the field? :)

pmai:
> Where is the definition of the field? :)

Already exists in sensorspecifics...

osi_sensorspecific.proto
```diff
@@ -56,7 +56,7 @@ message UltrasonicSpecificObjectData
     //
     optional double maximum_measurement_distance_sensor = 1;
 
-    // This value indicates the propability height for the classification in the
+    // This value indicates the probability height for the classification in the
```
jdsika, line 59: now that I have corrected the typo I am not sure if "probability height" is even a proper english term haha

jdsika:
Considering the fact that the RCS is a measured unit somehow with dependencies on frequency, viewing angle etc... maybe it is suited better in an detectedItemHeader

pmai:
> Considering the fact that the RCS is a measured unit somehow with dependencies on frequency, viewing angle etc... maybe it is suited better in an detectedItemHeader

Or something similar; let's see where ISO 23150 nowadays puts that information, maybe this gives a clearer picture, at least for 4.0.0; for 3.6.0 we likely should keep as similar to the moving objects as possible...

jdsika:
> > Considering the fact that the RCS is a measured unit somehow with dependencies on frequency, viewing angle etc... maybe it is suited better in an detectedItemHeader
> 
> Or something similar; let's see where ISO 23150 nowadays puts that information, maybe this gives a clearer picture, at least for 4.0.0; for 3.6.0 we likely should keep as similar to the moving objects as possible...

Then I would opt for a creating an issue for a potential breaking change in v4.0.0 as the fields of movingObjekts would need to be moved as well and at the same time for consistency merge this proposal for v3.6.0

thomassedlmayer:
As far as I can tell, sensor specific data are only referenced in moving objects in the ISO as of 2022-11-22 (even though they generically refer to a 'recognized object' in the description of RCS/lidar reflectivity).

I'd also say that it generally makes sense to put it in the generic header in the longterm as they are not specific to some object type but rather a commonality among any detected objects. If this is what we want, I can create another PR for the 4.0 milestone.

thomassedlmayer:
Sensor Model WG (08.02.): No opinions on that. Passed on to harmonization WG to check with ISO23150.

PhRosenberger:
> Sensor Model WG (08.02.): No opinions on that. Passed on to harmonization WG to check with ISO23150.

I would say it does not matter what is written in the ISO23150. We use these information for a different purpose and you want to have this information on all objects for your sensor model, anyways. The actual sensor will calculate e.g. an RCS for any detected object, as all are potentially moving at first sight and could be actually static objects by error. So in simulation you need it on all objects to be able to replicate such erroneous detected (potentially moving) objects.

pmai:
CCB 2023-03-13: Merge as-is.

PR #708, opened 2023-02-02
Move sensor specifics to DetectedItemHeader

### LINKED ISSUE #462: No RCS can be assigned to detected stationary objects

thomassedlmayer:
Signed-off-by: Thomas Sedlmayer <tsedlmayer@pmsfit.de>

This addresses #462. The PR introduces sensor-specifics in DetectedItemHeader and deprecates all sensor-specific fields that are located in DetectedMovingObject and DetectedStationaryObject directly.

This also should be discussed in terms of harmonization with ISO23150.

Resolves #462

### DIFF
--- a/osi_detectedobject.proto
+++ b/osi_detectedobject.proto
@@ -67,6 +67,34 @@ message DetectedItemHeader
     //
     repeated Identifier sensor_id = 6;
 
+    // Additional data that is specific to radar sensors.
+    //
+    // \note Field needs not to be set if simulated sensor is not a radar
+    // sensor.
+    //
+    optional RadarSpecificObjectData radar_specifics = 100;
+
+    // Additional data that is specific to lidar sensors.
+    //
+    // \note Field needs not to be set if simulated sensor is not a lidar
+    // sensor.
+    //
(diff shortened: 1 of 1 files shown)


yash-shah-asam:
@thomassedlmayer could you please handle this as the part of 23150 project.

thomassedlmayer:
As discussed in the last sensor modeling project meeting, we agreed that I will modify this PR to only deprecate certain fields instead of removing in order to be backwards compatible (minor release) but also already pave the way for a potential major release.

Before merging anything though, we need to check how this is done in the new ISO23150 release as soon as we get access.

PhRosenberger (approved):
No further comments from my side.

asadekasam:
OSI CCB 2026.08.13: It needs to be checked against the latest ISO 23150.

PR #710, opened 2023-02-10
Draft for discussion: Add detailed pedestrian description to MovingObject

Peintnerer:
#### Add a description
Related to WP12 in other models workgroup. For exchanging motion capture data of pedestrians (or humans in general) between simulators, we need a more detailed description of pedestrians. This is based on similar skeleton structures used for character rigging in CARLA, Unity, Unreal, etc.


osi_object.proto
```diff
@@ -512,6 +512,65 @@ message MovingObject
...
+    //
+    message SkeletonPoint {
+
+
+        // Skeleton points are identified by a naming scheme
+        //  
+        // To properly identify the skeleton points the pre-defined naming convention 
+        // must be used.
+        // 
+        //\image html OSI_SkeletonNamingConvention.jpg
```
tbleher, line 554: The file name currently has a spelling error ("Convenion" instead of "Convention").
In the graphics, there is one mis-spelling: "RIng" instead of "Ring".

osi_object.proto
```diff
@@ -512,6 +512,65 @@ message MovingObject
...
+        // 
+        // Reference System is the middle of the bounding box (\c MovingObject::base 
+        // \c BaseMoving::position).
+        // 
+        required Vector3d position = 2;
+
+        // Orientation of the skeleton point
+        //
+        // Reference System is the middle of the bounding box (\c MovingObject::base 
+        // \c BaseMoving::position).
```
tbleher, line 568: This confuses me. It mentions BaseMoving::position, which shouldn't have any impact on the orientation. I assume that it is relative to BaseMoving::orientation? Or is the orientation absolute?

tbleher (changes requested):
Thanks! Looks good to me in general, and I can imagine using it. Only two minor comments from my side.

osi_object.proto
Peintnerer: Yes, this is a copy-paste error in the comment. The correct reference would be BaseMoving::orientation. Will be fixed.

osi_object.proto
Peintnerer: Thank you for checking; I will fix the spelling errors!

osi_object.proto
```diff
@@ -512,6 +512,65 @@ message MovingObject
...
+
+        // Body height of the pedestrian  
+        // 
+        // Unit: m
+        //
+        optional double height = 1;
+
+        // List of all skeleton points of the pedestrian
+        // 
+        // The number of skeleton points may very, based on the detail level of
```
thomassedlmayer, line 531: *vary

osi_object.proto
```diff
@@ -512,6 +512,65 @@ message MovingObject
...
+
+
+        // Skeleton points are identified by a naming scheme
+        //  
+        // To properly identify the skeleton points the pre-defined naming convention 
+        // must be used.
+        // 
+        //\image html OSI_SkeletonNamingConvention.jpg
+        // 
+        required string identifier = 1;
```
thomassedlmayer, line 556: According to the [documentation](https://opensimulationinterface.github.io/osi-documentation/#_overview_of_osi_architecture) OSI does not use the 'required' keyword for fields but all fields should be marked as 'optional'.

jdsika:
I might have some feedback here. I will get back here!

ClemensLinnhoff:
[This](https://github.com/CMU-Perceptual-Computing-Lab/openpose) might be relevant. Did you consider their definitions on skeletal joint positions?

adrianschultz:
https://www.asam.net/static_downloads/ASAM_OpenSCENARIO_V1.2.0_Model_Documentation/modelDocumentation/content/AnimationAction.html

and

https://www.asam.net/index.php?eID=dumpFile&t=f&f=4908&token=ae9d9b44ab9257e817072a653b5d5e98ee0babf8#_appearance

might be helping

osi_object.proto
```diff
@@ -512,6 +512,65 @@ message MovingObject
...
+    // This is an extension to the \c MovingObject with additional information 
+    // describing a pedestrian in more detail.
+    // 
+    message PedestrianData {
+
+        // Body height of the pedestrian  
+        // 
+        // Unit: m
+        //
+        optional double height = 1;
```
ThomasNaderBMW, line 527: We have that information already from Bounding Box information.
So it can be left out and then Skeleton information itself is enough and could be included directly to sth. like pedestrian_attributes

ThomasNaderBMW:
15.05.WG ASAM OSI | Other Models | Focus: Enhanced Pedestrian Modeling:
- Enum instead of string for skeleton joints
- Level of detail: Approach is that just the joints are used that are sent (can be identified by the enum). Rule from root till depeest skeleton point (regarding the tree) all skeleton joints must be sent, so complete chain till that point.
- Will it be part of osi_detectedobject.proto? Depending on that we have to think about if it is vehicle_attributes or vehicle_classification (used in detectedobject)
- Skeleton Data is enough and can be used in pedestrian_attributes/classification. Pedestrian Data not necessary.

osi_object.proto
```diff
@@ -512,6 +512,205 @@ message MovingObject
...
+    message PedestrianAttributes {
+
+
+        // List of all skeleton points of the pedestrian
+        // 
+        // The number of skeleton points may vary, based on the detail level of 
+        // the pedestrian model used. For example, some simulators will not include 
+        // detailed data of the fingers or eyes of the pedestrian.
+        // 
+        repeated SkeletonPoint skeleton_data = 2;
```
thomassedlmayer, line 530: skeleton_data has the number 2. This can be changed to 1 in the next commit.

tbleher:
Two comments regarding the current patch-set:
* It seems like the new `PedestrianAttributes` message is not yet referenced anywhere, so currently no pedestrian can be described using this new message. I think it should be added at least to `MovingObject`, so that the values can be returned from `TrafficUpdate` (which contains `MovingObject`), and so that the attributes are part of the `GroundTruth`.
* It would be great to have also have some more coarse-grained values, like the current viewing direction.

I want to elaborate a bit on the second point:
The current proposal seems suited for transfering the actual movement of a pedestrian between different systems. However, not all models work with this level of detail. E.g. the open source OSIPedestrian model, which was created as part of the SETLevel project, only works on the bounding box level. 3D animation is done afterwards, e.g. using Unreal. This works fine in practice. However, currently the OSIPedestrian model cannot provide semantic information, e.g. where the pedestrian is currently looking at. This information has to come from OSIPedestrian, since it depends on the situation (example: if the pedestrian is waiting to cross a busy street, it has to look left and right. This semantic info should be provided by the model, and not later be added e.g. in Unreal). However, OSIPedestrian does not do complete animation, since that means it would have to add the whole logic for pedestrian animation inside the model (which is much better done afterwards in Unreal or Unity).
The current patch doesn't really allow a model like OSIPedestrian to add a simple info like e.g. the head direction, without knowing the rest of the geometry of the pedestrian. I propose to add this semantic information separately (and am willing to submit a patch for this).

Comments/Objections?

ThomasNaderBMW:
WG Other Models:
- Fix Build Errors
- Add Hirarchy-Picture
- Can be reviewed by CCB.

osi_object.proto
```diff
@@ -512,6 +512,191 @@ message MovingObject
...
+            optional float length = 4;
+
+            // Determines if bones of a pedestrian are explicitly missing from the model.
+            //
+            // In case a pedestrian has missing limbs this can be explicitly communicated
+            // by setting this boolean to TRUE.
+            //
+            optional bool missing = 5;
+
+            // The type of the bone
```
ThomasNaderBMW, line 606: bone.

ThomasNaderBMW (approved):
Some minor stuff.

ThomasNaderBMW:
CCB Review, 19.06.23:
Can be merged, but minor stuff (Comments, DCO, correct Builds) has to be fixed till Thursday 9 AM

osi_object.proto
```diff
@@ -512,6 +512,191 @@ message MovingObject
...
+            //
+            // A bone's name, position, and orientation is defined by the end-point closer
+            // to the root. For example: the "LOWER_ARM_L" will define the point in the left
+            // elbow of the model.
+            //
+            // If a bone which is more than one layer detached from the
+            // root point is used, all bones between that bone and the root also
+            // need to be defined in order to create a complete chain!
+            //
+            // If information about bones are missing, they may be left empty. A complete
```
LudwigFriedmann, line 572: Short-notice review: I'd remove "A complete set of bones for the entire skeleton cant't be expected!". This is potentially misleading.

osi_object.proto
Peintnerer: Has been fixed.

osi_object.proto
```diff
@@ -512,6 +512,191 @@ message MovingObject
...
+            // set of bones for the entire skeleton can't be expected!
+            // 
+            optional Type type = 1;
+
+            // Position of the bone
+            // 
+            // Reference System is the root, defined by bbcenter_to_root 
+            // (\c PedestrianAttributes::bbcenter_to_root). 
+            //
```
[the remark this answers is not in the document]
Peintnerer: Line is removed.

osi_object.proto
```diff
@@ -512,6 +512,191 @@ message MovingObject
...
+                TYPE_FULL_HAND_L = 12;
+
+                // Bone defines one of the hands.
+                //
+                TYPE_FULL_HAND_R = 13;
+
+                // Bone defines one of the thighs.
+                //
+                TYPE_UPPER_LEG_L = 14;
+                // Bone defines one of the thighs.
```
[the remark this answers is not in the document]
Peintnerer: Added in an empty line.

osi_object.proto
Peintnerer: Fixed.

osi_object.proto
```diff
@@ -1062,18 +1073,191 @@ message MovingObject
...
+            //
+            // Reference System is the root, defined by bbcenter_to_root
+            // (\c PedestrianAttributes::bbcenter_to_root).
+            //
+            optional Orientation3d orientation = 3;
+
+            // Orientation of the bone.
+            //
+            // Reference System is the root, defined by bbcenter_to_root
+            // (\c PedestrianAttributes::bbcenter_to_root).
```
pmai, line 1158: Duplicated field description, maybe should be:
```suggestion
            // Length of the bone.
            //
            // Measured along its major axis.
            //
            // Unit: m
```

osi_object.proto
```diff
@@ -1062,18 +1073,191 @@ message MovingObject
...
+            //
+            optional float length = 4;
+
+            // Determines a bone is explicitly missing from the pedestrian model.
```
pmai, line 1162:
```suggestion
            // Determines whether a bone is explicitly missing from the pedestrian model.
```

osi_object.proto
```diff
@@ -1062,18 +1073,191 @@ message MovingObject
...
+            //
+            optional bool missing = 5;
+
+            // The type of the bone.
+            //
+            // \note Bone of each type can be assigned, or left empty depending
+            // on the desired level of detail, or present data. However, if a
+            // bone is defined, all bones in the chain from that bone back to
+            // the root point must be defined to create a complete chain.
+            //
```
pmai, line 1175: I suggest moving this comment to the field definition, where level of detail is already mentioned. This is more a use thing than an enum type thing.
```suggestion
```

osi_object.proto
```diff
@@ -1062,18 +1073,191 @@ message MovingObject
...
+        // of the root point of the pedestrian skeleton model.
+        //
+        optional Vector3d bbcenter_to_root = 1;
+
+        // List of all bones of the pedestrian.
+        //
+        // The number of bones may vary, based on the detail level of
+        // the pedestrian model used. For example, some simulators will not include
+        // detailed data about the hands of a pedestrian.
+        //
```
pmai, line 1109:
```suggestion
        //
        // \note A bone of each type can be provided, or left out, depending
        // on the desired level of detail, or available data. However, if a
        // bone is defined, all bones in the chain from that bone back to
        // the root point must be provided to create a complete chain.
        //
```

osi_object.proto
```diff
@@ -1062,18 +1073,191 @@ message MovingObject
...
+                // Root point of the pedestrian. Defined by bbcenter_to_root
+                // (\c PedestrianAttributes::bbcenter_to_root).
+                //
+                TYPE_ROOT = 0;
+
+                // Bone defining the hip.
+                //
+                TYPE_HIP = 1;
+
+                // Bone defines the lower part of the spine.
```
pmai, line 1189: I suggest to always use defining, rather than defines, in this list...
```suggestion
                // Bone defining the lower part of the spine.
```

pmai (changes requested):
Minor improvements that might be needed.

osi_object.proto
```diff
@@ -1062,18 +1073,191 @@ message MovingObject
...
+
+                // Bone defines the left forearm.
+                //
+                TYPE_LOWER_ARM_L = 10;
+
+                // Bone defines the right forearm.
+                //
+                TYPE_LOWER_ARM_R = 11;
+
+                // Bone defines the left hand.
```
pmai, line 1229:
```suggestion
                // Bone defining the upper part of the spine.
                //
                TYPE_UPPER_SPINE = 3;

                // Bone defining the neck.
                //
                TYPE_NECK = 4;

                // Bone defining the head.
                //
                TYPE_HEAD = 5;

                // Bone defining the left shoulder.
                //
                TYPE_SHOULDER_L = 6;

                // Bone defining the right shoulder.
                //
                TYPE_SHOULDER_R = 7;

                // Bone defining the left upper arm.
                //
                TYPE_UPPER_ARM_L = 8;

                // Bone defining the right upper arm.
                //
                TYPE_UPPER_ARM_R = 9;

                // Bone defining the left forearm.
                //
                TYPE_LOWER_ARM_L = 10;

                // Bone defining the right forearm.
                //
                TYPE_LOWER_ARM_R = 11;

                // Bone defining the left hand.
```

osi_object.proto
```diff
@@ -1062,18 +1073,191 @@ message MovingObject
...
+
+                // Bone defines the right shin.
+                //
+                TYPE_LOWER_LEG_R = 17;
+
+                // Bone defines the left foot.
+                //
+                TYPE_FULL_FOOT_L = 18;
+
+                // Bone defines the right foot.
```
pmai, line 1257:
```suggestion
                // Bone defining the right hand.
                //
                TYPE_FULL_HAND_R = 13;

                // Bone defining the left thigh.
                //
                TYPE_UPPER_LEG_L = 14;

                // Bone defining the right thigh.
                //
                TYPE_UPPER_LEG_R = 15;

                // Bone defining the left shin.
                //
                TYPE_LOWER_LEG_L = 16;

                // Bone defining the right shin.
                //
                TYPE_LOWER_LEG_R = 17;

                // Bone defining the left foot.
                //
                TYPE_FULL_FOOT_L = 18;

                // Bone defining the right foot.
```

pmai (approved):

PR #551, opened 2021-07-20
#550 cmake build is missing key features to be properly usable via fetchcontent

globberwops:
﻿#### Reference to a related issue in the repository
Resolves #550 

#### Add a description
This PR aims to modernize the installation logic of the project by:
* removing redundant warnings
* prefix cache variables with the project name
* provide namespaced alias targets usable from the build tree
* provide a namespace for the export set

This modernization will make the project usable via [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) by a downstream project.

#### Take this checklist as orientation for yourself, if this PR is ready for the Change Control Board:
- [x] My suggestion follows the [style and contributors guidelines](https://opensimulationinterface.github.io/osi-documentation/open-simulation-interface/doc/howtocontribute.html).
- [x] I have taken care about the [documentation](https://opensimulationinterface.github.io/osi-documentation/open-simulation-interface/doc/commenting.html).
- [x] I have done the [DCO signoff](https://opensimulationinterface.github.io/osi-documentation/open-simulation-interface/doc/howtocontribute.html#developer-certification-of-origin-dco).
- [x] My changes generate no errors when passing CI tests. 
- [x] I have successfully implemented and tested my fix/feature locally.
- [x] Appropriate reviewer(s) are assigned.

PR #730, opened 2023-06-15
Fix flatbuffer builds for upcoming release

pmai:
This  pull request collects all fixes to the flatbuffer build to make it fully functional:

- Fix CMake build breakage due to better separation of source and binary directories in #551
- Replace flatbuffers submodule with find_package dependency
- Temporary work-around for flatc bug with same-named nested enum types
- Add project-scoped alias for flatbuffers-based library

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "flatbuffers"]
-	path = flatbuffers
-	url = https://github.com/google/flatbuffers.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,9 +100,10 @@ protobuf_generate_cpp(PROTO_SRCS PROTO_HEADERS ${OSI_PROTO_FILES})
 set(FLAT_HEADERS "")
 if(OSI_BUILD_FLATBUFFER)
   set(FLAT_FBS "")
-  add_subdirectory("flatbuffers"
-                   ${CMAKE_CURRENT_BINARY_DIR}/flatbuffers-build
-                   EXCLUDE_FROM_ALL)
+  find_package(flatbuffers REQUIRED)
+  if(NOT FLATBUFFERS_FLATC_EXECUTABLE)
+    set(FLATBUFFERS_FLATC_EXECUTABLE ${flatbuffers_DIR}/../../tools/flatbuffers/flatc)
+  endif()
   file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/descriptor.fbs" "namespace osi3;")
   file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/include")
   list(APPEND FLAT_FBS "${CMAKE_CURRENT_BINARY_DIR}/descriptor.fbs")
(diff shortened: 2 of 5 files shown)


.gitmodules
```diff
@@ -1,3 +0,0 @@
-[submodule "flatbuffers"]
```
PhRosenberger, line 1: I am a bit surprised that the submodule is deleted here and `find_package` is used in CMakeLists instead. @pmai why do you suggest this particular change?

Background info: I find it very convenient to use a submodule here and not have the dependency on an (possibly) existing package. It is one of the pro-points on the list for flatbuffers that this is somehow possible..

CMakeLists.txt
```diff
@@ -100,9 +100,10 @@ protobuf_generate_cpp(PROTO_SRCS PROTO_HEADERS ${OSI_PROTO_FILES})
 set(FLAT_HEADERS "")
 if(OSI_BUILD_FLATBUFFER)
   set(FLAT_FBS "")
-  add_subdirectory("flatbuffers"
-                   ${CMAKE_CURRENT_BINARY_DIR}/flatbuffers-build
-                   EXCLUDE_FROM_ALL)
+  find_package(flatbuffers REQUIRED)
```
PhRosenberger, line 103: See my comment above.

CMakeLists.txt
```diff
@@ -111,9 +112,9 @@ if(OSI_BUILD_FLATBUFFER)
     set(fbs "${proto_base}.fbs")
     add_custom_command(
       OUTPUT "${fbs}"
-      COMMAND $<TARGET_FILE:flatc> -I "${PROTOBUF_IMPORT_DIRS}" -o "${CMAKE_CURRENT_BINARY_DIR}" --proto "${CMAKE_CURRENT_SOURCE_DIR}/${proto}"
-      DEPENDS "${proto}" flatc
-      WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}"
+      COMMAND ${FLATBUFFERS_FLATC_EXECUTABLE} -I "${PROTOBUF_IMPORT_DIRS}" -I "${CMAKE_CURRENT_BINARY_DIR}" -o "${CMAKE_CURRENT_BINARY_DIR}" --proto "${proto}"
```
PhRosenberger, line 115: Is this change fixing an issue?

osi_lane.proto
PhRosenberger, line 1: @pmai could you already test the changes and are they definetly fixing the problems when the flatbuffers compiler takes action?

PhRosenberger (commented):
I am not sure about the removing of the submodule and the usage of `find_package` instead. Let's discuss this in the CCB meeting?

.gitmodules
pmai: While I also found this somewhat convenient, it has its own problems (both in nailing down the version of flatbuffers needlessly, and not making it easy to configure for different target settings), and it is not what we do for protobuf. And with todays availability of vcpkg and others for installing flatbuffers, it seems better to stay consistently out of the submodule business for these kinds of dependencies.

CMakeLists.txt
pmai: Yes: 
- Use of -I due to the cleaner separation of SOURCE and BINARY directories (we formerly generated files in the SOURCE directory, which we shouldn't)
- Since flatbuffers are not longer co-built, the flatc executable is not a target anymore, but rather referenced via variable.

osi_lane.proto
pmai: Yes, everything builds correctly and the correct enums are referenced. These were the  only instances where multiple same-named types were present in the same file, so this should fix it across the board. However further review testing should be conducted.

.gitmodules
PhRosenberger: Understood. On the other side, nailing down the version of flatbuffers (and possibly protobuf) could prevent some missunderstandings in combining different models / tool plugins. The flexibility to configure for different is indeed very important for our project. So if using submodules can be a problem to this point, I am ok with the change here.

ThomasNaderBMW:
CCB Review, 19.06.23:

Can be merged.

PhRosenberger (approved):

PR #695, opened 2022-12-06
Add BoundingBox section

ndunningbmw:
﻿#### Reference to a related issue in the repository
Add a reference to a related issue in the repository.

#### Add a description
Add a description of the changes proposed in the pull request.

**Some questions to ask**:
What is this change?
What does it fix?
Is this a bug fix or a feature? Does it break any existing functionality or force me to update to a new version?
How has it been tested?

#### Take this checklist as orientation for yourself, if this PR is ready for the Change Control Board:
- [ ] My suggestion follows the [style and contributors guidelines](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/specification/contributing/start_contributing.html).
- [ ] I have taken care about the [message documentation](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/specification/contributing/commenting_messages.html) and the [fields and enums documentation](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/specification/contributing/commenting_fields_enums.html).
- [ ] I have done the [DCO signoff](https://opensimulationinterface.github.io/osi-documentation/open-simulation-interface/doc/howtocontribute.html#developer-certification-of-origin-dco).
- [ ] My changes generate no errors when passing CI tests. 
- [ ] I have successfully implemented and tested my fix/feature locally.
- [ ] Appropriate reviewer(s) are assigned.

If you can’t check all of them, please explain why.
If all boxes are checked or commented and you have achieved at least one positive review, you can assign the label ReadyForCCBReview!

Signed-off-by: Nicholas Dunning <Nicholas.Dunning@bmw.de>

PR #732, opened 2023-06-16
Add new message for sub-bounding boxes

pmai:
This PR is a continuation of #695 (and #685) with fixed history. It proposes the addition of bounding box sub-sections to BaseMoving and BaseStationary objects, to help sub-divide and classify different parts of the overall object structure.

### DIFF
--- a/osi_common.proto
+++ b/osi_common.proto
@@ -431,13 +431,6 @@ message BoundingBox
     //
     optional Type contained_object_type = 4;
 
-    // Another type of object contained in the bounding box
-    // 
-    // This field shall only be used if \c contained_object_type is TYPE_OTHER
-    // and the object type is not defined withing \c contained_object_type.
-    //
-    optional string other_object_type = 5;
-
     // Opaque reference of an associated 3D model of the bounding box.
     //
     // \note It is implementation-specific how model_references are resolved to
@@ -448,9 +441,6 @@ message BoundingBox
 
     // Definition of different types of object contained within the bounding box
     //
-    // \note This enum field is mainly a placeholder to be extended in the future.
-    // feel free to suggest type definitions for future releases.
-    //
     enum Type
     {
         // Object of unknown type (must not be used in ground truth).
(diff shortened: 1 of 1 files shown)


thempen:
It was unlcear which pivot point the Position definition represents in relation to the dimension and rotation. Added a comment, that it is the center of the BoundingBox.

For Type "TYPE_OTHER", there was no possibility to express the object more detailed. I introduced a string type to describe it.

This other_object_type offers the possibility to reduce the Type enum and split it up an an additional PR to update it later on. This type should not be in conflict to other current approaches harmonizing object type definition. Through the other_object_type filed, the overall SubBoundingBox approach is already usable.

pmai:
CCB 2024-01-29: Type enum should be reviewed (e.g. by @PhRosenberger, @thomassedlmayer, @pmai) esp. w.r.t. sensor modeling needs and a minimum list should be proposed. If no such list emerges, then having no enums, but also dropping the string fallback (as e.g. model_reference can already be used as a last effort fallback) would be the preferred way forward. Will be re-reviewed at next CCB.

osi_common.proto
```diff
@@ -387,6 +387,97 @@ message LogicalLaneAssignment
...
+        // of the parent object.
+        //
+        TYPE_SIDE_MIRROR = 4;
+
+        // Additional, temporarily attached cargo to an object.
+        //
+        TYPE_CARGO = 5;
+
+        // An overhanging, integral part of on object, which is not 
+        // temporarily attached. e.g. a tree crown, or a light pole arm.
```
thomassedlmayer, line 475: Typo: "...of an object..."
Punctuation: Maybe put a comma instead of full stop before "e.g.".

thempen: Done.

thomassedlmayer: I think you forgot to change the "on" to "an".

pmai:
CCB 2024-02-26: Small addition of examples for TYPE_DOOR in the context of vehicles to be added by @pmai, otherwise is ready for merge as-is.

osi_common.proto
```diff
@@ -387,6 +387,99 @@ message LogicalLaneAssignment
...
+        // parking house gate. The protruding structure is meant to be an 
+        // additional part to a base structure.
+        //
+        TYPE_PROTRUDING_STRUCTURE = 3;
+
+        // Additional, temporarily attached cargo to an object.
+        //
+        TYPE_CARGO = 4;
+
+        // The door of an object.
```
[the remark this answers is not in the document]
pmai:
```suggestion
        // The door of an object.
        //
        // For vehicles, this includes driver and passenger doors, trunk and front hoods, and fuel or charging port covers.
```

osi_common.proto
```diff
@@ -447,6 +546,23 @@ message BaseStationary
...
     //
     repeated Vector2d base_polygon = 4;
+
+    // Sub-divisions of the overall bounding box of the \c BaseStationary object.
+    //
+    // The bounding box sections can include separate parts on partially-opaque
+    // objects such as trees with a distinction between trunk and crown.
+    // 
+    // \note The bounding box sub-divisions can extend beyond the overall
+    // bounding box, however no actual geometry must reside outside of the
```
jdsika, line 556: Except for side mirrors? I somehow don't like this. I am thinking about Praktiker "alles außer Tiernahrung"

jdsika (commented):
Reviewed with critical comment v3.7.0

osi_common.proto
thempen: Yeah! The problem is, that the side mirror is already excluded, so it can't be included in the sub bounding boxes.

jdsika: I think we have to create a TODO here for a major version change to discuss this again. If Sub-Bounding-Boxes would have existed in the beginning this exclusion wouldn't have happend.

ClemensLinnhoff: I would propose not to add TODOs in the documentation but rather create an issue with the breaking change label.

PR #736, opened 2023-07-27
Update release.yml

philipwindecker:
Added Antora Generator call at the end to trigger the GitHub Pages build.

#### Add a description
When a release is published, the release pipeline generates the new release package.
However, currently the Antora Generator only runs in that pipeline to create the slimmer package output. 
The GitHub pages are, therefore, not updated with the new release automatically.

To solve this, a final call was added to trigger the Generator after the package has been created and uploaded. This will ensure that the new tag is also included in the GitHub pages, but only if the release update was successful.


pmai (approved):

jdsika:
Reviewed for v3.7.0

PR #744, opened 2023-11-22
Add osi_trafficcommandupdate.proto to CMakeLists

ClemensLinnhoff:
﻿#### Reference to a related issue in the repository
#743 

#### Add a description
Add osi_trafficcommandupdate.proto to CMakeLists. Fixing the bug, that no header for osi_trafficcommandupdate is built.

#### Take this checklist as orientation for yourself, if this PR is ready for the Change Control Board:
- [x] My suggestion follows the [style and contributors guidelines](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/specification/contributing/start_contributing.html).
- [x] I have taken care about the [message documentation](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/specification/contributing/commenting_messages.html) and the [fields and enums documentation](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/specification/contributing/commenting_fields_enums.html).
- [x] I have done the [DCO signoff](https://opensimulationinterface.github.io/osi-documentation/open-simulation-interface/doc/howtocontribute.html#developer-certification-of-origin-dco).
- [x] My changes generate no errors when passing CI tests. 
- [x] I have successfully implemented and tested my fix/feature locally.
- [ ] Appropriate reviewer(s) are assigned.

PR #746, opened 2023-12-04
Add check that all proto files are included in build

pmai:
As decided in the CCB 2023-12-04 based on the issue fixed in #744, this adds checks to the build infrastructure to check that all proto files are actually referenced in CMakeLists.txt and setup.py.

### DIFF
--- a/.github/workflows/protobuf.yml
+++ b/.github/workflows/protobuf.yml
@@ -32,6 +32,11 @@ jobs:
       with:
         submodules: true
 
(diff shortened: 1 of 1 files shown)


ClemensLinnhoff (approved):
I tested the action commands locally against the old master branch and it reported the missing osi_trafficcommandupdate in the CMakeLists. So it seems to work.

jdsika:
Reviewed v3.7.0

PR #753, opened 2023-12-22
Logical Traffic Rules and Speed Limit

lemmer-fzi:
Introduce a message for traffic rule information for logical lanes. The message is intended to be used by traffic participants to easily get information in traffic rules without having to process traffic signs or other additional information.

While this contains an implementation for a speed limit traffic rule the message is designes so that it can be extended with additional traffic rule types in the future.

### DIFF
--- a/osi_logicallane.proto
+++ b/osi_logicallane.proto
@@ -3,6 +3,7 @@ syntax = "proto2";
 option optimize_for = SPEED;
 
 import "osi_common.proto";
+import "osi_object.proto";
 
 package osi3;
 
@@ -587,6 +588,10 @@ message LogicalLane
     //
     optional string street_name = 16;
 
+    // A list of traffic rules on the lane.
+    //
+    repeated TrafficRule traffic_rule = 17;
+
     //
     // Definition of available lane types.
     //
@@ -822,5 +827,117 @@ message LogicalLane
         //
         optional double end_s_other = 5;
     }
+
+    //
+    // Describes traffic rules on a lane. 
+    // The traffic rule can thereby induced by regulations, traffic signs 
+    // or by other means. If the modeled traffic rule is induced by a traffic sign
+    // the information should be identical with the respective traffic sign.
+    //
+    // Note: Every instance should be corresponding to only one specific rule.
+    // The type of the traffic rule should be set using the respective filed.
+    // Additionally, every message should contain the traffic rule validity information
+    // and the respective field for the respective traffic rule type.
+    //
+    // Note: Each traffic rule corresponds to only one lane. If the traffic rule
+    // is also valid on adjacent/successing/predecessing lanes it needs to be
+    // specified for each lane individually.
+    //
+    // \brief Logical Model of a traffic rule on a lane.
+    //
+    message TrafficRule {
+
+        // The type of the traffic rule.
+        //
+        // This specifies the type of the traffic rule to be modeled.
+        // Based on the type the respective message containing the information
+        // corresponding to the traffic rule should be filled.
+        //
+        optional TrafficRuleType traffic_rule_type = 1;
+
+        // The validity information of the traffic rule.
+        //
+        optional TrafficRuleValidity traffic_rule_validity = 2;
+
+        // Traffic rule information for traffic rule of type speed limit.
+        //
+        optional SpeedLimit speed_limit = 3;
+
+        //
+        // The type of the the traffic rule.
+        //
+        enum TrafficRuleType {
+
+            // Traffic rule is of type speed limit
+            //
+            TRAFFIC_RULE_TYPE_SPEED_LIMIT = 0;
+        }
+
+        // 
+        // \brief Validity information for a traffic rule.
+        //
+        message TrafficRuleValidity {
+            
+            //
+            // The starting point of the traffic rule validity on the lane.
+            // Must be in range [\c sStart,\c sEnd] of the reference line.
+            //
+            optional double start_s = 1;
+
+            //
+            // The ending point of the traffic rule validity on the lane.
+            // Must be in range [\c sStart,\c sEnd] of the reference line.
+            // Note: Should always be greater than the starting point.
+            //
+            optional double end_s = 2;
+
+            //
+            // List of vehicle types for which the speed limit is valid.
+            // If the traffic rule validity is independent of the vehicle type 
+            // the list should be empty.
+            //
+            repeated MovingObject.VehicleClassification.Type valid_for_type = 3;
+        }
+
+        //
+        // \brief Speed limit on a lane.
+        //
+        message SpeedLimit {
+
+            //
+            // The speed limit in the unit specified by the respective files
+            //
+            optional double speed_limit = 1;
+
+            //
+            // The unit in which the speed limit is specified.
+            //
+            optional Unit speed_limit_unit = 2;
+
+            // Unit of the specified speed limit.
+            //
+            enum Unit
+            {
+                // Meters per second.
+                //
+                // Unit: km/h
+                //
+                UNIT_METER_PER_SECOND = 1;
+
+                // Kilometers per hour.
+                //
+                // Unit: km/h
+                //
+                UNIT_KILOMETER_PER_HOUR = 2;
+
+                // Miles per hour.
+                //
+                // Unit: mph
+                //
+                UNIT_MILE_PER_HOUR = 3;
+            }
+        }
+    }
+
 }
 
--- a/osi_logicallane.proto
+++ b/osi_logicallane.proto
@@ -934,7 +934,7 @@ message LogicalLane
 
             //
             // The value of the speed limit.
-            // The unit filed in the TrafficSignValue message may only be set to
+            // The unit field in the TrafficSignValue message may only be set to
             // units associated with velocities and must not be UNKNOWN or OTHER.
             //
             // Note: All speed limits are to be led this way, independent
--- a/osi_logicallane.proto
+++ b/osi_logicallane.proto
@@ -830,12 +830,12 @@ message LogicalLane
 
     //
     // Describes traffic rules on a lane. 
-    // The traffic rule can thereby induced by regulations, traffic signs 
+    // The traffic rule can thereby be induced by regulations, traffic signs 
     // or by other means. If the modeled traffic rule is induced by a traffic sign
     // the information should be identical with the respective traffic sign.
     //
     // Note: Every instance should be corresponding to only one specific rule.
-    // The type of the traffic rule should be set using the respective filed.
+    // The type of the traffic rule should be set using the respective field.
     // Additionally, every message should contain the traffic rule validity information
     // and the respective field for the respective traffic rule type.
     // In case of traffic rule (priority) conflicts for rules of the same type that 
--- a/osi_logicallane.proto
+++ b/osi_logicallane.proto
@@ -4,6 +4,7 @@ option optimize_for = SPEED;
 
 import "osi_common.proto";
 import "osi_object.proto";
+import "osi_trafficsign.proto";
 
 package osi3;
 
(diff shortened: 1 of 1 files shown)


osi_logicallane.proto
```diff
@@ -821,5 +826,117 @@ message LogicalLane
...
+    //
+    message TrafficRule {
+
+        // The type of the traffic rule.
+        //
+        // This specifies the type of the traffic rule to be modeled.
+        // Based on the type the respective message containing the information
+        // corresponding to the traffic rule should be filled.
+        //
+        optional TrafficRuleType traffic_rule_type = 1;
```
thempen, line 855: Because of the expendability of the type, the traffic_rule_type field could be repeated. 
Therefore you would be able to have mutliple rules for the same type of vehicles and ranges. You have also a framwork to manage colliding rules. E.g. "no overtaking" and "you are allowed to overtake tractors".
You could for example use the order of the repeated field as a priority to work on. The order is fixed for protobuf and would not change: [https://protobuf.dev/programming-guides/encoding/#optional]()

thempen: Discussion from Other Models group: This message should only contain the currently valid rule. There should not be the necessity for the agent model to select between contradictory rules.
This should be further clarified in the documentation.

thomassedlmayer:
@lemmer-fzi The enums (other than the fields numbers) should start with 0. You changed the one that was already correct to be wrong now as well ;)

osi_logicallane.proto
```diff
@@ -821,5 +826,120 @@ message LogicalLane
...
     }
+
+    //
+    // Describes traffic rules on a lane. 
+    // The traffic rule can thereby induced by regulations, traffic signs 
+    // or by other means. If the modeled traffic rule is induced by a traffic sign
+    // the information should be identical with the respective traffic sign.
+    //
+    // Note: Every instance should be corresponding to only one specific rule.
+    // The type of the traffic rule should be set using the respective filed.
```
thomassedlmayer, line 837: Typo: 'filed' should be 'field'.

osi_logicallane.proto
```diff
@@ -821,5 +826,120 @@ message LogicalLane
         //
         optional double end_s_other = 5;
     }
+
+    //
+    // Describes traffic rules on a lane. 
+    // The traffic rule can thereby induced by regulations, traffic signs
```
thomassedlmayer, line 832: "... can thereby **be** induced by..."

osi_logicallane.proto
```diff
@@ -821,5 +826,120 @@ message LogicalLane
...
+            // Note: Should always be greater than the starting point.
+            //
+            optional double end_s = 2;
+
+            //
+            // List of vehicle types for which the speed limit is valid.
+            // If the traffic rule validity is independent of the vehicle type 
+            // the list should be empty.
+            //
+            repeated MovingObject.VehicleClassification.Type valid_for_type = 3;
```
thomassedlmayer, line 901: Instead of directly refering to the classification type, we could also add a wrapper message which contains the vehicle classification type. In the case that vehicle classification will be extended by some kind(s) of subtype(s) at some point in the future, it could be easily added here as well.

osi_logicallane.proto
lemmer-fzi: Fixed with newest commit.

osi_logicallane.proto
```diff
@@ -821,5 +827,122 @@ message LogicalLane
...
+        }
+
+        //
+        // \brief Speed limit on a lane.
+        //
+        message SpeedLimit {
+
+            //
+            // The value of the speed limit.
+            // The unit filed in the TrafficSignValue message may only be set to
```
thomassedlmayer, line 936: *field

pmai:
CCB 2024-03-25: Ready for merge from CCB point, will be merged after final review form  @tbleher.

osi_logicallane.proto
```diff
@@ -822,5 +828,123 @@ message LogicalLane
...
+    //
+    // Note: Every instance should be corresponding to only one specific rule.
+    // The type of the traffic rule should be set using the respective field.
+    // Additionally, every message should contain the traffic rule validity information
+    // and the respective field for the respective traffic rule type.
+    // In case of traffic rule (priority) conflicts for rules of the same type that 
+    // can not be depicted using the traffic rule validity only the currently 
+    // valid rule should be provided.
+    //
+    // Note: Each traffic rule corresponds to only one lane. If the traffic rule
```
tbleher, line 846: Maybe a theoretical concern, but at least worth mentioning: Currently I see no way of defining for which travel direction the rule is valid? Speed limits are generally directional (valid for only one travel direction), if they are defined by signs. What happens if there is a `bidirectional` lane with different rules for different traffic directions?

OK, so hopefully noone is crazy enough to do something like this, and maybe OSI does not need to support this. But I thought I should at least mention it :)

tbleher (commented):
Thanks @pmai for the pointer, and @lemmer-fzi for the good work :)

I have added a few minor comments inline. Two general comments:

The first (maybe minor) comment: I currently see no way to explicitly define "no speedlimit" (for German highways). So I guess "no speedlimit" is modelled as the absence of a TrafficRule. This makes it impossible for a model, however, to know whether the simulation environment has provided traffic rules or not. I suggest explicitly modelling "no speedlimit".

The second comment might be a bigger issue: how are variable speed limits handled (e.g. from variable message boards, see type vmsBoard in OpenDRIVE 1.8)?
Typically the logical lanes would be part of `GroundTruthInit` and would be static. But speed limits can be dynamic. So it should (I think) at least be clarified what should happen if a speed limit changes (see https://github.com/OpenSimulationInterface/osi-sensor-model-packaging/pull/108#issuecomment-2023146437 for a few problems in that area).

Should the whole logical lane be sent again in the SensorView (which would be bad for performance)? Should an empty lane be sent, with just the limits set? What if the limit changes from e.g. 130km/h to "unlimited" (as pointed out above, then currently no `TrafficRule` would be set).

So I think the interaction between `TrafficRule`, `vmsBoard` and `GroundTruthInit` should be clarified before merge (note: as far as I am aware `vmsBoard` is not part of OSI yet, but I would expect a simulation environment which encounters a `vmsBoard` with an active speed limit on it to insert a speed limit sign into the `SensorView`).

osi_logicallane.proto
```diff
@@ -822,5 +828,123 @@ message LogicalLane
...
+
+        // 
+        // \brief Validity information for a traffic rule.
+        //
+        message TrafficRuleValidity {
+            
+            //
+            // The starting point of the traffic rule validity on the lane.
+            // Must be in range [\c sStart,\c sEnd] of the reference line.
+            //
```
pmai, line 888: Suggested change as per CCB 2024-04-04:

```suggestion
            //
            // Note: The traffic rule applies only to traffic with notional
            // direction of travel from the start_s coordinate towards
            // the end_s coordinate. For monodirectional lanes this
            // must match the direction of travel as specified by the
            // move_direction field of the logical lane. For bidirectional
            // lanes this allows the specification of separate rules for
            // each direction of travel.
            //
```

osi_logicallane.proto
```diff
@@ -822,5 +828,123 @@ message LogicalLane
...
+            optional double start_s = 1;
+
+            //
+            // The ending point of the traffic rule validity on the lane.
+            // Must be in range [\c sStart,\c sEnd] of the reference line.
+            // Note: Should always be greater than the starting point.
```
pmai, line 894: As per CCB 2024-04-04:

```suggestion
```

osi_logicallane.proto
```diff
@@ -822,5 +828,123 @@ message LogicalLane
...
+            message TypeValidity {
+
+                //
+                // The type of object for which the traffic rule is valid.
+                //
+                optional MovingObject.Type type = 1;
+                
+                //
+                // Vehicle classification type for traffic participants 
+                // Should be set to UNKNOWN if type is not TYPE_VEHICLE or the rule
```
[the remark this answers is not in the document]
pmai: Will be handled in review cleanup.

osi_logicallane.proto
```diff
@@ -822,5 +828,123 @@ message LogicalLane
...
+                // is valid for all vehicle types.
+                //
+                optional MovingObject.VehicleClassification.Type vehicle_type = 2;
+
+                //
+                // Role of traffic participant.
+                // Should be set to UNKNOWN if type is not TYPE_VEHICLE or the rule
```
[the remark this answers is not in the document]
pmai: Will be handled in review cleanup.

osi_logicallane.proto
pmai: To be handled by requiring the direction of travel to match the start_s/end_s order of the TrafficRuleValidity, see below.

pmai (changes requested):

pmai (approved):

pmai:
> The first (maybe minor) comment: I currently see no way to explicitly define "no speedlimit" (for German highways). So I guess "no speedlimit" is modelled as the absence of a TrafficRule. This makes it impossible for a model, however, to know whether the simulation environment has provided traffic rules or not. I suggest explicitly modelling "no speedlimit".

As discussed in today's CCB meeting, this is not considered a specific problem for the traffic rules, as this is a general problem for OSI (detection whether the absence of a field/entry is due to lack of support or no content), hence should be handled at a different level. Especially for the traffic rules, there is only a single instance where no speed limit exists, i.e. on motorways for certain vehicle types, hence there will always be a rule in place somewhere that makes this detection possible.

> The second comment might be a bigger issue: how are variable speed limits handled (e.g. from variable message boards, see type vmsBoard in OpenDRIVE 1.8)? Typically the logical lanes would be part of `GroundTruthInit` and would be static. But speed limits can be dynamic. So it should (I think) at least be clarified what should happen if a speed limit changes (see [OpenSimulationInterface/osi-sensor-model-packaging#108 (comment)](https://github.com/OpenSimulationInterface/osi-sensor-model-packaging/pull/108#issuecomment-2023146437) for a few problems in that area).

As discussed in today's CCB meeting, this is already an existing problem, as traffic signs are usually also considered static but can change dynamically (see traffic sign variability), and in consequence certain lane properties are likely to change (e.g. lane type from TYPE_STOP to TYPE_NORMAL for signs turning the hard shoulder into an extra normal lane).

So this whole topic will have to be handled as part of the work of determining dynamic/static parts and the handling of updates. We always knew that GroundTruthInit was a partial solution that needs more definition going forward, and dynamic signs (which already are a part of OSI) just make this more pressing.

> So I think the interaction between `TrafficRule`, `vmsBoard` and `GroundTruthInit` should be clarified before merge (note: as far as I am aware `vmsBoard` is not part of OSI yet, but I would expect a simulation environment which encounters a `vmsBoard` with an active speed limit on it to insert a speed limit sign into the `SensorView`).

As stated above, dynamic traffic signs have been a part of OSI for a long time, how to handle them in the face of GroundTruthInit has been unspecified for as long.

pmai:
@thempen I have added the discussed text and made some spelling/formatting fixes. Please approve if this matches your expectations.

thempen (approved):

ClemensLinnhoff:
The wording might be a bit misleading. Logical lanes are the equivalent of OpenDRIVE lanes:

> If OSI is generated from OpenDRIVE, then [LogicalLanes](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/gen/structosi3_1_1LogicalLane.html) map directly to OpenDRIVE lanes.

 But in OpenDRIVE the term "traffic rule" refers to left handed or right handed traffic ([see documentation](https://publications.pages.asam.net/standards/ASAM_OpenDRIVE/ASAM_OpenDRIVE_Specification/latest/specification/16_annexes/enumerations/road/top_enu_e_trafficRule.html)).

thempen:
This was discussed in the CCB on monday (6.5.2024):
"Traffic rule" is seen as a much more generic description, as only left handed or right handed traffic, therefore in our opinion, the wording in OpenDRIVE should be changed in the long run. (It can't be changed right now, because of the backwad compatibility)
Having this name, as it is, could theoretically mislead a user. The messages however are so different, that it is nearly impossible to implement it in the wrong way. 
There is a big overlap between OpenDRIVE and OSI, but there will be differences, when naming conventions will not be seen as useful.

PR #756, opened 2024-01-09
Workflow: Update Python and Actions + Black formatting Python files

jdsika:
* Add .venv/ and venv/ to .gitignore
* Bump Python minimum to 3.8 as out of support
* Update GitHub Marketplace Actions
* Add workflow to check black formatting
* Add requirements_develop.txt
* Apply black formatting

### DIFF
--- a/format/OSITrace.py
+++ b/format/OSITrace.py
@@ -10,9 +10,10 @@ from osi3.osi_sensorview_pb2 import SensorView
 from osi3.osi_groundtruth_pb2 import GroundTruth
 from osi3.osi_sensordata_pb2 import SensorData
 import warnings
-warnings.simplefilter('default')
 
-SEPARATOR = b'$$__$$'
+warnings.simplefilter("default")
+
+SEPARATOR = b"$$__$$"
 SEPARATOR_LENGTH = len(SEPARATOR)
 BUFFER_SIZE = 1000000
 
@@ -31,7 +32,7 @@ def get_size_from_file_stream(file_object):
 MESSAGES_TYPE = {
     "SensorView": SensorView,
     "GroundTruth": GroundTruth,
-    "SensorData": SensorData
+    "SensorData": SensorData,
 }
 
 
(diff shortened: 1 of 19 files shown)


format/OSITrace.py
PhRosenberger, line 1: If I see that correctly, this was only refactroing in this file, right?

format/osi2read.py
PhRosenberger, line 1: This was also just refactored, right? - Why do you change all the '-signs to "-signs?

format/OSITrace.py
jdsika: Yes, I applied black formatting the same as it is applied in osi-validation. It is done in order to keep the same coding style throughout projects.

format/osi2read.py
jdsika: This is the way

format/OSITrace.py
PhRosenberger: Ok fine, thx

PhRosenberger (approved):
I am fine with the code refactoring and black formatting.
I cannot help that much with the workflow changes.
Hope to get some more detailed feedback from @pmai there.

pmai:
I have created a separate PR for non-python related build workflow changes, that also adds support for newer protobuf releases (upto current 25.x releases), leaving this PR for python-related changes to workflow and python code changes, if wanted.

The apt-get update dance should be dropped, since we should never depend on anything more specific from the base system.

I'm not a huge fan of the black reformatting (or usually any automatic reformattings), as it just seems to change one set of idiosynchracies for another, but whatever floats anyones boat (I think the code should be rewritten in places to make it clearer, something the reformatting really doesn't address).

pmai:
I've refactored to built upon the other workflow enhancements, so once those are merged only the python-related changes will remain in this PR.

jdsika:
> I have created a separate PR for non-python related build workflow changes, that also adds support for newer protobuf releases (upto current 25.x releases), leaving this PR for python-related changes to workflow and python code changes, if wanted.
> 
> The apt-get update dance should be dropped, since we should never depend on anything more specific from the base system.
> 
> I'm not a huge fan of the black reformatting (or usually any automatic reformattings), as it just seems to change one set of idiosynchracies for another, but whatever floats anyones boat (I think the code should be rewritten in places to make it clearer, something the reformatting really doesn't address).

not that I disagree with your stance towards formatting but in our case this will be done the same way as in the osi-validation and if a lot of people work on something it is helping to provide a consistent visual.

pmai:
CCB 2014-01-15: Merge as-is.

PR #766, opened 2024-01-22
Support more protobuf versions in GitHub workflow

### LINKED ISSUE #765: Check different Protobuf versions
#### Describe the bug
I suspect, that the pipeline will fail for some protobuf versions. I will investigate on a branch.

I expect the pipeline wto work for all protobuf versions >= 2.6.1

ClemensLinnhoff:
#765

#### Add a description
Fix pipeline behavior with different protobuf versions


ClemensLinnhoff:
Protobuf 2.6.1 does not work out of the box, since it does not have '-all' in the link/package:
https://github.com/OpenSimulationInterface/open-simulation-interface/actions/runs/7609941012

If '-all' is changed to '', it will download abseil, but cannot install it, since protobuf 2.6.1 has autotools (configure). So the if condition distinguishing between autotools and cmake  is not correct:
https://github.com/OpenSimulationInterface/open-simulation-interface/actions/runs/7609969004

ClemensLinnhoff:
With my changes, protobuf 2.6.1 can be installed. But now installing python fails:
https://github.com/OpenSimulationInterface/open-simulation-interface/actions/runs/7610124644

ClemensLinnhoff:
> Proto3 versions below don't work because of a compiler error. See for example https://github.com/OpenSimulationInterface/open-simulation-interface/actions/runs/7613993384/job/20735224384

I fixed this by setting CXXFLAGS=-std=c++11 for configure.

Now, only protobuf 2.6.1 does not work.

ClemensLinnhoff:
I tried again with protobuf 2.6.1. The problem occurs during building the python files form the .proto files.

When running `protoc --python_out osi3/ osi3/osi_version.proto` the following error occurs:

```
google/protobuf/descriptor.proto: File not found.
```

In osi_version.proto, google/protobuf/descriptor.proto is included. However I checked, and it is located in /usr/include and /usr/local/include. So I have no idea, why it is not found and why this only is an issue with protobuf 2.6.1 and not the other versions.

Has anyone ever used the combination osi - protobuf 2.6.1 - python?
@pmai do you have any idea what might be the issue?

jdsika:
Another sanity check: Do older versions of protobuffer also work with e.g. C++17? I would guess yes

ClemensLinnhoff:
> Another sanity check: Do older versions of protobuffer also work with e.g. C++17? I would guess yes

For the older versions I had to explicitly set a CXXFlag to c++11. But this works for all versions.

pmai:
> I now managed to get the setup.py to work with protobuf 2.6.1. However, now a unit test fails. Apparently Protobuf 2.6.1 does not support Python 3...
> 
> My recommendation: Let's set the minimum proto version to 3.0.0, as it already says in the documentation anyways.

At least officially protobuf states Python 3 support since 2.6.0... The error logs look more like there are changes in the location of the bundled proto include?

jdsika:
> > I now managed to get the setup.py to work with protobuf 2.6.1. However, now a unit test fails. Apparently Protobuf 2.6.1 does not support Python 3...
> > My recommendation: Let's set the minimum proto version to 3.0.0, as it already says in the documentation anyways.
> 
> At least officially protobuf states Python 3 support since 2.6.0... The error logs look more like there are changes in the location of the bundled proto include?

What about this here? https://github.com/protocolbuffers/protobuf/issues/882

ClemensLinnhoff:
Python is now completely decoupled from the protobuf version. So the changes to setup.py  are superseded. 
What is the status on proto 2 support for C++? Then we might need the changes. However, only the CI pipeline has minor changes.

asadekasam:
ASAM OSI CCB 2025-05-08: Merge is on hold until refactoring of repos is settled 
https://github.com/OpenSimulationInterface/open-simulation-interface/issues/864

PR #767, opened 2024-01-23
Fix streaming principle image reference

thomassedlmayer:
Image on streaming update principle is not displayed because its file extension has upper case letters.


pmai (approved):
CCB 2024-04-04: Merge as-is.

PR #774, opened 2024-02-12
Fix typos in proto files

### LINKED ISSUE #773: Typo in osi3::TrafficAction::EndActionsAction documentation
#### Describe the bug

There is a typo in the [detailed documentation of osi3::TrafficAction::EndActionsAction](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/V3.5.0/gen/structosi3_1_1TrafficAction_1_1EndActionsAction.html#_detailed_description): "exection" instead of "execution".

ClemensLinnhoff:
fixes #773

#### Add a description
There were a lot of typos in the proto files. I fixed the ones I found.

Two things I am unsure of and therefore did not fix yet:
1) Unfortunately, I found one spelling mistake in an enum in [osi_object.proto](https://github.com/OpenSimulationInterface/open-simulation-interface/blob/master/osi_object.proto#L256): It should by TYPE_GLASS and not TYPE_GLAS. Furthermore, sometimes British English is used in enums, e.g. COLOR_GREY. I suppose fixing it would break backwards compatibility?
-> Enum names shall not be changed in a minor release
2) In [osi_lane.proto](https://github.com/OpenSimulationInterface/open-simulation-interface/blob/master/osi_lane.proto#L13) the term antecessor is used in multiple places. Is this really the correct term? I am more used to ancestor in this context.


PhRosenberger:
As I also struggle with both questions, who else could support here?
@ThomasNaderBMW @pmai @jdsika @thempen

jdsika:
Ancestor would be a term that I see for human relations like "My ancestors were brave fighters in the galaktic resistance" whereas "antecessor" is often used for e.g. "the Ferrari 5 series is the antecessor of the Ferrari 567". The usage in combination with objects would make it more suitable in my option for our digital objects like lanes but ... I am not a native speaker. Most important in this case would be "does everyone understand what is meant by it"?
Regarding the "GLASS" and "GLAS" ... It would in fact break compatibility... Let's bring it up for a vote. I think we have fixed blunt mistakes in the past even without a major version change?

osi_object.proto
```diff
@@ -251,7 +251,7 @@ message StationaryObject
             //
             MATERIAL_STONE = 6;
 
-            // Glas structure.
+            // Glass structure.
             //
             MATERIAL_GLAS = 7;
```
jdsika, line 256: The enum is correct. Just the comment is wrong? Not an issue imo

jdsika (approved):

ClemensLinnhoff:
The cambridge dictionary does not even list antecessor: https://dictionary.cambridge.org/spellcheck/english/?q=antecessor

Wiktionary says it is a rare synonym of ancestor: https://en.wiktionary.org/wiki/antecessor

I personally have never heard of it until now, but I am also not a native speaker.

ClemensLinnhoff:
As for the typo of MATERIAL_GLASS, I am in favor of fixing it. I suspect, that this enum field is not that commonly used anyways. So the backwards compatibility issues will most likely be limited.

ClemensLinnhoff:
Furthermore, we have inconsistencies with British and US englisch. Mostly, US is used, but in some cases, unfortunately not only in comments, we have British, e.g. COLOR_GREY instead of COLOR_GRAY.

ClemensLinnhoff:
What does FMC stand for [here](https://github.com/OpenSimulationInterface/open-simulation-interface/blob/master/osi_hostvehicledata.proto#L587)? It is never explained.

ClemensLinnhoff:
I now added a spellcheck pipeline and fixed a lot more spelling mistakes. I also had to add quite a lot of exceptions, because of abbreviations and also a lot of German words.

PhRosenberger:
> What does FCW stand for [here](https://github.com/OpenSimulationInterface/open-simulation-interface/blob/master/osi_hostvehicledata.proto#L587)? It is never explained.

I guess FCW stands for Forward Collision Warning in this case.

osi_sensorview.proto
```diff
@@ -287,15 +287,15 @@ message LidarSensorView
 
         // normal to surface angle.
         //
-        // The normal of the transmitted beam to the object, roadmarking etc
+        // The normal of the transmitted beam to the object, road marking etc
```
pmai, line 290:
```suggestion
        // The normal of the transmitted beam to the object, road marking, etc.
```

osi_environment.proto
```diff
@@ -176,7 +176,7 @@ message EnvironmentalConditions
     // \par References:
     // [1] Shepard, F. D. (1996). <em>Reduced visibility due to fog on the highway.</em> Transportation Research Board, National Research Council (Ed.). National Academy Press. Washington, D.C., USA. ISBN 0-309-06006-0. \n
     // [2] Strassenverkehrs-Ordnung (StVO) as of dated  March 06, 2013 (BGBl. I S. 367), lastly changed by article 4a of the order from June 06, 2019 (BGBl. I S. 756). \n
-    // [3] stvo.de. (2013, April 01). <em>StVO Par. 17 Beleuchtung</em>. Retrieved January 25, 2020, from https://www.stvo.de/strassenverkehrsordnung/101-17-beleuchtung \n
+    // [3] StVO.de. (2013, April 01). <em>StVO Par. 17 Beleuchtung</em>. Retrieved January 25, 2020, from https://www.stvo.de/strassenverkehrsordnung/101-17-beleuchtung \n
```
pmai, line 179: While domain names are not case-sensitive, so that both spellings would work here, I do not think that changing this makes sense in general.

```suggestion
    // [3] stvo.de. (2013, April 01). <em>StVO Par. 17 Beleuchtung</em>. Retrieved January 25, 2020, from https://www.stvo.de/strassenverkehrsordnung/101-17-beleuchtung \n
```

pmai (commented):
Otherwise looks good to me. Spellchecker potentially needs more fine-tuning, and we cannot change enum names per-se in a minor release. We might be able to add alias names, however would first have to check protobuf behavior.

ClemensLinnhoff:
> Otherwise looks good to me. Spellchecker potentially needs more fine-tuning, and we cannot change enum names per-se in a minor release. We might be able to add alias names, however would first have to check protobuf behavior.

Thank you for the review, I documented the decision about leaving the enums in the PR description. I fixed the spelling mistakes in the comments belonging to the enums, but not the enums themselves.

From my point of view now the only open point is the use of antecessor vs. ancestor.

pmai:
CCB 2024-02-26: As antecessor is used also in e.g. the LanePairing field antecessor_lane_id a change here would require a major release. Therefore antecessor should stay as the term in the documentation as well. @pmai will investigate whether adding proper spellings for TYPE_GLAS and COLOR_GREY as aliases would work. If yes, they can be added. The existing enums have to stay in in any case to remain backward-compatible. After the above changes and finalization this PR is ready for merge.

ClemensLinnhoff:
I added antecessor to the custom word list. Now the spell check passes.

pmai:
I have added aliases for the correct spellings of COLOR_GRAY and MATERIAL_GLASS, retaining the old COLOR_GREY and MATERIAL_GLAS enums with deprectation warnings. This should work for proto2/proto3, and the documentation generation; however please check whether this change breaks anything else...

osi_common.proto
```diff
@@ -699,13 +699,13 @@ message ColorDescription
 }
 
 //
-// \brief Greyscale color model
+// \brief Grayscale color model
 //
-// ColorGrey defines a greyscale.
+// ColorGrey defines a grayscale.
 //
 message ColorGrey
```
jdsika, line 706: @pmai do we need an Alias here as well?

jdsika: Wait, let me guess: Alias only works for ENUMS and not signals? Should we add a TODO here? Does it bother?

jdsika (commented):
Possible changes during review v3.7.0

Issue #733, opened 2023-06-27
Virtual Detection Area in SensorData


jruebsam:
#### Virtual Detection Area

Currently it is not possible from a SensorData message to get an understanding 
what the current area of observation of a sensor  model is. 

I would like to have some kind of message, lets say a Virtual Detection Area
which describes in which kind of FOV the current sensor models operates.
This could be a simble generic message which contains  a repeated list of points, e.g. 

### Solution

```
message VirtualDetectionArea
{
    // List of points of the boundary of the detection area sorted counter-clockwise 
   // relative to sensor mounting position, projected onto ground surface
    repeated Vector2d boundary_point = 1;
}
```

Since this message is related to the output of a sensor model I would like to add this to the SensorData message.
Also this would be important for our current usecase, which is focused on visualization of SensorData messages.

PhRosenberger:
@jruebsam if I understand you correctly, you would like to know what the simulated sensor could have seen (a.k.a. nominal FoV) in comparison to the actual output data? I wonder why it is restricted to a ground projection and not 3D or even 4D (incl. velocity or intensity). However, do you know our concept of a "Unified relevance region from sensor model knowledge" from this paper: https://tuprints.ulb.tu-darmstadt.de/18950/ ? - This seems related to your thoughts, right?

Are you attending the GSVF next week? - Would be great to discuss it there in person:-)

jruebsam:
@PhRosenberger  THanks for the reply, yes it would be something like a nominal FOV. In our usecase we want to use this in a visualizer thats why its only 2D, since 3D FOVs can become quiet confusing. Unfornutanely I will not be able to join GSVF, but also thanks for the paper.

PhRosenberger:
@jruebsam Oh, what a pity that we cannot meet this week!

However, for your use case, I see the existing GenericSensorView as an option: https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/gen/structosi3_1_1GenericSensorView.html

What do you think?

jruebsam:
@PhRosenberger I thought about that, however GenericSensorView only contains a FOV and a Mounting Position but not a Range. It could be an option that we do some kind of extension at this location.

thomassedlmayer:
[...]

PR #781, opened 2024-02-26
feat: Add virtual detection area

jruebsam:
[733](https://github.com/OpenSimulationInterface/open-simulation-interface/issues/733)

#### Add a description
see Issue

### DIFF
--- a/osi_common.proto
+++ b/osi_common.proto
@@ -1089,3 +1089,15 @@ message KeyValuePair
     //
     optional string value = 2;
 }
+
+//
+// \brief Polygon in 3 dimensions
+//
+// A polygon in 3 dimensions which contains a list of vertices.
+//
+message Polygon3d
+{
+    // A list of vertices
+    //
+    repeated Vector3d vertex = 1;
+}
--- a/osi_sensordata.proto
+++ b/osi_sensordata.proto
@@ -313,4 +313,23 @@ message SensorData
     // in cartesian coordinates.
     //
     optional LogicalDetectionData logical_detection_data = 27;
+
+    //
+    // \brief Virtual detection area of a sensor
+    //
+    // The virtual detection area describes the area the sensor can see in a best case scenario
+    // where no occlusion or other statistical effects are considered. It is described by a set
+    // of polygons.
+    //
+    message VirtualDetectionArea
+    {
+        // List of polygons. Each polygon represents a surface of the virtual detection area
+        // and is given with respect to the virtual sensor coordinate system.
+        //
+        repeated Polygon3d polygon = 1;
+    }
+
+    // Virtual detection area of the sensor
+    //
+    optional VirtualDetectionArea virtual_detection_area = 28;
 }
--- a/osi_sensordata.proto
+++ b/osi_sensordata.proto
@@ -317,9 +317,13 @@ message SensorData
     //
     // \brief Virtual detection area of a sensor
     //
-    // The virtual detection area describes the area the sensor can see in a best case scenario
-    // where no occlusion or other statistical effects are considered. It is described by a set
-    // of polygons.
+    // The virtual detection area describes the nominal area the sensor is capable of covering
+    // in its current operating mode, without taking occlusion or other statistical effects into account.
+    // This information can be used for visualization or other development purposes as a rough guide
+    // to nominal sensor performance.
+    //
+    // It is described by a set of polygons in cartesian coordinates as a pragmatic approximation for
+    // the rough shapes expected.
     //
     message VirtualDetectionArea
     {


osi_sensordata.proto
```diff
@@ -313,4 +313,22 @@ message SensorData
     // in cartesian coordinates.
     //
     optional LogicalDetectionData logical_detection_data = 27;
+
+    //
+    // \brief Virtual detection area of a sensor
+    //
+    // The virtual detection area describes the are the sensor can see in a best case scenario
```
thomassedlmayer, line 320: Typo: area

osi_sensordata.proto
```diff
@@ -313,4 +313,22 @@ message SensorData
...
+    // where no occlusion or other statistical effects are considered. It is described by a set
+    // of polygons.
+    //
+    message VirtualDetectionArea {
+        // List of polygons. Each polygon represents a surface of the virtual detection area
+        // and is given with respect to the virtual sensor coordinate system.
+        //
+        repeated Polygon3d polygons = 1;
```
thomassedlmayer, line 328: Repeated field names should be singular.

osi_sensordata.proto
jruebsam: Removed.

osi_common.proto
```diff
@@ -1089,3 +1089,15 @@ message KeyValuePair
...
+//
+// \brief Polygon in 3 dimensions
+//
+// A polygon in 3 dimensions which contains a list of vertices.
+//
+message Polygon3d
+{
+    // A list of vertices
+    //
+    repeated Vector3d vertices = 1;
```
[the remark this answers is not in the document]
jruebsam: Fixed.

osi_sensordata.proto
```diff
@@ -313,4 +313,23 @@ message SensorData
     // in cartesian coordinates.
     //
     optional LogicalDetectionData logical_detection_data = 27;
+
+    //
+    // \brief Virtual detection area of a sensor
+    //
+    // The virtual detection area describes the area the sensor can see in a best case scenario
```
PhRosenberger, line 320: I would appreciate some more clarification and this point. Some thoughts to discuss:

1. What about ghost targets from multipath propagation that are actually outside the virtual detection area and appear inside?
2. I am very critical of the term "best case scenario". Is the idea of the virtual coverage area to eventually account for degradation due to environmental effects or not? - I thought it was possible, but the term here suggests the opposite. What do you think about our following definition from the [PerCollECT paper](https://tuprints.ulb.tu-darmstadt.de/18949/1/2021_ITSC_Specify_Sensor_Model_PrePrint.pdf) (p. 2) to use for "best case"? 

> During emission, propagation, reception, and signal processing, the information in the signal deviates from the original information about the environment, existing as ground truth (GT) under clean room conditions. Clean room conditions describe idealized measurement conditions, without the influence of weather, dust, dirt etc. The GT is taken in this clean room environment by an omniscient observer.

osi_common.proto
```diff
@@ -1089,3 +1089,15 @@ message KeyValuePair
...
+//
+// \brief Polygon in 3 dimensions
+//
+// A polygon in 3 dimensions which contains a list of vertices.
+//
+message Polygon3d
+{
+    // A list of vertices
+    //
+    repeated Vector3d vertex = 1;
```
PhRosenberger, line 1102: Why did you choose Vector3d / cartesian coordinates for the Polygon3d? - I suggest to use Spherical3d instead, because imho it is more common to use angles and ranges to describe the field of view of a sensor.

PhRosenberger: Actually, I wonder why to use Polygon3d anyways, as you could describe the FoV with repeated Spherical3d coordinates already. Did you insert this intermediate step with the surfaces for a specific reason?

In other words: Why do we limit the Virtual detection area to be a composition of surfaces rather then a more flexibel solution? - Sensors like the [Lovox Tele-15](https://www.livoxtech.com/de/tele-15) definetly need more flexibility to be described.

PhRosenberger (changes requested):
In general, I support the idea of the virtual detection area.

However, some more discussion is needed. Let's discuss the open points and questions I raised, shall we?

osi_sensordata.proto
thomassedlmayer: Would it be a problem if this message represents what the sensor model thinks about its own range in the current time frame under the current conditions? I mean, if the sensor model does not take such conditions into account (for its output/for its detection area), the area may as well stay the same all the time and be very basic. The requirements could vary depending on the use case.

Also, if I think about manufacturer specifications of sensors, they probably do not provide such clean room ranges/FoV's but rather realistic ranges under average conditions. They might even cut off the range by software. For certain use cases such "expected" ranges/FoV's are probably better base lines for visualization for comparison with the corresponding sensor data.

thomassedlmayer: Suggestion: The virtual detection area describes a nominal view of the sensor.

osi_common.proto
thomassedlmayer: I think, Vector3d was used because that's what 3d engines offer for setting up such polygons. Converting this to spherical coordinates for transmission is just another step, I guess. So for our use case of visualizing such area Vector3d is completely fine.

The use case for this PR is just a quite basic visualization of such sensor cones, so we did not want it to become more complicated than it has to be. But I would agree, that this the intention of this detection area is mentioned in the description more specifically.

jruebsam: I would say in general the Vector3d approach is more generic.
Using spherical3d defines something which always contains the origin at 0.
So for defining some abstract surface in 3d it will not be possible to use Spherical3d.

The question is  how generic should it be ?

On the other side as Thomas mentioned the Vector3d Format is much better suited towards
rendering applications.

osi_sensordata.proto
pmai:
```suggestion
    // The virtual detection area describes the nominal area the sensor is capable of covering
```

osi_sensordata.proto
```diff
@@ -313,4 +313,23 @@ message SensorData
...
+    // where no occlusion or other statistical effects are considered. It is described by a set
+    // of polygons.
```
pmai, line 322:
```suggestion
    // in its current operating mode, without taking occlusion or other statistical effects into account.
    // This information can be used for visualization or other development purposes as a rough guide
    // to nominal sensor performance.
    //
    // It is described by a set of polygons.
```

pmai (approved):
The prosposed wording changes might make the intent clearer.

osi_common.proto
jruebsam: edit: I just realized you can also exclude 0 from spherical ...

pmai:
CCB 2024-03-25: Merge with suggested changes.

PhRosenberger (approved):

jdsika:
@pmai 
We have the base_polygon. We should have defined this at the time?

```
//
// \brief Polygon in 2 dimensions
//
// A polygon in 2 dimensions which contains a list of vertices.
//
message Polygon2d
{
    // A list of vertices
    //
    repeated Vector2d vertex = 1;
}
```

Reviewed for v3.7.0

PR #783, opened 2024-02-28
Modernize python build setup, protoc dependency

pmai:
This moves most meta-data to pyproject.toml, automates the version handling and makes it available at runtime, depends on protoc via the protoc-wheel-0 package, makes sdist and bdist distributions work correctly, and removes dependencies that are not build dependencies, but rather used only in CI.


pyproject.toml
```diff
@@ -0,0 +1,35 @@
...
+name = "open-simulation-interface"
+description = "ASAM Open Simulation Interface Python Bindings."
+authors = [
+    {name = "ASAM Open Simulation Interface Project", email = "osi@asam.net"},
+]
+maintainers = [
+    {name = "ASAM Open Simulation Interface Project", email = "osi@asam.net"},
+]
+dependencies = [
+    "protobuf>=4.24.4",
```
jdsika, line 19: better == ?

pyproject.toml
```diff
@@ -0,0 +1,35 @@
+[build-system]
+requires = [
+    "setuptools",
+    "wheel",
+    "protoc-wheel-0==24.4",
```
ClemensLinnhoff, line 5: In the protobuf.yml, the protobuf version in configurable as an environment variable. Currently it is set to 3.20.1:

```
PROTOBUF_VERSION: 3.20.1
```
How does this fit to the hardcoded protobuf version here, and also in the dependencies below?
In [my PR](https://github.com/OpenSimulationInterface/open-simulation-interface/pull/766/) regarding this topic I tried to make the python protobuf version coherent with the installed protoc version on the system.

ClemensLinnhoff:
Unit tests need to be fixed. There seem to be some issues due to the renaming of the format folder, see [pipeline run](https://github.com/OpenSimulationInterface/open-simulation-interface/actions/runs/8095361160/job/22121877911?pr=783).

jdsika:
Proto 3.6.1 in use for OSI-Validation at BMW

pyproject.toml
pmai: Conservatively `==` would likely be better, however there might be some leeway here, which is why I have for now left it at `>=`, but might change to `==` prior to finalizing the PR.

pyproject.toml
pmai: The protobuf.yml is just our test build pipeline, i.e. how _we_ test any changes, especially also the C++ side.
This does not hard-code any actual requirements, it just controls our build environment so that our build environment stays half-way stable.

The pyproject.toml will define how others build the python packages. Since this requires easy access to a suitable protoc (and a matching runtime), we build-depend the python package on a fixed version of protoc-wheel-0, which we then also have to use as a dependency for the protobuf runtime. Since none of this works very well with dynamic dependencies, this more or less has to be hard-coded. In the end we might converge on one version between CI and python package, but that is really independent.

This PR prepares the complete separation of the C++ and python build processes, which should have happened in the first place, and also - in a later PR - separates our CI from the package requirements.

We really, really do not want to have hard version requirements across projects, so osi-validator can just depend on the osi3 python package with a >= requirement, and not have any dependencies on protobuf at all.

ClemensLinnhoff: > The protobuf.yml is just our test build pipeline.

Yes, but we now test the C++ build with 3.20.1 and the Python build with 24.4.

Just for my understanding: To build OSI for C++, I need to have the protoc compiler installed on my system, with any protobuf version. The python build is independent of that protoc compiler and it always uses version 24.4 (unless of course I change the toml file). Is that correct?

pmai: > > The protobuf.yml is just our test build pipeline.
> 
> Yes, but we now test the C++ build with 3.20.1 and the Python build with 24.4.

Yes, and we might or might not switch to something newer on the C++ side, however that is now independent of the Python side (since for Python it means what we ship, not what we test, whereas for C++ it only means what we test). In any case the C++ changes would be handled in a separate PR.

> Just for my understanding: To build OSI for C++, I need to have the protoc compiler installed on my system, with any protobuf version. The python build is independent of that protoc compiler and it always uses version 24.4 (unless of course I change the toml file). Is that correct?

Yes, though the python build scripts have a way of falling back to system- or user-supplied protoc, if you want that. Just by default they will build-depend on the protoc-wheel-0 package and use that for building.

thomassedlmayer:
@pmai In `formatting_scripts.adoc` there is a note that `converted.txth` is the default name of the output file. This should be changed as now the default name is the input file name.

jdsika:
@pmai will this be ready to merge next CCB?

pmai:
> @pmai In `formatting_scripts.adoc` there is a note that `converted.txth` is the default name of the output file. This should be changed as now the default name is the input file name.

Yes, that was however already wrong prior to this PR, so I would suggest fixing in separate PR.

pmai:
CCB 2025-03-25: Merge as-is.

jdsika:
I was a bit confused about the diff at first but this was a mix of moving files and linter. so OK

Reviewed for v3.7.0

PR #787, opened 2024-03-08
Deprecate various car types in favor of new alias TYPE_CAR

pmai:
This is a first result of the harmonization effort across ASAM Open standards

In preparation of additional types we already deprecate some obvious candidates for removal, namely our plethora of car types.

### DIFF
--- a/osi_object.proto
+++ b/osi_object.proto
@@ -819,7 +819,7 @@ message MovingObject
             //
             // Definition: Hatchback or sedan with length between 4.5 and 5 m.
             //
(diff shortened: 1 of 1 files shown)


jakobkaths (approved):

tbleher:
The change looks good to me 👍

pmai:
CCB 2024-04-04: Merge as-is.

thempen (approved):
The changes let the type better align to other standards. It will still be backward compatible and can be used in the new way through the alias functionality. So I do not see a downside. Therefore I approve this PR.

jdsika (approved):

PR #789, opened 2024-03-15
Add system time to sensor data

thomassedlmayer:
This PR adds a system/component-specific time to OSI SensorData. This could be used to simulate discrepancies between simulation time frame and component time frame. It could also be used to preserve the component's internal timestamps when using OSI for real sensor measurement data while having the simulation timestamp's zero time set to simulation/trace start (as it is recommended by OSI).

Outcome from WP 19: Real sensor data


osi_sensordata.proto
```diff
@@ -313,4 +313,11 @@ message SensorData
     // in cartesian coordinates.
     //
     optional LogicalDetectionData logical_detection_data = 27;
+
+    // The system time of the sensor data, given in UTC (Unix Epoch timestamp).
+    //
+    // The system time can be used to set the corresponding component's internal
+    // time which may not coincide with the simulation time frame.
```
pmai, line 320:
```suggestion
    // The system time of the modeled source of the sensor data, given
    // in UTC (Unix Epoch timestamp).
    //
    // The system time can be used to transmit the internal time of the
    // simulated component that supplies the sensor data, which might
    // not coincide with the simulation time as transmitted in the
    // timestamp field. Example use cases include recorded data traces
    // or the simulation of time synchronization mechanisms and errors.
```

pmai (approved):
Looks good with the proposed rewording to better set expectations.

thempen (approved):
Looks good. I don't see any downsides in introducing this field.

PhRosenberger:
The only remark I have is that this time could be different in different sensors that are running on the same car, so putting these fields in each DetectionData could provide the flexibility to reflect such scenarios. I mean, when there are multiple lidars witch more than 1 [LidarDetectionData](osi3::LidarDetectionData), each could have different real world time stamps, right?
Let's further discuss in the meeting :-)

PhRosenberger (commented):
The only remark I have is that this time could be different in different sensors that are running on the same car, so putting these fields in each DetectionData could provide the flexibility to reflect such scenarios. I mean, when there are multiple lidars witch more than 1 LidarDetectionData, each could have different real world time stamps, right?
Let's further discuss in the meeting :-)

pmai:
> The only remark I have is that this time could be different in different sensors that are running on the same car, so putting these fields in each DetectionData could provide the flexibility to reflect such scenarios. I mean, when there are multiple lidars witch more than 1 LidarDetectionData, each could have different real world time stamps, right? Let's further discuss in the meeting :-)

That would be an additional change, if that is wanted; right now the system time is the time of the source of the sensordata, so for a single sensor it is its time, after e.g. sensor fusion it is the system time of the sensor fusion unit, etc. If someone wants to keep track of all of this across, and need to funnel this into OSI rather than measuring at multiple points, and they need something different than the detection time stamps, then another field could be added later on. A lot of ifs though.

pmai:
CCB 2024-03-25: Merge as-is after conflicts resolved.

jdsika:
Reviewed for v3.7.0

PR #790, opened 2024-03-18
Improve omit_static_information field description

FabianPfeuffer:
﻿#### Reference to a related issue in the repository
#770

#### Add a description
A slight improvement and additional information for the description of the SensorViewConfiguration's omit_static_information data field.

#### Take this checklist as orientation for yourself, if this PR is ready for the Change Control Board:
- [x] My suggestion follows the [style and contributors guidelines](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/specification/contributing/start_contributing.html).
- [x] I have taken care about the [message documentation](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/specification/contributing/commenting_messages.html) and the [fields and enums documentation](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/specification/contributing/commenting_fields_enums.html).
- [x] I have done the [DCO signoff](https://opensimulationinterface.github.io/osi-documentation/open-simulation-interface/doc/howtocontribute.html#developer-certification-of-origin-dco).
- [x] My changes generate no errors when passing CI tests. 
- [x] I have successfully implemented and tested my fix/feature locally.
- [x] Appropriate reviewer(s) are assigned.

PR #795, opened 2024-03-22
Add pyspelling html filter

thomassedlmayer:
I noticed that adding new hyperlinks to the proto files requires to add any character sequences from the hyperlink as exceptions to the spellchecker (see #790). So I found that it is possible to add a html filter that excludes such character sequences.

[Here](https://github.com/thomassedlmayer/open-simulation-interface/commit/0b376bce907a40fdebc05d480a204625f8db9b61) I added a exemplary `<a>`-tag with a hyperlink that contains strings that should trigger a spellchecker fail. With the new html filter the hyperlink does not trigger the fail anymore but now it fails because it now detects a lot more strings that contain mutated vowels which were not detected before. I guess the html filter converts those special characters so that the spellchecker can also check those words.

@ClemensLinnhoff Could you please check if this type of filter does not screw things up? Was there any reason why this was not added before? If it works, we should probably add the "new" mutated vowel-words as exceptions but remove any exceptions that were previously added because they are contained inside hyperlinks (or also html tags/attribute names like `cellspacing`, `href`, `colspan` etc.). Also, wouldn't it make sense to add German as secondary language so that we don't have to add every German word to the exception list?

### DIFF
--- a/.github/.pyspelling.yml
+++ b/.github/.pyspelling.yml
@@ -17,3 +17,8 @@ matrix:
       delimiters:
       - open: '(?s)^(?P<open> *\/\/)'
         close: '^(?P=open)$'
+  - pyspelling.filters.html:
+      comments: true
+      attributes:
+      - title
+      - alt
--- a/.github/spelling_custom_words_en_US.txt
+++ b/.github/spelling_custom_words_en_US.txt
@@ -20,6 +20,7 @@ auch
 auf
 auml
 Ausfahrt
+außer
 Baugebiet
 Baustelle
 Baustellen
@@ -89,6 +90,7 @@ dt
 Duesseldorf
 Durchgangsverkehr
 dvr
+dürfen
 easting
 EBIKES
 edn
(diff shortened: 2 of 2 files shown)


ClemensLinnhoff:
> but now it fails because it now detects a lot more strings that contain mutated vowels which were not detected before.

Seems like it. So either we add all these words with the Umlauts to the spelling_custom_words_en_US.txt or we need to figure out how to check multiple languages. I would opt to add them to the custom words and avoid using non-english words in the future.

thomassedlmayer:
I see. If there is no straightforward way of adding another language to the checker, let's maybe postpone this for now.

I added the necessary exceptions to the whitelist so that the spellchecker does not fail anymore. There may still be some words in the white list that are not required with the html filter in place, like html-tags. We can also maybe get rid of those.

pmai (approved):
CCB 2024-04-04: Merge as-is.

jdsika (commented):
Reviewed for v3.7.0

Issue #797, opened 2024-03-26
Missing ParkingSlot/spot type in osi3::StationaryObject::Classification


yonivai:
#### Describe the feature

We didn't find an appropriate representation in OSI3 for a parking spot for parking lot scenario.
after discussing with @PhRosenberger , we think that osi3::StationaryObject is a good representation for it.
unfortunately, we didn't find a type for parkingSlot in osi3::StationaryObject::Classification.

#### Describe the solution you would like
we would like to add a 19th Enum type to osi3::StationaryObject::Classification - "TYPE_PARKING_SPOT".
#### Describe alternatives you have considered
N/A
#### Describe the backwards compatibility
N/A
#### Additional context
https://opensimulationinterface.github.io/open-simulation-interface/structosi3_1_1StationaryObject_1_1Classification.html#ad497488a7119617633b54181ea08457d

Issue #714, opened 2023-02-15
Add a parking space object


schmidtlorenz:
#### Describe the feature
In OpenDrive parking spaces are defined as objects. Such an object is missing in OSI.

#### Describe the solution you would like
There needs to be an object representing parking spaces in OSI which aligns with the description in OpenDrive.
One way would be as Stationary Object. However, Stationary Objects might be falsely identified as hazard or road blockage.
Hence, RoadMarking would be more fitting. This might also allow to partially implement access rules for parking spaces via the traffic sign description in RoadMarking. The remaining problem is to find a way to express the upcoming (OpenDrive 1.8) feature of a "parking space ID" (as seen in ,e.g., a car park).

An alternative would be to add a new class "ParkingSpace", that could be structured according to the implementation in OpenDrive.

#### Describe the backwards compatibility
I see no backwards compatibility issues.

thomassedlmayer:
Harmonization WG 2023/03/01: Use cases are not clear yet. Could you give some more information on the use case? Is this a requirement for sensor modeling/traffic participant modeling/AD function? What information on the parking spaces is required? Also, what about moving parking spaces? @tbleher @schmidtlorenz 
From harmonization point of view we did not see the need to just adapt whatever there is in OpenDrive without further reasons. In terms of physical lanes it would also be interesting how that would be handled in the ISO23150.

tbleher:
One use case is a traffic agent based on OSI which wants to park autonomously. It needs to "see" all parking spaces so it can choose one. I would definitely be interested in having parking spaces (as they are in OpenDRIVE right now) also represented in OSI (though unfortunately right now I have little time to author a PR).

thomassedlmayer:
There will be a ASAM project tackling the parking space issues in a harmonized way in OSI, OpenDrive and OpenScenario DSL/XML, in case you want to join: https://code.asam.net/simulation/openx/-/issues/1

PR #798, opened 2024-03-26
Update StationaryObject::Classification::Type with new field for park…

yonivai:
https://github.com/OpenSimulationInterface/open-simulation-interface/issues/797

### DIFF
--- a/osi_object.proto
+++ b/osi_object.proto
@@ -217,6 +217,10 @@ message StationaryObject
             // in the environment, like street lights.
             //
             TYPE_EMITTING_STRUCTURE = 18;
+            
+            // Object is a parking spot.
+            //
+            TYPE_PARKING_SPOT = 19;
         }
 
         // Definition of material types.


thomassedlmayer:
This was already discussed some time ago (see #714) but we did not come to any conclusion because the drivers of the discussion dropped out or lost interest. However, as Lorenz said, having free space marked as object in OSI may introduce problems. @PhRosenberger @yonivai Did you consider this in your discussion?
What is the use case for this?

PhRosenberger:
>  @PhRosenberger @yonivai Did you consider this in your discussion? What is the use case for this?
I was not aware of #714, however the use case is to actually detect a parking spot as a bounding box, so this is something like a stationary object to me, even if it is a special case, as you can actually drive into it. On the other hand, there are stationary objects like a newspaper on the street that are kind of similar. This is the reason, why I would vote for adding this type to OSI.

yonivai:
> This was already discussed some time ago (see #714) but we did not come to any conclusion because the drivers of the discussion dropped out or lost interest. However, as Lorenz said, having free space marked as object in OSI may introduce problems. @PhRosenberger @yonivai Did you consider this in your discussion? What is the use case for this?

Hi @thomassedlmayer,
as @PhRosenberger said, we weren't aware of this discussion. 
I do think that Ground Truth wise there is place for a parking spot to be represented. 
if we are looking at a used case scenario, we are seeing it as a static representation of the world (like a map characteristics , before the scenario begins they already exists, and will be for the rest of the scenario).
also, we already have some types that a vehicle can drive onto it in the types of the stationary object:
https://opensimulationinterface.github.io/open-simulation-interface/structosi3_1_1StationaryObject_1_1Classification.html#ad497488a7119617633b54181ea08457da0035009db5b2fadfc746e432cd4d3eb9
https://opensimulationinterface.github.io/open-simulation-interface/structosi3_1_1StationaryObject_1_1Classification.html#ad497488a7119617633b54181ea08457da3061d9cf215369f5d91dad750d08e05d
https://opensimulationinterface.github.io/open-simulation-interface/structosi3_1_1StationaryObject_1_1Classification.html#ad497488a7119617633b54181ea08457daebee7560a0a5b63846221860fcf64321

yonivai:
Hi, is there any news on this matter? 
will we progress on that? or should we discuss about it?

thomassedlmayer:
As the last meeting of the responsible working group took place before this PR was opened, this unfortunately won't make it into the release. The review phase for the 3.7.0 release has started today.
But we may pursue this issue in future harmonization efforts. There is a working group developing non-normative harmonization specifications for various ASAM standards (e.g., OpenDrive, OpenScenario, OSI). This group is currently looking at traffic participant harmonization, but the aim would be to take forward other issues such as parking space representation in the future. I think it makes sense to check if the representation of a parking space as stationary object is still appropriate in other standards before adapting it.

PhRosenberger:
I would like to add something to @thomassedlmayer's last comment:
As mentioned, there are several working groups to consider for this topic.
Harmonization is important and I have been informed that other standards groups such as OpenDRIVE are also discussing related issues with respect to parking use cases.

@yonivai we would propose to use this internally as already suggested, as we also see a good chance to integrate it as suggested by you.
However, it was not possible to include it in the upcoming release due to the already started review phase. So we would encourage to participate in the next project phase towards the next release of OSI.

yash-shah-asam:
@thomassedlmayer : Could you please take care of this once the direction the project of parking lots and garages is done.

PR #800, opened 2024-04-03
Update and harmonize python installation instructions

pmai:
Adjust python install instructions to be more uniform across platforms and to reflect availability of packages on PyPI.

### DIFF
--- a/doc/setup/installing_linux_python.adoc
+++ b/doc/setup/installing_linux_python.adoc
@@ -4,12 +4,31 @@ include::{root-path}_config.adoc[]
 endif::[]
 = Installing OSI for Python on Linux
 
+== Installing from PyPI
+
+Steps::
+
+. Open a terminal.
++
+. Optionally create and activate a new virtual environment.
++
+----
+python3 -m venv venv
+source venv/bin/activate
+----
++
+. Install Open Simulation Interface.
++
+----
+pip install open-simulation-interface
+----
+
+== Installing from source
+
 *Prerequisites::
 
 * You have installed everything described in <<top-installing-osi-prerequisits>>.
-* You have installed _pip3_.
-* You have installed _python-setuptools_.
-* For a local installation, you have installed _virtualenv_.
+* You have installed _pip_.
 
 Steps::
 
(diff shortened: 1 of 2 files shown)


doc/setup/installing_linux_python.adoc
```diff
@@ -26,27 +45,22 @@ git clone https://github.com/OpenSimulationInterface/open-simulation-interface.g
...
 ----
+python3 -m venv venv
 source venv/bin/activate
 ----
 +
 . Install Open Simulation Interface.
 .. Local installation
 +
 ----
-python3 -m pip install .
```
jdsika, line 45: I thought you only use it with the prefix: python3 -m pip*

doc/setup/installing_linux_python.adoc
```diff
@@ -26,27 +45,22 @@ git clone https://github.com/OpenSimulationInterface/open-simulation-interface.g
...
+pip install .
 ----
 +
 .. Global installation
 +
 ----
-sudo pip3 install .
+sudo pip install .
```
jdsika, line 65: Do you really sudo?

doc/setup/installing_linux_python.adoc
pmai: Only in specific circumstances (e.g. upgrading pip in place, or where there was a multitude of python installations 2 vs 3 - also somewhat ubuntu specific); on most systems pip install is the suggested way to do stuff, and generally is e.g. the way PyPI recommends it.

Both ways work, but as there already was a mixture between the two, I picked one to stay consistent.

doc/setup/installing_linux_python.adoc
pmai: I don't ;), but then I never install globally anyway.

More seriously: That was part of the existing instructions, and yes, if you want the package installed globally then you likely need root permissions on most systems. On Ubuntu e.g. if you do not have root permissions, it will just install for the user.

Not that I would recommend anything like that, hence the local installation is the first suggested one.

pmai:
CCB 2024-05-08: Merge as-is.

PhRosenberger (approved):

PR #801, opened 2024-04-05
Fix osi trace robustness

pmai:
﻿As identified by @ClemensLinnhoff there is a precedence error in retrieve_offsets which can lead to errors if the function is called again after a completed read.

This PR fixes this bug and adds a test case to detect such errors.


ClemensLinnhoff (approved):

thomassedlmayer (approved):

pmai:
CCB 2024-05-06: Merge as-is.

Issue #805, opened 2024-05-02
Check comments for mandatory fields


ClemensLinnhoff:
#### Describe the bug
In some fields it is still noted as a comment if they are mandatory, e.g. \note This field is mandatory instead of using the rule is_set

#### Describe the expected behavior
Always use rule is_set when field is considered mandatory.

PR #806, opened 2024-05-03
Add rules according to notes in the documentation

ClemensLinnhoff:
#805 

#### Add a description
Added rules according to existing notes in the documentation.

### DIFF
--- a/osi_common.proto
+++ b/osi_common.proto
@@ -235,7 +235,7 @@ message ExternalReference
     
     // The type of the external references.
     //
-    // Value describing the type of the original source.
+    // Mandatory value describing the type of the original source.
     //
     // For OpenX/ASAM standards it is specified as follows:
     // - net.asam.opendrive
--- a/osi_common.proto
+++ b/osi_common.proto
@@ -249,7 +249,7 @@ message ExternalReference
     // is_set
     // \endrules
     //
-    optional string type = 2;    
+    optional string type = 2;
     
     // The external identifier reference value.
     //
--- a/osi_common.proto
+++ b/osi_common.proto
@@ -235,7 +235,7 @@ message ExternalReference
     
     // The type of the external references.
     //
-    // Mandatory value describing the type of the original source.
+    // Value describing the type of the original source.
     //
     // For OpenX/ASAM standards it is specified as follows:
     // - net.asam.opendrive
@@ -244,6 +244,10 @@ message ExternalReference
     // For third-party standards and user-defined objects, 
     // reverse domain name notation with lower-case type field
     // is recommended to guarantee unique and interoperable identification.
+    //
+    // \rules
+    // is_set
+    // \endrules
     //
     optional string type = 2;    
     
--- a/osi_route.proto
+++ b/osi_route.proto
@@ -42,7 +42,6 @@ message Route
     //
     // \rules
     // is_set
-    // is_globally_unique
     // \endrules
     //
     optional Identifier route_id = 1;
--- a/osi_route.proto
+++ b/osi_route.proto
@@ -42,6 +42,7 @@ message Route
     //
     // \rules
     // is_set
+    // is_globally_unique
     // \endrules
     //
     optional Identifier route_id = 1;
--- a/osi_trafficcommand.proto
+++ b/osi_trafficcommand.proto
@@ -144,7 +144,6 @@ message TrafficAction
         //
         // \rules
         // is_set
-        // is_globally_unique
         // \endrules
         //
         optional Identifier action_id = 1;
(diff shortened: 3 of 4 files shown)


PhRosenberger (approved):

osi_common.proto
```diff
@@ -235,7 +235,7 @@ message ExternalReference
     
     // The type of the external references.
     //
-    // Mandatory value describing the type of the original source.
+    // Value describing the type of the original source.
```
pmai, line 238:
```suggestion
    // Mandatory value describing the type of the original source.
```

The rules need to be redundant to the specification.

osi_route.proto
```diff
@@ -42,6 +42,7 @@ message Route
     //
     // \rules
     // is_set
+    // is_globally_unique
```
pmai, line 45: I think this is not correct, as globally unique requires the ID to be unique across all messages of all types, whereas the current requirement is just uniqueness within all route messages (to one TP), which also makes more sense.

So the validator would need to properly handle a rule that checks for this kind of scoped uniqueness, which it currently does not; hence

```suggestion
```

osi_trafficcommand.proto
```diff
@@ -142,6 +142,11 @@ message TrafficAction
         // \note This id must be unique within all traffic command
         // messages exchanged with one traffic participant.
         //
+        // \rules
+        // is_set
+        // is_globally_unique
```
pmai, line 147: Dito, see above:

```suggestion
```

pmai (changes requested):
Looks good otherwise, assuming that the added rules have been checked to work in the validator.

osi_route.proto
ClemensLinnhoff: @pmai I was too quick in committing this. On second thought: Why would anything be unique across all messages of all types? If this is really the intention, we already have a problem: moving object ID in the ground truth is set as is_globally_unique, as well as sensor id in the sensor_view. Does your interpretation mean, that if I have a sensor_id = 1 I cannot have a moving object with id = 1? This does not make sense to me.

So in my interpretation is_globally_unique means, that the uniqueness is related to that specific field. Either way, we definitely need a section in the documentation describing the rules to avoid these different interpretations.

ClemensLinnhoff:
> Looks good otherwise, assuming that the added rules have been checked to work in the validator.

Yes, I checked that the rules are correctly parsed into the yml files by rules2yml of osi-validation.

osi_route.proto
pmai: It means, as you say, that all of the globally unique IDs must be unique across all messages/fields, as also implemented in the osi-validation rule implementation. I agree that for sensor_id it probably does not make sense, but it stems from the fact that we have moving object, stationary object, traffic signs and traffic lights, etc., and initially it was thought that it would be better to keep all of these IDs non-overlapping, hence the is_globally_unique rule (also ask @jdsika for the reasoning/discussions at the time).

This probably still makes sense, especially as there are still plans to merge those classes at some time. However for other ID fields (especially newly introduced ones), something more strict makes more sense. But we have no rule for this as of yet.

pmai (approved):

osi_route.proto
ClemensLinnhoff: Okay, I understand. But then we should remove globally unique from all non-object IDs, such as sensor ID. I can make the change here.

pmai:
CCB 2024-05-06: Merge as-is.

Potentially further rule additions that are direct result of existing textual rules in the release could be added in maintenance/patch releases, depending on our future approach to maintenance and maintenance releases.

ClemensLinnhoff:
> CCB 2024-05-06: Merge as-is.
> 
> Potentially further rule additions that are direct result of existing textual rules in the release could be added in maintenance/patch releases, depending on our future approach to maintenance and maintenance releases.

Before you megre, should we remove is_globally_unique from the sensor_ids then? Or is this considered a breaking change since the rule already existed in 3.6?

PhRosenberger:
> Before you megre, should we remove is_globally_unique from the sensor_ids then? Or is this considered a breaking change since the rule already existed in 3.6?

Let's not rush this and discuss the rules that are already there in a separate issue/PR. We are not sure if this is a quick fix for a patch or if it needs further discussion..

ClemensLinnhoff:
> > Before you megre, should we remove is_globally_unique from the sensor_ids then? Or is this considered a breaking change since the rule already existed in 3.6?
> 
> Let's not rush this and discuss the rules that are already there in a separate issue/PR. We are not sure if this is a quick fix for a patch or if it needs further discussion..

But if it is the way @pmai describes, a lot of trace files will definitely fail during validation. It is quite common to have sensor_id set to 0 und the host vehicle also having ID 0. I would consider the rule is_globally_unique in the sensor_id a bug.

ClemensLinnhoff:
Since my previous comment has been ignored and this PR has been just merged, I have created a new issue for the documentation bug: #809

PR #812, opened 2024-05-10
Remove is_globally_unique rule from messages, where it does not make sense

### LINKED ISSUE #809: Description of Identifier rules wrong in documentation
#### Describe the bug

This is the definition of the [Identifier](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/gen/structosi3_1_1Identifier.html) in the documentation:

> Has to be unique among all simulated items at any given time. For ground truth, the identifier of an item (object, lane, sign, etc.) must remain stable over its lifetime. [Identifier](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/gen/structosi3_1_1Identifier.html) values may be only be reused if the available address space is exhausted and the specific values have not been in use for several time steps. Sensor specific tracking IDs have no restrictions and should behave according to the sensor specifications.

It says, that every ID shall be globally unique except for sensor specific tracking IDs. This does not make much sense. A sensor ID is completely independent from object IDs or route IDs. Furthermore, sensor IDs depend on the sensor setup and are independent from the scenario, where the object IDs are defined. Same with the route IDs: They are defined in the map and are independent from sensor setup and scenario.

I propose:

- Set rule is_globally_unique in all fields that are actually supposed to be globally unique (NOT sensor ID or tracking ID for example).
- Change description of Identifier to state that the ID is only globally unique, if specified by the rule is_globally_unique.
- Optional: Introduce new rule is_unique to specify uniqueness within the individual field

ClemensLinnhoff:
#809 

#### Add a description
Removed rule is_globally_unique from messages, where a globally unique ID does not make any sense.

Some IDs depend on the scenario, e.g. object IDs in Ground Truth. Some IDs depend on the map, e.g. lane IDs. And some IDs depend on the sensor setup, e.g. sensor IDs.

I can't think of a reason, why there shouldn't be a moving object with ID 0, a lane with ID 0 and a sensor with ID 0. Therefore, I removed the is_globally_unique rule from every message, that is not a ground truth object.

### DIFF
--- a/osi_sensorview.proto
+++ b/osi_sensorview.proto
@@ -64,11 +64,6 @@ message SensorView
     // object output; it is distinct from the IDs of its physical detectors,
     // which are used in the detected features.
     //
-    // \rules
-    // is_globally_unique
-    // is_set
-    // \endrules
-    //
     optional Identifier sensor_id = 3;
 
     // The virtual mounting position of the sensor (origin and orientation of
--- a/osi_sensorview.proto
+++ b/osi_sensorview.proto
@@ -64,6 +64,10 @@ message SensorView
     // object output; it is distinct from the IDs of its physical detectors,
     // which are used in the detected features.
     //
+    // \rules
+    // is_set
+    // \endrules
+    //
     optional Identifier sensor_id = 3;
 
     // The virtual mounting position of the sensor (origin and orientation of
--- a/osi_streamingupdate.proto
+++ b/osi_streamingupdate.proto
@@ -82,5 +82,7 @@ message StreamingUpdate
     // Entities that will no longer be updated, because they are considered
     // obsolete by the sender.
     //
+    // \note Referenced IDs are globally unique.
+    //
     repeated Identifier obsolete_id = 9;
 }
--- a/osi_streamingupdate.proto
+++ b/osi_streamingupdate.proto
@@ -82,7 +82,5 @@ message StreamingUpdate
     // Entities that will no longer be updated, because they are considered
     // obsolete by the sender.
     //
-    // \note IDs are globally unique.
-    //
     repeated Identifier obsolete_id = 9;
 }
--- a/osi_lane.proto
+++ b/osi_lane.proto
@@ -41,7 +41,6 @@ message Lane
     // \note Note ID is global unique.
     //
     // \rules
-    // is_globally_unique
     // is_set
     // \endrules
     //
@@ -713,7 +712,6 @@ message LaneBoundary
     // The ID of the lane boundary.
     //
     // \rules
-    // is_globally_unique
     // \endrules
     //
     optional Identifier id = 1;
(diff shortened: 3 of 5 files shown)


osi_streamingupdate.proto
```diff
@@ -82,7 +82,5 @@ message StreamingUpdate
     // Entities that will no longer be updated, because they are considered
     // obsolete by the sender.
     //
-    // \note IDs are globally unique.
-    //
```
pmai, line 86: This should stay, as the IDs this applies to are globally unique.

osi_sensorview.proto
```diff
@@ -64,11 +64,6 @@ message SensorView
     // object output; it is distinct from the IDs of its physical detectors,
     // which are used in the detected features.
     //
-    // \rules
-    // is_globally_unique
-    // is_set
```
pmai, line 69: is_set still seems correct

pmai (changes requested):
See comments, these seem overbroad in some places.

In any case I think just removing the rules and statements, meaning that the uniqueness is now completely undefined seems like the wrong way to go about it. I'd rather refine the definition of globally unique to clarify that cross-type uniqueness is only meant for the groundtruth objects, leaving globally unique to still mean unique across simulation time.
How to handle this the best in terms of the rule names is another question.

osi_streamingupdate.proto
ClemensLinnhoff: But then this prone to be misunderstood. Because the IDs this points to are unique, but because per definition these IDs already exist, this ID field, where the note is, is not unique.

osi_streamingupdate.proto
```diff
@@ -82,7 +82,5 @@ message StreamingUpdate
     // Entities that will no longer be updated, because they are considered
     // obsolete by the sender.
     //
```
ClemensLinnhoff, line 84:
```suggestion
    //
     \note IDs of ground truth sub-messages (e.g. moving objects) are globally unique.
```

ClemensLinnhoff: How about something like this to make it more clear?

thomassedlmayer: The description of StreamingUpdate states that this field just references an existing object's globally unique id.
Suggestion: "Referenced IDs are globally unique."
I agree that the "reference" part comes a bit short in this field description.

ClemensLinnhoff: I added your suggestions @thomassedlmayer

osi_streamingupdate.proto
ClemensLinnhoff: I fixed it with the suggestion of @thomassedlmayer in the conversation below.

yash-shah-asam:
@pmai Could you please discuss this with the OSI Checker bundle group?

Issue #392, opened 2020-06-17
SensorViewConfiguration missing timestamp


caspar-ai:
#### Describe the bug
There is a disconnect between the documentation and the implementation with regards Top-Level Messages and `SensorViewConfiguration`.

#### Describe how to reproduce the bug
The [interface conventions documentation](https://opensimulationinterface.github.io/osi-documentation/open-simulation-interface/doc/interfaceconventions.html?highlight=toplevel#top-level-messages) states that all top level messages require an interface as the first field and a timestamp as the second.  It also explicitly mentions `SensorViewConfiguration` as one such example of a top level message.

The [definition](https://github.com/OpenSimulationInterface/open-simulation-interface/blob/master/osi_sensorviewconfiguration.proto#L69) of a `SensorViewConfiguration` includes an interface version but does not include a timestamp.

#### Describe the expected behavior
The expected fix depends on the desired behaviour:

1. If the `SensorViewConfiguration` is an exception to the top level rule then I would expect the documentation to be updated, and a comment added inline by the message definition.
2. If it isn't an exception, a timestamp field should be added.

Krishna626:
@caspar-ai Hi Casper, while going through the SensorViewConfiguration [definition](https://github.com/OpenSimulationInterface/open-simulation-interface/blob/master/osi_sensorviewconfiguration.proto#L69), I could find the simulation start time at line 191. Isn't it what you are expecting?

caspar-ai:
@Krishna626 I'm not an expert in this area, so I'm not 100% sure.  On first read though it doesn't look like it to me.

In particular, compare to, for example, the [ground truth definition](https://github.com/OpenSimulationInterface/open-simulation-interface/blob/master/osi_groundtruth.proto#L50) which defines the timestamp as the _current_ time.  Whereas the field you point to on line 191 defines it as the _start_ time of the simulation, i.e. that field would never change throughout the lifetime of the exchange.  At least that's how I read it.

thomassedlmayer:
As this issue was brought up by @PhRosenberger in #764 I tried to get some insight on the problem. 

It seems like the part on mandatory 'version' and 'timestamp' fields in top level messages was removed from the documentation at some point. I could imagine that it isn't too important for SensorViewConfiguration to include a timestamp since it is either sent prior to simulation start (during initialization) or contained in another top level interface message which already includes a timestamp.

[...]

PR #833, opened 2024-10-14
828 MCAP specification for osi tracefiles

### LINKED ISSUE #828: MCAP example for OSI tracefiles
#### Describe the feature
One or multiple OSI tracefiles should be packaged within an MCAP container [MCAP repository](https://github.com/foxglove/mcap). MCAP is now used for ros2 and is MIT licensed. Recently the requirement for signing a CLA with Foxglove Inc. was lifted. Therefore the broad usage is now unproblematic.

#### Describe the solution you would like
We analyse the meta data used and proposed in the MCAP project and map it to the meta data information used in the OSI project. The osi tracefile writer python script is extended to also wirte these MCAP traces. The documentation of osi regarding traces is extended with the example and usage guidelines for MCAP. A good usecase are e.g. traffic participant models which have multiple osi trace outputs which are time synchronized and could profit from being stored in a single container.

#### Describe alternatives you have considered
Consider mentioning the possibility of using ASAM MDF as container. Remark: MCAP is used broadly in simulation context already.

#### Describe the backwards compatibility
No issues. We should name the tracefiles contained within the MCAP container according to the existing naming conventions.

TimmRuppert:

Closes https://github.com/OpenSimulationInterface/open-simulation-interface/issues/828

#### Add a description

- Add an initial support for mcap files as an alternative to tradtional .osi and .txth files 
- Precision of .osi and .txth spec
- Removed example code as this here is a spec and not the implementation
- Mentioning of a companion repo for utils which implement this spec (separation of spec and tutorials/examples/code)

**Some questions to ask**:
What is this change?
- Spec change to support mcap
How has it been tested?
- Feature. Breaking changes for naming convention are currently in discussion..


doc/architecture/trace_file_binary_and_human_readable_formats.adoc
```diff
@@ -2,11 +2,16 @@ ifndef::include-only-once[]
...
 
-**Name format**
+== Binary .osi Format
+Messages are separated by a four-byte, little-endian, unsigned integer specifying the length of each message.
 
-The names of OSI trace files should have the following format:
+== Human-readable .txth Format
+Messages are stored as plain text, separated by newlines.
+
+== Naming Convention
```
ClemensLinnhoff, line 13: The naming convention should be recommended for all trace file formats (incl. mcap).

ClemensLinnhoff:
For metadata discussion: timestamp definitions in OSI:
- Common Timestamp: https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/gen/structosi3_1_1Timestamp.html
- SensorView: https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/gen/structosi3_1_1SensorView.html#ad30bef352de1b64dd211718c18c540f1
- SensorData.timestamp: https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/gen/structosi3_1_1SensorData.html#a82adb4a7724d1e4835ae0921fab4080c
- SensorData.system_timehttps://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/gen/structosi3_1_1SensorData.html#ae64496913ffa0c41f154260cf48725e6

doc/architecture/trace_file_mcap_format.adoc
```diff
@@ -0,0 +1,39 @@
...
+- `data` field: String-encoded `google::protobuf::FileDescriptorSet` for the OSI top-level message
+
+== Channel
+- `message_encoding` field: Must be "protobuf"
+- `metadata` field: Should include a brief description of data origin and purpose
+
+== Message
+- `publish_time` field: 
+  - Must reflect the timestamp of the stored OSI top-level message if it contains a timestamp field
+  - Should represent a plausible value to express message order in relation to timestamps of other top-level messages if the message doesn't contain a timestamp (e.g. SensorViewConfiguration)
```
TimmRuppert, line 20: Remove this sentence for now as `SensorViewConfiguration` has no real timestamp, see  discussion in #392

doc/architecture/trace_file_mcap_format.adoc
```diff
@@ -0,0 +1,39 @@
+= MCAP Format
+
+== General Requirements
+- Must comply with the https://mcap.dev/spec[MCAP format specification] version `0x30`
+- Must allow other non-OSI data to be present in the MCAP file
+- Message records must be written into `chunk records` for indexed files
+
```
TimmRuppert, line 7: Add note explaining that only top-level messages with timestamp can be saved

doc/architecture/trace_file_mcap_format.adoc
```diff
@@ -0,0 +1,39 @@
...
+  * `protobuf`: SemVer version of protobuf library
+- Must include metadata with the name `creation_date` containing at least the following key-value pairs:
+  * `timestamp`: ISO 8601 formatted creation time
+- Must include metadata with the name `description` containing at least the following key-value pairs:
+  * `text`: Scenario description
+- Additional custom metadata may be added
+
+== Compression
+- OSI-compliant tooling must support compression types: `none`, `lz4`, and `zstd`
+- Recommended compression in use should be either `lz4` or `zstd`
```
TimmRuppert, line 39: Remove sentence as utilities will support a default compression.

doc/architecture/trace_file_mcap_format.adoc
```diff
@@ -0,0 +1,39 @@
...
+  - Should represent a plausible value to express message order in relation to timestamps of other top-level messages if the message doesn't contain a timestamp (e.g. SensorViewConfiguration)
+  - Must be in nanoseconds
+- `log_time` field: Must reflect the time when the message was enqueued for MCAP file addition
+  - Must reflect the time when the top-level OSI message was enqueued for addition to the MCAP file
+  - Zero time doesn't need to coincide with the Unix epoch, but it is recommended
+  - Must be in nanoseconds
+
+== Metadata
+- Must include metadata with the name `versions` containing at least the following key-value pairs:
+  * `osi`: SemVer version of OSI specification
```
ClemensLinnhoff, line 29: This shall be the highest compatible OSI version of all OSI messages in the MCAP file.

doc/architecture/trace_file_mcap_format.adoc
```diff
@@ -0,0 +1,39 @@
...
+  * `protobuf`: SemVer version of protobuf library
```
ClemensLinnhoff, line 30: Do we need the protobuf version at all or can we remove them? Also from the file name specification of the .osi format.

doc/architecture/trace_file_mcap_format.adoc
```diff
@@ -0,0 +1,39 @@
...
+- Must include metadata with the name `creation_date` containing at least the following key-value pairs:
+  * `timestamp`: ISO 8601 formatted creation time
```
ClemensLinnhoff, line 32: Add description of the exact time this timestamp shall represent

doc/architecture/trace_file_mcap_format.adoc
```diff
@@ -0,0 +1,39 @@
...
+- Message records must be written into `chunk records` for indexed files
+
+== Schema
+- `name` field: Full message type name, including package (e.g., `osi3.SensorData`)
+- `encoding` field: Must be `protobuf`
+- `data` field: String-encoded `google::protobuf::FileDescriptorSet` for the OSI top-level message
+
+== Channel
+- `message_encoding` field: Must be "protobuf"
+- `metadata` field: Should include a brief description of data origin and purpose
```
TimmRuppert, line 15: Add OSI and protobuf version as metadata for each channel for easier debugging etc. with one file supporting channels with different versions.

jdsika:
I would (rather quickly) need an example MCAP file for OSI v3.7.0 with (at best all) objects filled according to standard.

TimmRuppert:
> I would (rather quickly) need an example MCAP file for OSI v3.7.0 with (at best all) objects filled according to standard.

Here a three example files: [Example_OSI_MCAP.zip](https://github.com/user-attachments/files/17391322/Example_OSI_MCAP.zip)
- `Only_SD_v1.mcap` contains multiple sensor data on one channel (topic)
- `Two_SD_different_rate_v1.mcap` contains multiple sensor data on **two** channels
- `All_Top-Level_with_Timestamp_v1.mcap` contains all top-level messages which have a timestamp field 

I quickly tried to adapt most outcomes of the last meeting.

doc/architecture/trace_file_binary_and_human_readable_formats.adoc
TimmRuppert: For documentation purposes, the conclusion of the meeting was that further discussion is required. There are differing opinions on whether a naming convention for MCAP files should be recommended, mandatory or not be required at all. Some support this requirement (to ease the work with external partners), while others are concerned about the complications it may introduce, such as handling situations where trace files are created dynamically over and over again while working on other tasks (as trace files are merely a tool, not the end goal).

If a naming convention for MCAP files is included in the specification, several considerations must be addressed:

- An MCAP file can technically store messages (in different channels) from various OSI versions.
- Including the Protobuf version in the naming may not offer any substantial advantage.
- Should the "number of frames" reflect only OSI data, or should it also account for non-OSI data that might be present in the MCAP?
- How should the "type" be defined if multiple channels contain different types of top-level messages (a proposed solution is to use "multi"), or if there are multiple instances of the same kind of top-level message?
- What point in time should the "timestamp" represent? We agreed to wait for https://github.com/OpenSimulationInterface/open-simulation-interface/issues/834

TimmRuppert:
> Thanks! Some ground truth data from esmini with objects etc from as many types as possible would be great :))

I just updated esmini and it seems like there is an issue with the FMU.. Anyways, for starters here is one of our highway examples where I converted a native tracefile SensorView to an mcap file (the ground truth is therefore a submessage)

[demo_SV_onramp_V1.zip](https://github.com/user-attachments/files/17408401/demo_SV_onramp_V1.zip)

I can provide a GT top level message as well but I am a bit tight on schedule for today and tomorrow.

jdsika:
No, thank you. It was just important for me to have a "third set of eyes" creating a file in order to debug something. Otherwise everyone is always blaming in circles :)

jdsika:
FYI: OSI TRacefile writer in [openPASS](https://gitlab.eclipse.org/eclipse/openpass/gt-gen-core/-/blob/main/Core/Service/TraceWriter/trace_writer.cpp)

doc/architecture/trace_file_mcap_format.adoc
TimmRuppert: Done. Precised text.

doc/architecture/trace_file_mcap_format.adoc
TimmRuppert: Done. Removed text.

doc/architecture/trace_file_mcap_format.adoc
TimmRuppert: Done but only added osi version.

TimmRuppert:
I have updated the spec based on our last discussion. In the next meeting we need to address the following things :  

- File naming convention (see https://github.com/OpenSimulationInterface/open-simulation-interface/pull/833#discussion_r1803044372)
- Requirement of the protobuf version in the meta and/or the naming convention
- Clarify timestamp details based on https://github.com/OpenSimulationInterface/open-simulation-interface/issues/834
  - timestamps in the file name according to the naming convention
  - in the `publish_time` field of the MCAP message
- General review and potential new things somebody would like to be considered

TimmRuppert:
Documenting today's meeting concerning the points mentioned above: 

> * File naming convention (see [828 mcap example for osi tracefiles #833 (comment)](https://github.com/OpenSimulationInterface/open-simulation-interface/pull/833#discussion_r1803044372))

- We agreed on
  - `<opt prefix> ` To help sort after maneuver or similar
  - `<opt timestamp>` To help sort after recording time (if no prefix) or quickly identify a recording. Should represent abs. timestamp of zero-time of top-level messages (synchronized global time)
  - `<required type>` Like for .txth and .osi but additionally `multi`
  - `<opt suffix>` Further details you might want to provide without having to inspect the .mcap metadata (e.g. if applicable a min. required OSI version)

> * Requirement of the protobuf version in the meta and/or the naming convention

- move to the channel metadata 
- might not be used by anyone except for some debugging edge-cases but does not hurt either 


> * Clarify timestamp details based on [Documentation page for timing #834](https://github.com/OpenSimulationInterface/open-simulation-interface/issues/834)
>   
>   * timestamps in the file name according to the naming convention
>   * in the `publish_time` field of the MCAP message
> * General review and potential new things somebody would like to be considered
- log_time = publish_time = timestamp of top-level message

jdsika:
Consider the following case:
We will have one mcap trace with a lot of information and traces and a user wants to add osi traces to it. This has as a consequence that as much information as possible must be placed in the trace/channel meta data and not

doc/architecture/trace_file_binary_and_human_readable_formats.adoc
TimmRuppert: During today's meeting we proposed a convention (see comments below or updated files).  Thus resolving this thread.

doc/architecture/trace_file_mcap_format.adoc
TimmRuppert: During today's meeting we agreed to store it the channels metadata (see comments below or updated files).  Thus resolving this thread.

thomassedlmayer:
List of potential optional metadata fields that mainly emerged from the Gaia-X project (focus on measurement data):

- [ ] (1) **Start/stop timestamp / duration** (include/exclude depending on mcap capabilities of extracting those from the mcap timestamps)
- [ ] (2) **Frame rate** (@ClemensLinnhoff You mentioned frame rate as an interesting metadata field. I thought about how to handle variable frame rates: Maybe define it to be an approximated frame rate in the case of variable frame rates?)
- [ ] (3) **Content granularity:** Level of granularity of contained data, e.g. trace contains object lists / detection lists (could also be extended to flag if a trace contains (detected) lane network, traffic lights, traffic signs, environmental conditions)
- [ ] (4) **Traffic direction:** Left-hand/right-hand
- [ ] (5) **Location:** E.g. country, state, city, etc. (could also be put into trace description)
- [ ] (6) **Contained road types:** Indicate that road types like motorway/rural/city/traffic calmed zone etc. are contained in the trace
- [ ] (7) **Contained lane types:** Indicate that biking/walking/parking lanes are contained in the trace
- [ ] (8) **Scenario identifier:** User-/company-/project-specific scenario identifiers
- [ ] (9) **Host moving object description**, e.g. measurement vehicle details -> Could also be put in "Used data sources" (@ClemensLinnhoff You mentioned, this could be useful to you; Do you have an example use case for synthetic data?)
- [ ] (10) **Target moving objects:** OSI-id(s)+description of object(s) of interest (vehicle/pedestrian performing certain actions of interest)
- [ ] (11) **Events:** Tag certain events of interest with timestamp+description, e.g. cut-out action of vehicle x at timestamp y, sensor fault at timestamp z
- [ ] (12) **Used data sources/tools:** How/when was the trace generated/processed (repeated field to record tool/processing chain), e.g.
   - 2024-10-17 15:54:08+02:00: Captured on measurement vehicle x with sensor y (firmware version 1.0)
   - 2024-10-17 15:55:20+02:00: Processed with tool x version 1.0
   - 2024-10-17 15:54:08+02:00: Synthetically generated with tool x version 1.0/sensor model x version 1.0
- [ ] (13) Creator information: Contact/name/company/license

@TimmRuppert @ClemensLinnhoff @jdsika Feel free to add your opinion on which we should include.

In case we keep a lot of the fields above I would list the less important metadata definitions with less normative priority:
1) Required ("must"): osi_version, protobuf
2) Recommended ("should"): e.g. description, creator information, data sources, frame rate
3) Metadata hints: We suggest to use the proposed structure if the respective information is available and the trace creator wants to include it (e.g. location, granularity, contained road/lane types, events, traffic direction).

ClemensLinnhoff:
2) This is also something that mcap natively supports, as far as I know. @TimmRuppert how does it handle varying frame rates? Does it take the mean?
5) This is probably only valid for measurements which is not the standard case for OSI, so I would not put this in the standard specification.
9) My example use-case would be for re-simulating a measurement for model validation. Then you know, which kind of measurement vehicle is simulated.

TimmRuppert:
1. start and stop timestamps are present in the "summary" section and thus easily accessible (either via [API](https://mcap.dev/docs/cpp/rEFA639F56BD791D4) or [CLI](https://github.com/OpenSimulationInterface/open-simulation-interface/issues/828#issuecomment-2331698830)). Therefore, I would recommend to not store them in the metadata again.
2. The [average framerate](https://github.com/foxglove/mcap/blob/main/go/cli/mcap/cmd/info.go#L157) is given by the message-count and start/stop timestamps. This is for example supported by mcap [CLI](https://github.com/OpenSimulationInterface/open-simulation-interface/issues/828#issuecomment-2331698830) . Therefore, I would recommend to not store this in the metadata again.
3. I think this is somewhat covered by the shacls and hard to unify without further spec. I would agree with Clemens and leave that out for the moment
4. -9. Sure, why not
10. This might work for simple scenarios. But if we think about more complex city scenarios, deciding which traffic participant might be relevant is sometimes up to the function consuming this data. 
11. Strongly agree
12. We thought about something similar as well.  Instead of a repeated field (where the "field" would need to be defined by us as only string values are accepted) we could also do it per channel and add it to the channel metadata? 
13. Good idea, how about splitting this into "license" and "contact person" (or similar). Storing a license in the file might help sharing/reusing data.

In any case everything needs good key-name and category-name.

TimmRuppert:
General: 
- Can only contain one scenario with a unique global time 
- one mcap file is a dataset 


Metadata:
- Strongly recommended additional detailed metadata, category `asam_osi`
  - description e.g. cut-in 
  - creator e.g. person or company (not tool) csv
  - license csv of spdx identifiers 
  - data_sources e.g. csv of model, recorder, scenario player
- Channel-wise: `zero_time` :  ISO 8601  YYYYMMDDThhmmss.f representing the wall clock time 		
  
- Optional if you want to use more detailed metadata
  - Add a mcap wide `metadata_glossary` category (super-keys, name according to mcap) including prefixes. This prefix should be an identifier for metadata in channel. The value here in the general metadata should specify which spec this prefix represents  
  - `gaiax_hdmap` : `https://github.com/GAIA-X4PLC-AAD/ontology-management-base/tree/main/hdmap/`
  - Usage in channel `gaiax_hdmaps_HdMapShape` : `value 123`
  - Add another example for left hand / right traffic rule of openDrive
  - Add another example for some random iso

jdsika:
I would like to suggest to add a comment to "Nested Top Level Messages" as e.g. the SensorView in SensorData like the following:

When using ASAM OSI MCAP as a container for OSI traces the user is allowed to remove nested OSI Top level messages and add them as separate channels into the MCAP container"

This comment shoule be added above the top level messages in the .proto files as well.

What is your opinion about this? I think the nested messages were created as the collection of traces in one container was not defined at the time with .osi files?

TimmRuppert:
> What is your opinion about this? I think the nested messages were created as the collection of traces in one container was not defined at the time with .osi files?

Besides the points @ClemensLinnhoff mentioned, this might result in a good portion of looking back and forth in the file in order to the retrieve the corresponding messages. Especially as the `SensorData::timestamp` is not required to (but should) be the GroundTruth::timestamp. So there is no enforcement of a real identifier to match messages. Furthermore, repeated [SensorView](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/gen/structosi3_1_1SensorView.html) is repeated field in the SensorData. How would one specify which ones are meant?

Nonetheless, I understand and totally share your motivation. _Maybe_ something for OSI4. It would _just_ require a lot of breaking changes.

pmai:
> > What is your opinion about this? I think the nested messages were created as the collection of traces in one container was not defined at the time with .osi files?
> 
> I think the nesting is not only for trace files, but also for the messages between FMUs directly. I think it is just simpler to send one SensorData instead of a SensorData, a SensorView and a GroundTruth. So I would not add this to the proto files but just in the MCAP trace file documentation.

I don't think this is even at all related to trace files. I expect a trace file to contain the messages as they are being sent/received. No one is forcing anyone to use the nested messages, if they don't want to (I personally think they are usually a mistake in the case of SensorData->SensorView just for traceability purposes, but that's just me). But if they are used then they should be stored as is.

Now if someone wants to be creative they can do all kinds of things as they like. It's not the standards job to say how to use it. So I think this needs no mention anywhere, since it is purely up to the user and use case.

ClemensLinnhoff (approved):

doc/architecture/trace_file_mcap_format.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+
+|===
+
+
+**Example**
+
+The following list shows examples of valid OSI MCAP file names:
+
+- `20210818T150542Z_highway_sv.mcap`
```
thomassedlmayer, line 114: Why is there "highway" between the timestamp and the type specification?

doc/architecture/trace_file_mcap_format.adoc
```diff
@@ -0,0 +1,121 @@
...
+[#tab-MCAP-file-naming-convention]
+.MCAP file naming convention
+[cols="1,1"]
+|===
+|Field |Explanation 
+
+|opt. prefix
+|An optional prefix which may be used to specify the type of scenario (e.g. `cut-in`) or uniqueness of the setup (e.g. `target-5m`). May not contain any `_` characters.
+
+|opt. timestamp
```
thomassedlmayer, line 85: Maybe we should state that the timestamp can only exceptionally be omitted when there really is no reference to a global time in the file?

I think if you have a real-world capture or any other trace file that has any meaningful relation to a global time frame, it should be visible in the filename.

doc/architecture/trace_file_mcap_format.adoc
```diff
@@ -0,0 +1,121 @@
...
+== File-wide Metadata
+- Must include metadata with the name `versions` containing at least the following key-value pair:
+** `osi`: SemVer version of the minimum required OSI version
+- Must include metadata with the name `asam_osi` containing at least the following key-value pairs:
+** `zero_time`: ISO 8601 YYYYMMDDThhmmss.f formatted point in time representing the zero time of the scenario
+** `timestamp`: ISO 8601 YYYYMMDDThhmmss.f formatted creation time of the file 
+- It is strongly recommended to include metadata with the name `asam_osi` containing the following key-value pairs:
+** `description`: Short human-readable scenario description
+** `creator`: csv of person or company (not tool) creating the file
+** `license`` csv of spdx identifiers
```
thomassedlmayer, line 46: I think SPDX also has options to combine multiple licenses (expressions), so I don't know if csv is even necessary here.

doc/architecture/trace_file_mcap_format.adoc
```diff
@@ -0,0 +1,121 @@
...
+** Must reflect the timestamp of the stored OSI top-level message
+** Must be in nanoseconds
+
+
+== File-wide Metadata
+- Must include metadata with the name `versions` containing at least the following key-value pair:
+** `osi`: SemVer version of the minimum required OSI version
+- Must include metadata with the name `asam_osi` containing at least the following key-value pairs:
+** `zero_time`: ISO 8601 YYYYMMDDThhmmss.f formatted point in time representing the zero time of the scenario
+** `timestamp`: ISO 8601 YYYYMMDDThhmmss.f formatted creation time of the file
```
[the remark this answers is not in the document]
TimmRuppert: I would personally also favor the unix epoch in ms or nanoseconds. The proposal is using ISO 8601 as it seems like that was used for trace file names etc. before. 

That would also make more sense to be compliant with https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/gen/structosi3_1_1EnvironmentalConditions.html#a636bb78627046f34208f42f586ab2086?

Lets wait an see if anyone disagrees.

doc/architecture/trace_file_mcap_format.adoc
```diff
@@ -0,0 +1,121 @@
...
+== General Requirements
+- Must comply with the https://mcap.dev/spec[MCAP format specification] version `0x30`
+- Must allow other non-OSI data to be present in the MCAP file
+- Message records must be written into `chunk records` for indexed files
+- Only OSI top-level messages containing a timestamp field are permitted to be directly stored in MCAP channels
+- Must contain only a single scenario with a unique global time
+- An MCAP file is considered a single dataset
+
+== Schema
+- `name` field: Full message type name, including package (e.g., `osi3.SensorData`)
```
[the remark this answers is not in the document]
TimmRuppert: I understand the [protobuf documentation](https://protobuf.dev/reference/cpp/api-docs/google.protobuf.descriptor/#Descriptor.full_name.details) as if this is defined.  But to be more precise I will change it to  `fully-qualified name of the protobuf message type`

> Especially, because you used "OSI3::SensorData" in line 60 (not in the same context but anyways).

There I meant the [osi3::SensorData Struct](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/gen/structosi3_1_1SensorData.html). But it makes sense to simply write  "A channel containing OSI `SensorData` messages" and circumvent this. I will change line 60

doc/architecture/trace_file_mcap_format.adoc
```diff
@@ -0,0 +1,121 @@
...
+** Must include an `osi_version` key, specifying the OSI SemVer version of the OSI top-level message contained within the channel
+** Must include a `protobuf_version` key, specifying the protobuf SemVer version used to create the OSI top-level message contained within the channel
+** Should include a `description` key, explaining the data's origin and purpose in natural language.
+
+
+== Message
+- `publish_time` field: 
+** Must reflect the timestamp of the stored OSI top-level message
+** Must be in nanoseconds
+- `log_time` field: Must reflect the time when the message was enqueued for MCAP file addition
```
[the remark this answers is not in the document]
TimmRuppert: Can you please elaborate further?

doc/architecture/trace_file_mcap_format.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+== File-wide Metadata
+- Must include metadata with the name `versions` containing at least the following key-value pair:
+** `osi`: SemVer version of the minimum required OSI version
+- Must include metadata with the name `asam_osi` containing at least the following key-value pairs:
+** `zero_time`: ISO 8601 YYYYMMDDThhmmss.f formatted point in time representing the zero time of the scenario
+** `timestamp`: ISO 8601 YYYYMMDDThhmmss.f formatted creation time of the file 
+- It is strongly recommended to include metadata with the name `asam_osi` containing the following key-value pairs:
+** `description`: Short human-readable scenario description
+** `creator`: csv of person or company (not tool) creating the file
```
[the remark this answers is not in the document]
TimmRuppert: Yes I will write it out once.

doc/architecture/trace_file_mcap_format.adoc
```diff
@@ -0,0 +1,121 @@
...
+.MCAP file naming convention
+[cols="1,1"]
+|===
+|Field |Explanation 
+
+|opt. prefix
+|An optional prefix which may be used to specify the type of scenario (e.g. `cut-in`) or uniqueness of the setup (e.g. `target-5m`). May not contain any `_` characters.
+
+|opt. timestamp
+|Defines the absolute start time for a scenario or recording. If following the recommended zero time for the timestamps of the top-level messages, this time must represent the zero time. The format must adhere to ISO 8601 [cite:iso8601].
```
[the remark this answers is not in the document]
TimmRuppert: To my limited understanding of the ISO 8601, the .osi and .txth spec is not stating a format. I just provides the example `20210818T150542Z` and refers to ISO 8601 

Something like `YYYYMMDDThhmmssZ` would be a valid format ISO 8601 but  `YYYY-MM-DD-HH` as well right right? 

I would like to add the format  `YYYYMMDDThhmmssZ` and the mention that it must be in UTC (not global, due to the Z) to the .osi/.txth trace file naming convention but this assumes that generalizing the example is not considered a breaking change.

doc/architecture/trace_file_mcap_format.adoc
```diff
@@ -0,0 +1,121 @@
...
+- `publish_time` field: 
+** Must reflect the timestamp of the stored OSI top-level message
+** Must be in nanoseconds
+- `log_time` field: Must reflect the time when the message was enqueued for MCAP file addition
+** Must reflect the timestamp of the stored OSI top-level message
+** Must be in nanoseconds
+
+
+== File-wide Metadata
+- Must include metadata with the name `versions` containing at least the following key-value pair:
```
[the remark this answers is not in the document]
TimmRuppert: I do not have any opinion about this. @jdsika had the idea to add prefixes, so lets see what he suggests.

doc/architecture/trace_file_mcap_format.adoc
TimmRuppert:
```suggestion
- `20210818T150542Z_sv_highway.mcap`
```

doc/architecture/trace_file_mcap_format.adoc
```diff
@@ -0,0 +1,121 @@
...
+include::{root-path}_config.adoc[]
+endif::[]
+= MCAP Format
+
+== General Requirements
+- Must comply with the https://mcap.dev/spec[MCAP format specification] version `0x30`
+- Must allow other non-OSI data to be present in the MCAP file
+- Message records must be written into `chunk records` for indexed files
+- Only OSI top-level messages containing a timestamp field are permitted to be directly stored in MCAP channels
+- Must contain only a single scenario with a unique global time
```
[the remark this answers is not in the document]
thomassedlmayer: I remember that we talked about a common/unified time frame but not about that an MCAP file should only be allowed to contain a single scenario.

Also, I thought that the term "unique global time" is maybe a bit confusing as well. I'd rather write something like common time frame.

doc/architecture/trace_file_mcap_format.adoc
ClemensLinnhoff: In my opinion the meaningful relation to a global time is the exception here, since simulation should be the "normal" use-case rather than measurements.

doc/architecture/trace_file_mcap_format.adoc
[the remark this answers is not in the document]
thomassedlmayer: The naming convention even states that the recommended format is YYYMMDDThhmmssZ (even though with exemplary numbers). I would just add the format specification and maybe even consider decimal places. The recommended format remains the same as far as I'm concerned.

doc/architecture/trace_file_mcap_format.adoc
```diff
@@ -0,0 +1,121 @@
...
+** Must be in nanoseconds
+
+
+== File-wide Metadata
+- Must include metadata with the name `versions` containing at least the following key-value pair:
+** `osi`: SemVer version of the minimum required OSI version
+- Must include metadata with the name `asam_osi` containing at least the following key-value pairs:
+** `zero_time`: ISO 8601 YYYYMMDDThhmmss.fTZD formatted point in time representing the zero time of the scenario
+** `timestamp`: ISO 8601 YYYYMMDDThhmmss.fTZD formatted creation time of the file 
+- It is strongly recommended to include metadata with the name `asam_osi` containing the following key-value pairs:
```
thomassedlmayer, line 43: Should there be two "asam_osi" metadata records (see two lines above) or should all the metadata fields be in one metadata record?

doc/architecture/trace_file_mcap_format.adoc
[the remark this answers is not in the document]
TimmRuppert:
```suggestion
- Must contain only a single scenario with a common time frame
```

doc/architecture/trace_file_overview_file_formats.adoc
```diff
@@ -3,22 +3,25 @@ ifndef::include-only-once[]
...
+= Overview file formats
 
-There are two formats for storing multiple serialized OSI messages in one trace file.
+There are three formats for storing OSI messages in trace files:
 
 *.osi::
-Binary trace file.
-Messages are separated by a length specification before each message.
-The length is represented by a four-byte, little-endian, unsigned integer.
-The length does not include the integer itself.
```
thomassedlmayer, line 14: The information of this line got lost.

doc/architecture/trace_file_mcap_format.adoc
TimmRuppert: While I don't know if it is technically possible to have two metadata records with the same name (I assume it might be), it would make more sense to have all of that in one record. Considering your proposal to always speak of "a metadata record" in this context (see other comment) it should be clearer once this has been added.

doc/architecture/trace_file_mcap_format.adoc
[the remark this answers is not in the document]
thomassedlmayer: Still, I don't agree that an mcap file must contain only a single scenario. I think the word "scenario" could be misleading. You can put an arbitrary number of scenarios in one trace file. IMO the only relevant information here is that all the messages in the file must have a common time frame.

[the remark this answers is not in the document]
TimmRuppert:
```suggestion
- Must be limited to a single, unified sequence of events within the same time frame.
```

pmai:
I have taken the liberty to morph the current state into something that tries to be a more precise and normative specification while trying to be more minimalist in what it touches, and leaves other topics (like changes in the naming convention, the specific mapping of non-OSI meta-data into meta-data records) for either separate PRs or other layered specifications: #841.

It also tries to make the spec more robust, by specifying more explicitly how to use MCAP elements (e.g. placement of records, chunking, ...).

The other major change is that it now recommends making `log_time` and `published_time` identical, if no specific reasons would speak to making them differ, as this enables much better use of MCAP indexing facilities to do random access of traces.

People who want to replay while keeping jitter of their middleware (due to asynchronous communication) intact can still do so, but more sane use cases that abstract away middleware jitter (or are synchronous in nature) can still reap the benefits of the MCAP format index machinery (one might suggest to MCAP that they might like to add indexing on published_time to enable both use cases at the same time, but that's a different story).

PR #841, opened 2024-11-24
Initial multi channel trace file format specification

pmai:
Based on the initial specification work in #833 with refactoring for better integration into OSI and with clearer separation of concerns.

Update:
* [MCAP Specification](https://mcap.dev/spec)
* [MCAP Library](https://github.com/foxglove/mcap)
* [First implementation of this ASAM OSI MCAP specification](https://github.com/Lichtblick-Suite/asam-osi-utilities)
* [Using MCAP for OSI visualization](https://github.com/lichtblick-suite/asam-osi-converter)

### DIFF
--- a/doc/architecture/trace_file_formats.adoc
+++ b/doc/architecture/trace_file_formats.adoc
@@ -93,7 +93,7 @@ Values must match the following regular expression:
 NOTE: Even in pure simulation use cases there is usually a relationship to a simulated calendar time, as on-board components and environment simulation have necessary relationships to a simulated calendar time (for example, the embedded `HostVehicleData` will carry relevant simulated calendar time information, for localization and other purposes).
 
 `creation_time`::
-The point in real calendar time when the recording of the data in the trace file was started.
+The point in real calendar time when the recording of the data in the trace file was started, e.g. when the recording system began writing the data to a file.
 Must be provided as a full ISO8601 formatted date time string, including timezone data, conforming to the https://www.w3.org/TR/xmlschema11-2/#dateTimeStamp[XML Schema dateTimeStamp] lexical space.
 
 `description`::
--- a/doc/architecture/trace_file_formats.adoc
+++ b/doc/architecture/trace_file_formats.adoc
@@ -88,7 +88,7 @@ Must be provided as a full ISO8601 formatted date time string, including timezon
 Values must match the following regular expression:
 `-?([1-9][0-9]{3,}|0[0-9]{3})-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[01])T(([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9](\.[0-9]+)?|(24:00:00(\.0+)?))(Z|(\+|-)((0[0-9]|1[0-3]):[0-5][0-9]|14:00))`
 +
-NOTE: that even in pure simulation use cases there is usually a relationship to real time, as on-board components and environment simulation have necessary relationships to real time (for example, the embedded `HostVehicleData` will carry relevant real time information, for localisation and other purposes).
+NOTE: that even in pure simulation use cases there is usually a relationship to real time, as on-board components and environment simulation have necessary relationships to real time (for example, the embedded `HostVehicleData` will carry relevant real time information, for localization and other purposes).
 
 `creation_time`::
 The point in time when the trace file was created.
@@ -181,9 +181,6 @@ Each message record in a chunk record must contain the following fields:
 
 `channel_id`::
 The ID of the channel that the message belongs to.
-
-`publish_time`::
-The point in time when the message was published.
 
 `sequence`::
 Optional message counter to detect message gaps.
--- a/doc/architecture/trace_file_formats.adoc
+++ b/doc/architecture/trace_file_formats.adoc
@@ -40,6 +40,9 @@ The OSI multi channel trace file format is based on the MCAP file format, which 
 The OSI multi channel trace file format is a specialization of the MCAP file format, with additional constraints and requirements specific to OSI.
 Hence, any valid OSI multi channel trace file is also a valid MCAP file, but not the other way around.
 
+The OSI multi channel trace file format does not specify an MCAP profile.
+Therefore, it can be used with any or an empty profile field in the MCAP header record, as long as the given profile does not contradict or disallow any of the requirements of this specification.
+
 The following rules apply to OSI multi channel trace files:
 
 - The file extension to be used is `.mcap`.
--- a/doc/architecture/trace_file_formats.adoc
+++ b/doc/architecture/trace_file_formats.adoc
@@ -68,11 +68,11 @@ Rather, it indicates the version of the OSI trace file format itself.
 
 `min_osi_version`::
 The minimum version of the OSI schema used in the trace file OSI channels.
-This is a string in the format `major.minor.patch`, for example `3.0.0`.
+This is a string in the format `major.minor.patch`, for example `3.4.0`.
 
 `max_osi_version`::
 The maximum version of the OSI schema used in the trace file OSI channels.
-This is a string in the format `major.minor.patch`, for example `3.0.0`.
+This is a string in the format `major.minor.patch`, for example `3.7.0`.
 
 `min_protobuf_version`::
 The version of the protobuf implementation used in the trace file.
@@ -80,18 +80,18 @@ This is a string in the format `major.minor.patch`, for example `3.17.3`.
 
 `max_protobuf_version`::
 The version of the protobuf implementation used in the trace file.
-This is a string in the format `major.minor.patch`, for example `3.17.3`.
+This is a string in the format `major.minor.patch`, for example `3.26.0`.
 
 `zero_time`::
-The point in time corresponding to time 0 in all timestamps in the trace file.
+The point in real or simulated calendar time corresponding to time 0 in all timestamps in the trace file.
 Must be provided as a full ISO8601 formatted date time string, including timezone data, conforming to the https://www.w3.org/TR/xmlschema11-2/#dateTimeStamp[XML Schema dateTimeStamp] lexical space.
 Values must match the following regular expression:
 `-?([1-9][0-9]{3,}|0[0-9]{3})-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[01])T(([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9](\.[0-9]+)?|(24:00:00(\.0+)?))(Z|(\+|-)((0[0-9]|1[0-3]):[0-5][0-9]|14:00))`
 +
-NOTE: that even in pure simulation use cases there is usually a relationship to real time, as on-board components and environment simulation have necessary relationships to real time (for example, the embedded `HostVehicleData` will carry relevant real time information, for localization and other purposes).
+NOTE: Even in pure simulation use cases there is usually a relationship to a simulated calendar time, as on-board components and environment simulation have necessary relationships to a simulated calendar time (for example, the embedded `HostVehicleData` will carry relevant simulated calendar time information, for localization and other purposes).
 
 `creation_time`::
-The point in time when the trace file was created.
+The point in real calendar time when the recording of the data in the trace file was started.
 Must be provided as a full ISO8601 formatted date time string, including timezone data, conforming to the https://www.w3.org/TR/xmlschema11-2/#dateTimeStamp[XML Schema dateTimeStamp] lexical space.
 
 The `net.asam.osi.trace` meta-data record may also contain the following recommended `metadata` entries:
@@ -99,11 +99,8 @@ The `net.asam.osi.trace` meta-data record may also contain the following recomme
 `description`::
 A human-readable description of the data contained in the multi channel trace file.
 
-`creator`::
+`authors`::
 A list separated by commas of entities (not tools) involved in the creation of the data contained in the file.
-
-`license`::
-If the contents of the file is licensed under any SPDX registered license, this entry should contain the SPDX identifier for the license.
 
 `data_sources`::
 A list separated by commas of data sources used in the creation of the data contained in the file.
@@ -123,7 +120,8 @@ The schema record must contain the following fields:
 A file-wide unique non-zero identifier for the schema record.
 
 `name`::
-The fully qualified name to the message within the OSI descriptor set. For example, for the `SensorView` message type, this would be `osi3.SensorView`.
+The fully qualified name to the message within the OSI descriptor set.
+For example, for the `SensorView` message type, this would be `osi3.SensorView`.
 
 `encoding`::
 The value `protobuf`.
@@ -149,8 +147,8 @@ The ID of the schema record that describes the message type of the channel.
 
 `topic`::
 A unique name for the channel within the trace file.
-When recording OSI traces for a model packaged according to the OSI Sensor Model Packaging (OSMP) layer, using the naming conventions defined in the OSMP specification for variables as topics is recommended.
-For example, for a sensor model with two SensorView inputs and one SensorData output, the topic names would be `OSMPSensorViewIn[1]`, `OSMPSensorViewIn[2]`, and `OSMPSensorDataOut`, accordingly.
+When recording OSI traces for a model packaged according to the OSI Sensor Model Packaging (OSMP) layer, using the naming conventions defined in the OSMP specification for variables as topics is recommended, potentially with a suitable added prefix for disambiguation.
+For example, for a sensor model named RadarSensorFL with two SensorView inputs and one SensorData output, the topic names would be `RadarSensorFL.OSMPSensorViewIn[1]`, `RadarSensorFL.OSMPSensorViewIn[2]`, and `RadarSensorFL.OSMPSensorDataOut`, accordingly.
 In other cases, the topic name should be chosen to reflect the purpose of the channel, and should include some indication of the message type.
 
 `message_encoding`::
@@ -192,8 +190,8 @@ Otherwise this should be set to zero to indicate that no reliable sequence numbe
 `log_time`::
 This field is in nanoseconds and uses the same epoch as the `publish_time` field.
 It is used to determine the order of messages in the trace file, and provides for time-based random access to the data streams.
-Unless there is a specific reason to set this field to a different value, it should be set to the same value as `publish_time`, as this reflects the time flow as reflected in the OSI message stream.
-Only if recreation of the message stream with actual message transmission times is required - for example for asynchronous packaging layers, this field should be set to the time when the message was enqueued for addition to the trace file.
+Unless there is a specific reason to set this field to a different value, it should be set to the same value as `publish_time`, as this reflects the time flow of the OSI message stream.
+Only if recreation of the message stream with actual message transmission times is required, for example for asynchronous packaging layers, this field should be set to the simulation time when the message was enqueued for addition to the trace file.
 
 `publish_time`::
 The timestamp taken from the timestamp field of the stored OSI top-level message.
--- a/doc/architecture/trace_file_formats.adoc
+++ b/doc/architecture/trace_file_formats.adoc
@@ -5,7 +5,10 @@ endif::[]
 [#top-osi_trace_file_formats]
 = OSI trace file formats
 
-There are two formats for storing multiple serialized OSI messages in one trace file.
+== Single channel trace file formats
+There are two formats for storing and reading multiple serialized OSI messages of the same type in one trace file.
+These formats are very bare bones and do not contain any additional information, like meta-data, schemas, nor do they allow for random access.
+For more advanced use-cases, consider using the multi channel trace file format.
 
 *.osi::
 Binary trace file.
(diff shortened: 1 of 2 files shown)


pmai:
Note that this leaves out for now specifics of how to structure additional non-OSI meta-data, which should likely be handled in layered specifications based on the rdns prefixes, and any larger changes to the naming convention. The later is only a recommendation in any case, and if further changes are wanted they should apply to the whole convention and be handled in a separate PR (e.g. one could think about making the timestamp optional, or dropping the protobuf version, but that would make sense - or not - for all formats and hence is not related to the multi trace file format).

doc/architecture/trace_file_formats.adoc
```diff
@@ -16,9 +19,187 @@ The length does not include the integer itself.
...
+The version of the protobuf implementation used in the trace file.
+This is a string in the format `major.minor.patch`, for example `3.17.3`.
+
+`zero_time`::
+The point in time corresponding to time 0 in all timestamps in the trace file.
+Must be provided as a full ISO8601 formatted date time string, including timezone data, conforming to the https://www.w3.org/TR/xmlschema11-2/#dateTimeStamp[XML Schema dateTimeStamp] lexical space.
+Values must match the following regular expression:
+`-?([1-9][0-9]{3,}|0[0-9]{3})-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[01])T(([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9](\.[0-9]+)?|(24:00:00(\.0+)?))(Z|(\+|-)((0[0-9]|1[0-3]):[0-5][0-9]|14:00))`
++
+NOTE: that even in pure simulation use cases there is usually a relationship to real time, as on-board components and environment simulation have necessary relationships to real time (for example, the embedded `HostVehicleData` will carry relevant real time information, for localisation and other purposes).
```
ClemensLinnhoff, line 91:
```suggestion
NOTE: that even in pure simulation use cases there is usually a relationship to real time, as on-board components and environment simulation have necessary relationships to real time (for example, the embedded `HostVehicleData` will carry relevant real time information, for localization and other purposes).
```

doc/architecture/trace_file_formats.adoc
```diff
@@ -16,9 +19,187 @@ The length does not include the integer itself.
...
+
+`creation_time`::
+The point in time when the trace file was created.
```
ClemensLinnhoff, line 94: Is this the beginning or end of writing the trace file?

doc/architecture/trace_file_formats.adoc
```diff
@@ -16,9 +19,187 @@ The length does not include the integer itself.
...
+
+`min_protobuf_version`::
+The version of the protobuf implementation used in the trace file.
+This is a string in the format `major.minor.patch`, for example `3.17.3`.
+
+`max_protobuf_version`::
+The version of the protobuf implementation used in the trace file.
+This is a string in the format `major.minor.patch`, for example `3.17.3`.
+
+`zero_time`::
```
ClemensLinnhoff, line 85: I would put this in the optional section, as for many simulation use-cases I don't see a mandatory relation to real-time.

doc/architecture/trace_file_formats.adoc
```diff
@@ -16,9 +19,187 @@ The length does not include the integer itself.
...
+
+All messages in an OSI channel must be stored in chunk records in the data section of the trace file.
+
+Each message record in a chunk record must contain the following fields:
+
+`channel_id`::
+The ID of the channel that the message belongs to.
+
+`publish_time`::
+The point in time when the message was published.
```
ClemensLinnhoff, line 186: In #833 we defined the publish_time as:

```
Must reflect the timestamp of the stored OSI top-level message
```

This makes it easier to use the indexing to access certain OSI time frames.

doc/architecture/trace_file_naming.adoc
```diff
@@ -76,3 +82,9 @@ The recommended file name is:
 ----
 20210818T150542Z_sv_312_300_1523_highway.osi
 ----
+
+For a corresponding multi channel trace file containing `SensorView` and `GroundTruth` messages, the recommended file name is:
+
+----
+20210818T150542Z_multi_312_300_1523_highway.mcap
+----
```
ClemensLinnhoff, line 90: I know, you would like to move the naming convention discussion to a separate issue. But I am still very much against using a timestamp in a simulation data file name. I also know, it is "just" a recommendation. But I would still not even recommend it.

ClemensLinnhoff (dismissed):

doc/architecture/trace_file_formats.adoc
pmai: I don't know, this is taken from Timm's definitions. Personally I would actually not care very much either way, as without additional recording related meta-data (when, how why, with which equipment was this recorded) I don't see very much use for this field for real measurements, and for simulation use the file system timestamps are usually sufficient.

doc/architecture/trace_file_formats.adoc
pmai: Given that OSI messages contain (and system components use) fields that reference simulated real time all over the place, there is usually always a relationship to some simulated real time, even in pure simulation use cases. Similarly for anything GPS related you want to have valid timestamps, and to even map GPS time to UTC you need the date.

So I would venture that only very rare cases will get away with not having any time relationship (and would have to be validated to ensure there is no missed effects), and hence this should be recorded here. I.e. for simulations that have no particular focus on time effects, usually a convenient fixed time is chosen to correllate with time zero.

In any case this being part of the mandatory meta-data reflects Timm's PR, where it is mandatory as well.

doc/architecture/trace_file_formats.adoc
pmai: > In #833 we defined the publish_time as:

Actually this was a copy and paste left over, see actual publish_time definition below (will remove this old duplicate).
> 
> ```
> Must reflect the timestamp of the stored OSI top-level message
> ```
> 
> This makes it easier to use the indexing to access certain OSI time frames.

It sadly does not, as solely `log_time` is used for indexing purposes in MCAP, see contortions below with `log_time` to deal with that.

pmai, line 186:
```suggestion
```

doc/architecture/trace_file_naming.adoc
pmai: See discussion above on `zero_time`: I don't think that the missing relationship to some real-time is as common as you assume; that said I can certainly see making the timestamp optional for certain use cases could be a thing; but as you say, I think this needs to be handled in a seprate PR anyways.

doc/architecture/trace_file_formats.adoc
ClemensLinnhoff: Yes, but if we want to put the same values in publish_time and log_time, I propose to make them both identical with the timestamp of the message.

pmai: That's what the current definition does, with special allowance to still use log_time as intended if there is a - perceived - need.

doc/architecture/trace_file_naming.adoc
ClemensLinnhoff: Thats odd. I regularly see trace files from OEMs, Tier1s, tool vendors and research institutes without the timestamp.

doc/architecture/trace_file_formats.adoc
ClemensLinnhoff: see below.

TimmRuppert (changes requested):

doc/architecture/trace_file_formats.adoc
ClemensLinnhoff: Then I propose to put the end time in there, when the mcap file was finalized.

doc/architecture/trace_file_formats.adoc
```diff
@@ -16,9 +19,187 @@ The length does not include the integer itself.
 *.txth::
 Human-readable plain-text trace file.
 Messages are separated by newlines.
+Each message is a serialized OSI message in JSON format.
```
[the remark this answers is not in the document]
pmai: Good catch, that was a think-o on my part:

```suggestion
Each message is a serialized OSI message in protocol buffer text format.
```

doc/architecture/trace_file_naming.adoc
```diff
@@ -44,6 +44,12 @@ Trace file contains `MotionRequest` messages.
 `su`::
 Trace file contains `StreamingUpdate` messages.
 
+`multi`::
+Trace file contains multiple types of messages for use with multi channel trace file format.
+In this case the number-of-frames field should be the largest number of frames across all channels.
```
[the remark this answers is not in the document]
pmai: Because this is the most minimal change to the existing naming convention while still sticking to its spirit, and at least still halfway useful for e.g. gauging approximate content. Dito for the next two fields.

If we want to change the naming convention, then a separate PR is the way to go - personally I would love to get rid of all the protobuf versions, but a naming convention is only useful if it does not change all of the time or has twenty different variants. Especially if it is already optional, so not actually restricting anyone.

doc/architecture/trace_file_formats.adoc
```diff
@@ -5,7 +5,10 @@ endif::[]
...
 = OSI trace file formats
 
-There are two formats for storing multiple serialized OSI messages in one trace file.
+== Single channel trace file formats
+There are two formats for storing and reading multiple serialized OSI messages of the same type in one trace file.
+These formats are very bare bones and do not contain any additional information, like meta-data, schemas, nor do they allow for random access.
+For more advanced use-cases, consider using the multi channel trace file format.
 
 *.osi::
 Binary trace file.
```
pmai, line 14:
```suggestion
Single channel binary trace file.
```

doc/architecture/trace_file_formats.adoc
```diff
@@ -16,9 +19,184 @@ The length does not include the integer itself.
 *.txth::
 Human-readable plain-text trace file.
```
pmai, line 20:
```suggestion
Single channel human-readable plain-text trace file.
```

doc/architecture/trace_file_formats.adoc
pmai: Finalized time stamp would however mean that the meta-data record can only be written at the end of recording (at the end of the data section, as metadata records are not allowed in the summary section at the end). Not impossible to do, but has this effect - start of recording would mean the meta-data record can be written at the beginning (not that this matters all that much, just to keep in mind in discussions).

doc/architecture/trace_file_formats.adoc
```diff
@@ -16,9 +19,184 @@ The length does not include the integer itself.
...
+
+`data`::
+A binary FileDescriptorSet as produced by `protoc --include_imports --descriptor_set_out`.
+
+The schema record must be stored in the summary section of the trace file, and must be referenced by the OSI channels that use the schema.
+
+[#sec-osi-trace-file-multi-channel]
+=== OSI channel
+
+An OSI channel is a data stream within the OSI multi channel trace file that contains serialized OSI messages of the same type.
```
pmai, line 139:
```suggestion
An OSI channel is a data stream within the OSI multi channel trace file that contains serialized OSI top-level messages of the same type.
Note that non-top-level messages must not be stored directly in OSI channels.
```

doc/architecture/trace_file_naming.adoc
pmai: I regularly see trace files with wrong moving object coordinate usage, missing mandatory fields, etc., too. Does not mean this is good or correct usage. And if it works by accident, it still works.

doc/architecture/trace_file_formats.adoc
TimmRuppert: Other tools e.g. ros also use the start of the recording. Nonetheless, I would not name it recording as OSI is mainly focusing on simulations (and also enabling real measurements) where it is simply "writing" but not "recording".

ClemensLinnhoff: I don't have a strong opinion on this. We just need to specify it. We could also put the start of writing to the file as creation_time.

doc/architecture/trace_file_naming.adoc
ClemensLinnhoff: Not using something or using something in the wrong way are completely different things.

doc/architecture/trace_file_formats.adoc
```diff
@@ -5,7 +5,10 @@ endif::[]
 [#top-osi_trace_file_formats]
```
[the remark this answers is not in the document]
TimmRuppert: Now that there is an issue for that we can resolve this converstation to reduce the effect in one PR  https://github.com/OpenSimulationInterface/open-simulation-interface/issues/842

doc/architecture/trace_file_naming.adoc
```diff
@@ -9,7 +9,7 @@ endif::[]
 The names of OSI trace files should have the following format:
 
 ----
-<timestamp>_<type>_<osi-version>_<protobuf-version>_<number-of-frames>_<custom-trace-name>.osi
+<timestamp>_<type>_<osi-version>_<protobuf-version>_<number-of-frames>_<custom-trace-name>.[osi|txth|mcap]
```
ClemensLinnhoff, line 12: If we are moving the naming convention discussion to #843 this change should be reverted.

pmai (commented):
Review comments from discussion with @jdsika and @thomassedlmayer

ClemensLinnhoff:
I don't think any further review from @TimmRuppert or me is necessary here. Feel free to merge your solution whenever you are ready.

jdsika:
@pmai is it possible to commit the discussed changes until 1pm and I have a final review and put the PR into ready state with the tag "readyForCCB"?

doc/architecture/trace_file_formats.adoc
```diff
@@ -5,20 +5,200 @@ endif::[]
...
+`min_protobuf_version`::
+The version of the protobuf implementation used in the trace file.
+This is a string in the format `major.minor.patch`, for example `3.17.3`.
+
+`max_protobuf_version`::
+The version of the protobuf implementation used in the trace file.
+This is a string in the format `major.minor.patch`, for example `3.17.3`.
+
+`zero_time`::
+The point in time corresponding to time 0 in all timestamps in the trace file.
```
pmai, line 86:
```suggestion
The point in real or simulated calendar time corresponding to time 0 in all timestamps in the trace file.
```

doc/architecture/trace_file_naming.adoc
pmai: Moves to #843.

doc/architecture/trace_file_naming.adoc
```diff
@@ -44,6 +44,12 @@ Trace file contains `MotionRequest` messages.
 `su`::
 Trace file contains `StreamingUpdate` messages.
 
+`multi`::
+Trace file contains multiple types of messages for use with multi channel trace file format.
+In this case the number-of-frames field should be the largest number of frames across all channels.
+The OSI version field should be based on the `version` field in the file meta-data of the multi-channel trace file.
```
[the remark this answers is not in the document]
pmai: Moves to #843.

doc/architecture/trace_file_formats.adoc
```diff
@@ -5,20 +5,198 @@ endif::[]
...
+Due to the nature of the format, it allows for random access to the data streams, and is suitable for more advanced use-cases.
+
+The OSI multi channel trace file format is based on the MCAP file format, which is a generic multi channel trace file format.
+The OSI multi channel trace file format is a specialization of the MCAP file format, with additional constraints and requirements specific to OSI.
+Hence, any valid OSI multi channel trace file is also a valid MCAP file, but not the other way around.
+
+The following rules apply to OSI multi channel trace files:
+
+- The file extension to be used is `.mcap`.
+- The file must be a valid MCAP file according to the https://mcap.dev/spec[MCAP format specification] version `0x30`.
```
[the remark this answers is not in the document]
pmai: The spec goes about it in a very round-about way:

> **Magic**
> An MCAP file must begin and end with the following [magic bytes](https://en.wikipedia.org/wiki/File_format#Magic_number):

> 
> `0x89, M, C, A, P, 0x30, \r, \n`

> 
> The byte following "MCAP" is the major version byte. **0x30 is the ASCII character 0.** **Any changes to this specification document (i.e. adding fields to records, introducing new records) will be binary backward-compatible within the major version.**

(Emphasis by me)

So it is stated quite unclearly, but I would treat this as the major version of the specification, given the statements.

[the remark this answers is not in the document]
jdsika: That is what confuse me. "include the major version like this" but not "this spec IS major version 0" ...

doc/architecture/trace_file_formats.adoc
```diff
@@ -5,20 +5,198 @@ endif::[]
...
+
+`min_protobuf_version`::
+The version of the protobuf implementation used in the trace file.
+This is a string in the format `major.minor.patch`, for example `3.17.3`.
+
+`max_protobuf_version`::
+The version of the protobuf implementation used in the trace file.
+This is a string in the format `major.minor.patch`, for example `3.26.0`.
+
+`zero_time`::
```
jdsika, line 85: @pmai I asked around and we have to make this optional/recommended

Argumentation:
1) Naming convention contains it currently and IS optional
2) OSI optional TimeOfDay time_of_day = 2; and optional int64 unix_timestamp = 8; are optional in environment.proto

doc/architecture/trace_file_formats.adoc
```diff
@@ -5,20 +5,198 @@ endif::[]
...
+- The file must contain at least one OSI message stream in a channel as defined in section <<sec-osi-trace-file-multi-channel>>.
+- The file may contain additional non-OSI message streams in other channels.
+
+NOTE: All OSI and non-OSI message streams stored in the same file share a common time base.
+Storing of unrelated message streams in one trace file is therefore not generally useful and should be avoided.
+
+[#sec-osi-trace-file-multi-global-meta-data]
+=== Multi channel trace file global meta-data
+
+The file must contain a meta-data record with the `name` field being `net.asam.osi.trace` and the following mandatory `metadata` entries:
```
TimmRuppert, line 61:
```suggestion
The file must contain a meta-data record with the `name` field being `net.asam.osi.trace`. This name must be unique, and no other meta-data record with the same name is permitted. The record must also include the following mandatory metadata entries:
```

One can easily add metadata with the same name twice. 


```bash
mcap list metadata /tmp/sv_example.mcap 
name              	offset	length	metadata                                                                                                                                 	
net.asam.osi.trace	46    	181   	{"max_osi_version":"3.7.0","max_protobuf_version":"3.21.12","min_osi_version":"3.7.0","min_protobuf_version":"3.21.12","version":"3.7.0"}	
net.asam.osi.trace	227   	122   	{"description":"Example mcap trace file created with the ASAM OSI utilities library."}   

```

jdsika: That is the reason why the specification explicitly forbids it :)

pmai:
```suggestion
The file must contain exactly one meta-data record with the `name` field being `net.asam.osi.trace` and the following mandatory `metadata` entries:
```

jdsika (approved):

ijelec (commented):
There are just two points on naming convention and time definitions that were discussed during our meeting. No other objections to merging.

doc/architecture/trace_file_naming.adoc
```diff
@@ -76,3 +82,9 @@ The recommended file name is:
 ----
 20210818T150542Z_sv_312_300_1523_highway.osi
 ----
+
+For a corresponding multi channel trace file containing `SensorView` and `GroundTruth` messages, the recommended file name is:
+
+----
+20210818T150542Z_multi_312_300_1523_highway.mcap
```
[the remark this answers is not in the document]
pmai: Agreed.

TimmRuppert:
> 1. The OSI multi channel trace file format is a binary file format that allows for storing multiple serialized OSI message streams of the same or different types in one trace file, along with additional meta-data, and other related data streams. - what is  "other related data streams" how do we know whether it is related or not.

Ultimately, I think it's up to the user to decide. For example, a sensor model that also outputs some non-OSI debug data could still be considered a related data stream. On the other hand, a list of my latest cookie recipes probably wouldn't be ;) 

> 2. For each OSI top-level message type that is used in one or more OSI channels, the OSI multi channel trace file must contain a corresponding schema record in the summary section. - What happens to non-top-level message type?

The answer is basically question 3.

> 3. An OSI channel is a data stream within the OSI multi channel trace file that contains serialized OSI top-level messages of the same type. Note that non-top-level messages must not be stored directly in OSI channels. - How are non top level messages stored?

Well, you don’t. It seems the documentation might leave room for misinterpretation. Perhaps it should be made more explicit - for example, by stating clearly at the beginning of the section 'The following rules apply to OSI multi-channel trace files' that non-top-level messages may only be stored as sub-elements within the channel of a top-level message.

TimmRuppert (approved):

jdsika:
TODO:
Be clearer about what the "profile" string should and should not be

doc/architecture/trace_file_formats.adoc
```diff
@@ -5,20 +5,199 @@ endif::[]
...
+
+=== Overview
+
+The OSI multi channel trace file format is a binary file format that allows for storing multiple serialized OSI message streams of the same or different types in one trace file, along with additional meta-data, and other related data streams.
+Due to the nature of the format, it allows for random access to the data streams, and is suitable for more advanced use-cases.
+
+The OSI multi channel trace file format is based on the MCAP file format, which is a generic multi channel trace file format.
+The OSI multi channel trace file format is a specialization of the MCAP file format, with additional constraints and requirements specific to OSI.
+Hence, any valid OSI multi channel trace file is also a valid MCAP file, but not the other way around.
+
```
thomassedlmayer, line 42:
```suggestion

The OSI multi channel trace file format does not specify an MCAP profile.
Therefore, the use of the profile record in the MCAP header section is not prescribed.

```

pmai: Maybe more to the point:

```suggestion
The OSI multi channel trace file format does not specify an MCAP profile.
Therefore, it can be used with any or no profile record in the MCAP header section, as long as the given profile does not contradict or disallow any of the requirements of this specification.
```

jdsika (approved):

pmai:
CCB 2025-07-03: Decision for PR #841: Review until the 2025-07-11, when the CCB meets again at 09:00-11:00 CEST for further discussion and potential decision on merge.

doc/architecture/trace_file_formats.adoc
```diff
@@ -5,20 +5,202 @@ endif::[]
...
+The OSI multi channel trace file format does not specify an MCAP profile.
+Therefore, it can be used with any or an empty profile field in the MCAP header record, as long as the given profile does not contradict or disallow any of the requirements of this specification.
+
+The following rules apply to OSI multi channel trace files:
+
+- The file extension to be used is `.mcap`.
+- The file must be a valid MCAP file according to the https://mcap.dev/spec[MCAP format specification] version `0x30`.
+- The file must be an indexed MCAP file, including chunk index records in the summary section, with all message records written into chunk records.
```
beatschrauber, line 50: An editorial note: I think it should be 'shall' instead of 'must'. This applies to most, if not all, uses of the word here.

AsamDiegoSanchez:
CCB 2025-07-11: Decision for PR https://github.com/OpenSimulationInterface/open-simulation-interface/pull/841: All agreed to merge it with the remark that new PRs will be need to solve the bibliography references

PR #852, opened 2025-02-05
Fix open_simulation_interface-config.cmake.in

doganulus:
The corresponding filename in the patch must be the same as [this filename](https://github.com/OpenSimulationInterface/open-simulation-interface/blob/e78dff7443bb71e563431175fbd09bafba27da2c/CMakeLists.txt#L230)

When using CMake's `find_package` to use the library in another project, the mismatch between filenames results in an error.


pmai (approved):
Good catch, should be applied.

yash-shah-asam:
ASAM OSI CCB 10 April 2025: Approved and assigned to v3.7.1

PR #863, opened 2025-04-11
Adapt vehicle types from openx harmonization

thomassedlmayer:
This PR aims to adapt the vehicle type definitions and descriptions that are defined by the [OpenX Traffic Participants Specification](https://simulation.pages.asam.net/-/harmonization/trafficparticipants/antora-generator-traffic-participants/-/jobs/65424/artifacts/public/ASAM_TrafficParticipants_Specification/current/specification/07_traffic_participants/07_02_vehicles.html).

I also replaced the old meaningless ("Vehicle is a car") descriptions with the ones of the OpenX specification.

@yash-shah-asam Could you check if there is a permalink to the OpenX specification, I could use? I think it would be reasonable to include a permalink to the specification and somehow signify that OSI tries to adhere to these definitions.

I think there a still some issues open for discussion:
1. OSI currently defines `TYPE_DELIVERY_VAN` which should probably be deprecated (as it contains a role). I added the preferred alias `TYPE_VAN`.
2. OSI currently defines bicycles to not include a motor ("Vehicle is a bicycle (without motor and specific lights)") whereas the OpenX specification includes motor-assisted bicycles.
3. `TYPE_MOTORBIKE` contradicts the naming in the OpenX specification ("Motorcycle"). I added an alias. I'm not sure if we should deprecate `TYPE_MOTORBIKE` or just keep both.
4. I added `TYPE_MICROMOBILITY_DEVICE` which includes stand-up scooters but OSI already defines the type `TYPE_STANDUP_SCOOTER`. Should it be deprecated? It seems like an issue that in the description of micro-mobility devices the stand-up scooter is even mentioned as an example.

Apart from these issues, the changes are relatively straightforward and don't contradict the current OSI definitions from my point of view.

@jakobkaths


osi_object.proto
```diff
@@ -835,67 +837,143 @@ message MovingObject
...
             //
+            // \note Deprecated differentiation, use TYPE_VAN instead
+            //
             TYPE_DELIVERY_VAN = 6;
 
-            // Vehicle is a (heavy) truck.
+            // A van is a motorized vehicle with a larger cargo area than a car,
+            // used for transporting goods or people.
+            //
+            TYPE_VAN = 6;
```
jdsika, line 847: Using the same index twice works as an alias, right?

thomassedlmayer: Yes, but it requires `option allow_alias = true;` to be set. This was already added in 3.7.0 for the vehicle types because the alias TYPE_CAR (index 4) was added.

jakobkaths:
Thanks for the initiative and quick implementation, @thomassedlmayer ! I would agree to deprecate DELIVERY_VAN and MOTORBIKE. Regarding the BICYCLE and what they do / don't include let's dicuss in the group again. Same for the scooter...

osi_object.proto
```diff
@@ -835,67 +837,143 @@ message MovingObject
...
+            // right-of-way.
+            // Therefore, in case crossings with the road infrastructure occur,
+            // the exclusive right-of-way is ensured, e.g. by railway barriers.
+            // Often as a series of connected vehicles.
             //
             TYPE_TRAIN = 14;
 
             // Vehicle is a wheelchair.
             //
             TYPE_WHEELCHAIR = 15;
```
jakobkaths, line 938: The Wheelchair was sorted into Micromobility Device as well in the OpenX definition. We have to make sure that we don't keep surplus definitions here, either. But again, we could discuss in the group if this is maybe prominent enough to spearate it into a dedicated type.

osi_object.proto
```diff
@@ -835,67 +837,143 @@ message MovingObject
...
+            // \note The vehicle can be, but doesn't need to be, attached to
+            // another vehicle.
             //
             // \image html OSI_TYPE_TRAILER.svg
             //
             TYPE_TRAILER = 9;
 
             // Vehicle is a motorbike or moped.
             //
+            // \note Deprecated differentiation, use TYPE_MOTORCYCLE instead
```
jakobkaths, line 890:
```suggestion
            // \note Deprecated name, use TYPE_MOTORCYCLE instead
```

osi_object.proto
thomassedlmayer: The discussion result was to separate out both wheelchair and stand-up scooter types from micro-mobility device in the harmonized traffic participation specification and specifically state this in the description.

I adopted the changed descriptions from [here](https://code.asam.net/simulation/harmonization/trafficparticipants/traffic-participants-specification/-/commit/624d53287bdee12a6eb24e2ec5456524def293b3).

thomassedlmayer:
@jakobkaths I synced the [wording/grammar changes](https://code.asam.net/simulation/harmonization/trafficparticipants/traffic-participants-specification/-/commit/0261f28b550be432ce84523c09ff6c103169a59b) from the traffic participant specification and put the label ReadyForCCB so this can be reviewed this in the CCB tomorrow morning. Could you also have a look again?

jakobkaths (approved):

jdsika (approved):

pmai (approved):

PR #865, opened 2025-06-11
689 add raytracerview config - RAYTRACER_FORMAT

SandroReith:
Issue #699

#### Add a description
Reformatted to utilize the ray tracing approach
Offers a generic description, independent of tool provider or sensor technology

**Some questions to ask**:
What is this change?
Format of the Raytracing View.
Content of the Raytracing View
Name of the enum

What does it fix?
Reduces dependency on specific tools and technologies
Facilitates standardization and unification

How has it been tested?
proposal

### DIFF
--- a/osi_sensorviewconfiguration.proto
+++ b/osi_sensorviewconfiguration.proto
@@ -1088,33 +1088,19 @@ message RaytracerViewConfiguration
         // Consider proposing an additional format if using
         // \c #RAYTRACER_FORMAT_OTHER.
         //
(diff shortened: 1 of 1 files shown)


thomassedlmayer (commented):
@SandroReith: Maybe we could already merge this PR into the other PR (#699) and continue working there?

Another topic: Since this is about the unambiguous specification of a binary format, isn't it necessary to also specify the float size/precision. endianness and vector/matrix value order?

PR #872, opened 2025-08-07
Install OSI3 under conventional install locations

doganulus:
The usual practice is to install libraries directly under `${INSTALL_PREFIX}/lib`.

This PR is a small step toward discussing the adoption of CMake’s [`GNUInstallDirs`](https://cmake.org/cmake/help/v3.25/module/GNUInstallDirs.html) conventions for standardized install locations.

If this approach sounds reasonable, I’d prefer the OSI fully embrace `GNUInstallDirs` conventions instead of using custom install location variables.


jdsika:
This PR will be integrated in one of the other PRs related to the https://github.com/OpenSimulationInterface/open-simulation-interface/pull/870 - thank you for this very constructive proposal!

PR #877, opened 2025-09-25
Add new allowable rules minimum/maximum_length

pmai:
These checks are now supported by qc-osi-trace checker, so allow them in the CI tests.


AsamDiegoSanchez:
OSI CCB meeting 25.09.2025: group has approved the PR.

Issue #874, opened 2025-08-08
References may not be consistent


arauschert:
#### Describe the bug
There are two concepts of references:
1. model_reference (reference to 3d model asset)
2. ExternalReference (which can reference also a 3d asset)

This may confuse the user.

For example in the RoadMarking element, there is no model_reference attribute, while for all other elements (traffic light, traffic sign, ground truth, moving object, stationary object) there are model_references.

A second example would be that the BoundingBox element also has a model_reference attribute, which duplicates the model_reference attribute of the parent elements (traffic light/sign, moving object...).

Furthermore, the road marking element of type stop line, where there is no direct link to the corresponding traffic sign / traffic light.

#### Describe the expected behavior
 The question would be in references can be defined in a consistent way.

#### Show some screenshots
<img width="504" height="976" alt="Image" src="https://github.com/user-attachments/assets/d2e49163-84cc-46b1-a873-404474fdcd59" />

arauschert:
Hi @pmai ! Thanks for your answers.
The part that confused me about referencing 3d models in the ExternalReference is here:
<img width="1121" height="328" alt="Image" src="https://github.com/user-attachments/assets/d45cddd7-966c-4f1a-a541-f9c0c465f919" />
If we use GroundTruth::model_reference we can leave this field empty. Does that mean that the ExternalReference refers to a 3d model asset rather than an OpenDRIVE file?

>I think a good case could be made to add model_reference to RoadMarking, however it would have to be clarified what kind of model is expected there for which cases, as there are multiple types of road markings, and the role of the model might differ (or not - someone working out a proposal should look at this).

Yes, hat would be great because we're missing the model_reference attribute in the RoadMarking element and would need it. How should we proceed with this? To me it would be quite straightforward to add it. Which different types of road markings do you mean? They are all painted on the road surface.

About the stop line: sorry, I was a bit unclear. Yes, I mean a road marking element with TYPE_SYMBOLIC_TRAFFIC_SIGN. How can this be semantically linked to a traffic light?

pmai:
[...]

PR #886, opened 2025-10-31
Add model_reference attribute to RoadMarking (#874)

arauschert:
https://github.com/OpenSimulationInterface/open-simulation-interface/issues/874

#### Add a description
Adding the model_reference attribute to the RoadMarking class. 
This was missing and is now consistent with the other classes TrafficLight, MovingObject...

Backward-compatible, as added as optional and at the end (attribute #6).

- [x] I have done the [DCO signoff](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/specification/contributing/dco.html).

### DIFF
--- a/osi_roadmarking.proto
+++ b/osi_roadmarking.proto
@@ -101,6 +101,11 @@ message RoadMarking
     // \note It is implementation-specific how model_references are resolved to
     // 3d models.
     //
+    // \note Road markings' base objects are positioned differently from other
+    // OSI objects: their local z-axis corresponds to how the marking is placed
+    // on the xy-plane of the lane, rather than pointing upwards as it does for
+    // other objects.
+    //
     optional string model_reference = 6;
 
     //


thomassedlmayer (approved):
Looks good to me. I don’t think there are any special cases for specific road marking types as discussed in #874. All road marking types must conform to the base dimensions regardless of their type. So the model reference role should be consistently implementable for all types.

My only remark for road marking model placement is that road markings' base objects are positioned differently from other OSI objects: their local z-axis corresponds to how the marking is placed on the xy-plane of the lane ([see OSI documentation for more details](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/V3.7.0/gen/structosi3_1_1RoadMarking.html#a8b19d63b7b7ac893eadf0c3b66a1746b)), rather than pointing upwards as it does for other objects. But as the model reference use is left to the implementer, I guess this just plays a role when you're trying to be consistent (in one way or another) in your model reference implementation.

osi_roadmarking.proto
```diff
@@ -96,6 +96,18 @@ message RoadMarking
...
+    // Opaque reference of an associated 3D model of the road marking.
+    //
+    // \note It is implementation-specific how model_references are resolved to
+    // 3d models.
+    //
+    // \note Road markings' base objects are positioned differently from other
+    // OSI objects: their local z-axis corresponds to how the marking is placed
+    // on the xy-plane of the lane, rather than pointing upwards as it does for
+    // other objects.
+
```
pmai:
```suggestion
    //
```

Issue #909, opened 2026-07-09
Exact type classification of minivan unclear


johschmitz:
In 

https://github.com/OpenSimulationInterface/open-simulation-interface/blob/689eab70138c8fb8c78cce544529753e7647206d/osi_object.proto#L868-L871

it is mentioned that a minivan shall be classified as a car, but not exactly which car type shall be used.

#### Describe the bug
In 

https://github.com/OpenSimulationInterface/open-simulation-interface/blob/689eab70138c8fb8c78cce544529753e7647206d/osi_object.proto#L868-L871

it is mentioned that a minivan shall be classified as a car, but not exactly which car type shall be used.

#### Describe how to reproduce the bug
1. Read the documentation

#### Describe the expected behavior
One explicitly defined class for minivans such as `TYPE_CAR`

#### Show some screenshots
N.A.

#### Describe the OS you are using
N.A.

#### Additional context

Additionally, I expect that this will cause a lot of unnecessary issues when checking sensor output against ground truth since the vehicle perception systems will have a very hard time deciding between a car type minivan and other vans. After all the shape is the main decision criteria, since the perception system has no direct access to vehicle weight information and other details used for the UNECE classification.

I think a) this needs to be re-considered and b) at least one explicit category for minivans needs to be defined if it stays like this.

My personal recommendation would be to losen this and allow this category to be used for minivans as well. Additionally, we should cross-check whether another UNECE classification or ISO norm more related to ADAS/AD/DCAS/Sensors etc. can be used as a reference instead of the one that has been used.

pmai:
> it is mentioned that a minivan shall be classified as a car, but not exactly which car type shall be used.

Since all other types are deprecated, only TYPE_CAR remains, so currently there is no ambiguity, and the specific wording is BTW part of the harmonized ASAM enum definition, hence any change would require a change there (which would be hard to do, as the different classifications in different standards have different specific syntax for the car classification).

Since all of this is centered around fully harmonized definitions, any discussion would need to happen in the harmonization groups, as OSI cannot single-handedly deviate here.

[...]

PR #910, opened 2026-07-09
Update comment to specify minivan TYPE_CAR categorization

johschmitz:
#909 

#### Add a description
Make description unambiguous. 

**Some questions to ask**:
What is this change?
What does it fix?
How has it been tested?

- [ ] I have done the [DCO signoff](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/specification/contributing/dco.html).


thomassedlmayer:
Shouldn't it be clear already which category to use because the other car classifications are deprecated and are not recommended to be used?

johschmitz:
@thomassedlmayer maybe for someone deep in the trenches of OSI this is clear. For someone like me who just has very occasional contact it is not clear. My expectation as a software engineer though would be that it is written as clear as possible for a broad audience. Also my expectation nowadays as an LLM user would be to make APIs and their documentations simple and explicit. It would take an LLM agent significantly more tokens to find out this whole depreciation based constraints and implement it right compared to just writing this explicitly like I am proposing. 
In the end it is up to the maintainers to decide whether to merge this or not. It is just a single line and a single click to merge, I don't think it is worth a lot of additional discussions.

jdsika:
but it does not hurt at all to make this small change?

jdsika:
Trying to recap our discussion we had in at online meeting in the coordination group before holidays:

All comments are directly copied from the [ASAM TrafficParticipants specification](https://publications.pages.asam.net/standards/ASAM_TrafficParticipants_Specification/ASAM_TrafficParticipants_Specification/v1.0.2/specification/07_traffic_participants/07_02_vehicles.html) and we should not alter the comments here in OSI but instead change the specification itself if there is a need. The CCB suggests to close this PullRequest.

johschmitz:
Makes sense to me to copy the text a directly as possible but doesn't make sense to me to not allow to interlink the documentation in a project specific way that fully respects the intention of the traffic participant spec.

How about moving this into a sentence in a note block instead? Apparently note blocks are "allowed" since you are using them in other types as well?
Overall a very frustrating experience for an open source project, instead of suggestions how to improve the PR just closing a PR is not the way to attract more contributions IMHO. Please let me know if moving it to a note block is acceptable before I touch the PR content again.
I think it would then also help to add the information about the harmonization project to the other note blocks that discuss the related depreciations.

thomassedlmayer:
@johschmitz I agree that having `TYPE_CAR` in there would slightly improve the OSI-version of the definition. I wasn't part of the CCB discussions but as someone who participated in the harmonization project I'd say that we really shouldn't create a precedent which might lead to differing definitions in the long-term.

The mapping of cars in OSI is extensively discussed in the [VehicleClassification::Type](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/current/gen/structosi3_1_1MovingObject_1_1VehicleClassification.html#a65c82aaafe22ed6996ab78713354d189) description. That's why I said before that I think it should be sufficiently clear from the documentation which category to use: the former car-specific enums are deprecated in favor of TYPE_CAR, which represents the recommended mapping for cars. This is also reflected in the individual descriptions of the deprecated types. In my opinion the documentation consistently points users toward TYPE_CAR even if in that single case the documentation only mentions "cars". In that sense, all paths through the documentation lead to TYPE_CAR. Another note saying that "car" maps to `TYPE_CAR` to me just seems like a another pointer to TYPE_CAR.

In my opinion the better solution would be to discuss a designated way to insert standard-specific enums/names in harmonized descriptions and also apply this to other standards accordingly. This could certainly be discussed in the next harmonization project (https://code.asam.net/simulation/harmonization/trafficparticipants/traffic-participants-specification/-/issues). Maybe it's useful in other instances too.

___

### More importantly,
having looked at both OSI documentation and Traffic Participants Specification again, I noticed that the sentence was actually moved into the main/normative description block in the harmonized spec after the OSI adaptations have been merged. This is actually something that should be adapted.

image
image

johschmitz:
Updated my commit, maybe this could satisfy all stakeholders now? Otherwise please suggest edits that would lead towards that point.

Note: I corrected mini vans to minivans since I believe that is the correct english spelling (see https://en.wikipedia.org/wiki/Minivan, https://www.merriam-webster.com/dictionary/minivan). But if typos need to be copied over from the original text as well I can reintroduce it.

